PR #1, opened 2016-07-20
Initial mslice

EltayebAhmed:
Initial pull request
Build to .ui files in the project root and the quickview package to be able to run MainQuickView.py


OwenArnold:
Looking at the messages for your commits, could you please use meaningful commit messages in future, so that I, or anyone else can get back to a previous working version of your prototype by checking out a specific commit. "After Lunch" does not tell me at what feature point in the history I'm about to check the source code out for. I suggest using the `git stash` features if you trying to avoid loosing work locally. git is more than a saving/backup tool.

MSliceMainView.py
```diff
@@ -0,0 +1,221 @@
...
+
+class MainView(object):
+    #__metaclass__ = abc.ABCMeta
+    #TODO metaclass?
+
+    #@abc.abstractmethod
+    def update_log(self, log):
+        pass
+
+    # <These Might go>
```
OwenArnold: You say that "these might go". I take it that they are effectively dead-code then. 

I strongly suggest that you ONLY add methods to your MainView as you go along. In other words delete out all the methods that you are not currently using. They'll only get in your way, slow you down, and mislead your design.

MSliceMainView.py
```diff
@@ -0,0 +1,221 @@
+#This file in its current state only exists for reference
+#import abc
```
OwenArnold: You are not using this module, so delete the line and all associated decorators.

MainPresenter.py
```diff
@@ -0,0 +1,5 @@
+
+class MainPresenter():
+    def __init__(self,MainView):
```
OwenArnold: Good. I suggest keeping the `MainView` argument as lower case i.e. `main_view` . You are passing an object of type `MainView` not the class.

MainQuickView.py
```diff
@@ -0,0 +1,28 @@
+from workspacemanager.WorkspaceManagerQuickView import WorkspaceManagerQuickView
+from powderprojection.PowderProjectionQuickView import PowderProjectionQuickView
+from MainPresenter import MainPresenter
+import workspacemanager,powderprojection
+
+#TODO Error in logging framework, FIX/ASK !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+class MainQuickView():
```
OwenArnold: What is MainQuickView?. I know, but someone else won't. Docstring needed at the very least.

README.md
```diff
@@ -1 +1 @@
-# mslice
\ No newline at end of file
+# MSlice3.0
```
OwenArnold: Can we have some more information in this. It doesn't need to contain too much. But it should include 

1. A statement about the fact that this is a prototype, and currently not production ready. 
1. What is this application supposed to do. Only need to give one or two sentences for this at a very high level.
1. Information on what modules are needed to run this. The fact that MantidPlot needs to be installed etc. Which version of MantidPlot. 
1. How to run the tests
1. How to run the `QuickView`
1. What existing projects this is based upon (MSlice 2.0, MSlice, etc)

Here are some good examples of readmes [1](https://github.com/DMSC-Instrument-Data/plankton), [2](https://github.com/DMSC-Instrument-Data/instrument-prototype)

You can make a separate Issue/PR for this if you wish.

command.py
```diff
@@ -0,0 +1,24 @@
+#This file in its current state only exists for reference
+class Command:
```
OwenArnold: Delete out ALL commands you are not currently using. The point of developing a GUI in the `MVP` way is to allow it to grow organically. We don't project future requirements, and especially not bits of code until we actually need them.

mslice_main_window.py
```diff
@@ -0,0 +1,56 @@
...
+from command import Command
+from ui_Mslice import Ui_MainWindow
+
+class MSliceGui(MainView,QtGui.QMainWindow,Ui_MainWindow):
+    #TODO Should i set a metaclass?
+    def __init__(self):
+        super(QtGui.QMainWindow,self).__init__()
+        self.setupUi(self)
+        #Create a presenter for the GUI
+        self.presenter = MainViewPresenter(self)
```
OwenArnold: Should be `self._presenter` please

mslice_main_window.py
```diff
@@ -0,0 +1,56 @@
+from PyQt4 import QtGui
+
+from MSliceMainView import MainView
+from workspacemanager.WorkspaceManagerPresenter import MainViewPresenter
+from command import Command
+from ui_Mslice import Ui_MainWindow
+
+class MSliceGui(MainView,QtGui.QMainWindow,Ui_MainWindow):
+    #TODO Should i set a metaclass?
```
OwenArnold: If you are allowed to create instances of this GUI, such that you can inherit from both `MainView` and `Ui_MainWindow` then no don't add it. I had to use the `metaclass` fix as a work around.

powderprojection/MantidProjectionCalculator.py
```diff
@@ -0,0 +1,18 @@
+from powderprojection.ProjectionCalculator import ProjectionCalculator
+from mantid.simpleapi import SofQWNormalisedPolygon,mtd
+from math import sqrt,cos
+
+
+class MantidProjectionCalculator(ProjectionCalculator):
+    def CalculateProjections(self,input_workspace,output_workspace,qbinning,axis1,axis2):
```
OwenArnold: Should not be using upper case for function names see [PEP8](https://www.python.org/dev/peps/pep-0008/#function-names)

powderprojection/PowderProjectionQuickView.py
```diff
@@ -0,0 +1,10 @@
+from quickview.QuickView import QuickView
+from powderprojection.PowderProjectionPresenter import PowderProjectionPresenter
+from powderprojection.MantidProjectionCalculator import MantidProjectionCalculator
+
+
+class PowderProjectionQuickView(QuickView):
+    def __init__(self,mainView,commands):
+        super(PowderProjectionQuickView,self).__init__(commands)
+        proj_calculator = MantidProjectionCalculator()
```
OwenArnold: `MantidProjectionCalculator` is a type which inherits from `ProjectionCalculator`. Looking through the code you never create a `mock` of this. This rings warning bells. 

The `PowderProductionPresenter` takes a `proj_calculator` as an argument. I therefore strongly suspect that you have NO tests for the `PowderProjectionPresenter`, desipite the fact that you have added it to the `QuickView`!

In future I think you should ALWAYS.

1. Write the test you need.
1. Create your presenter, calculators etc. And add the features one at a time to make the tests pass.
1. Refactor the original code to make it neater.
1. Repeat

Only when you have all your tests in place, should you consider adding corresponding `QuickViews` to support these features. DO NOT use quickviews as the basis for your testing.

Please fix this.

quickview/ui_mock.py
```diff
@@ -0,0 +1,55 @@
+from getter_ui import Ui_Dialog
+from PyQt4 import QtGui
+
+
+class Getter(Ui_Dialog,QtGui.QDialog):
```
OwenArnold: Class called `Getter` means nothing to me. Needs docstring, but proper fix is to name it something meaningful. Is this a `FileLoaderDialog`?

quickview/ui_mock.py
```diff
@@ -0,0 +1,55 @@
...
+    def is_done(self):
+        return self._done
+
+    def get_data(self):
+        return self._data
+
+    def on_click(self):
+        sender = self.sender()
+        text = str(self.lineEdit.text())
+        if sender == self.btnInt:
```
OwenArnold: Not sure what's going on here. You have different buttons for different types of input?

workspacemanager/MantidWorkspaceProvider.py
```diff
@@ -0,0 +1,17 @@
+from WorkspaceProvider import WorkspaceProvider
+from mantid.simpleapi import mtd,Load,DeleteWorkspace,Load,GroupWorkspaces,RenameWorkspace,SaveNexus
+
+class MantidWorkspaceProvider(WorkspaceProvider):
```
OwenArnold: This is generally good. Doc strings needed, and casing should match PEP8 standards.

workspacemanager/WorkspaceManagerPresenter.py
```diff
@@ -0,0 +1,75 @@
+from workspacemanager.command import Command
+import os.path
+
+
+#TODO askOwen is passing mtd.getObjectNames() ideal? pass mtd? Answer: Will do for time being
```
OwenArnold: Pass a list of strings to the View for display. As a word of warning. In other operations you will need to pass the actual workspaces, so don't try to solve every problem with strings.

workspacemanager/WorkspaceManagerPresenter.py
```diff
@@ -0,0 +1,75 @@
...
+        if command == Command.LoadWorkspace:
+            #TODO specify workspace name on load
+            #TODO what to do on fail?
+            workspace_to_load = self._workspaceMangerView.get_workspace_to_load_path()
+            base = os.path.basename(workspace_to_load)
+            ws_name = os.path.splitext(base)[0]
+            self._workSpaceprovider.Load(Filename=workspace_to_load, OutputWorkspace=ws_name)
+            self._workspaceMangerView.display_loaded_workspaces(self._workSpaceprovider.getWorkspaceNames())
+
+        if command == Command.SaveSelectedWorkspace:
```
OwenArnold: 1. Should be an `elif` 
2. You should add a fall-through block that raises an exception if no `Command` matches! This should be a scenario in your tests.
3. Refactor this so that the "what" and "how" are better separated in the code. Notify should look more like this.

``` python
def notify(self, command):
    if command == Command.LoadWorkspace:
        load_workspace()
    elif command == Command.SaveSelectedWorkspace:
        save_selected_workspace()
#etc
```

workspacemanager/WorkspaceManagerPresenter.py
```diff
@@ -0,0 +1,75 @@
...
+                return
+            selected_workspace = selected_workspaces[0]
+            newName = self._workspaceMangerView.get_workspace_new_name()
+            self._workSpaceprovider.RenameWorkspace(selected_workspace, newName)
+            self._workspaceMangerView.display_loaded_workspaces(self._workSpaceprovider.getWorkspaceNames())
+
+    def get_selected_workspaces(self):
+        return self._workspaceMangerView.get_workspace_selected()
+
+    def refresh(self):
```
OwenArnold: I suggest a more descriptive name than `refresh` and definitely a docstring.

workspacemanager/WorkspaceManagerPresenterTest.py
```diff
@@ -0,0 +1,173 @@
...
+class WorkspaceManagerPresenterTest(unittest.TestCase):
+    def setUp(self):
+        self.workspace_provider = mock.create_autospec(spec=WorkspaceProvider)
+        self.view = mock.create_autospec(spec=WorkspaceView)
+
+
+
+    def test_load_one_workspace(self):
+        self.presenter = WorkspaceManagerPresenter(self.view, self.workspace_provider)
+        self.view.get_workspace_to_load_path = mock.Mock(return_value=r'C:\a\b\cde.nxs')
```
OwenArnold: This needs better annotation in general. For example 
```python
# Creating a view that will return a path to a nexus file on call to get_workspace_to_load_path
self.view.get_workspace_to_load_path = mock.Mock(return_value=r'C:\a\b\cde.nxs')
```
This will help the next maintainer a lot.

OwenArnold: I suggest making a local variable from your string path name so that it can be matched better in the assertion.

```python
expected_path = 'C:\a\b\cde.nxs'
```
or similar. This prevents future coding mistakes

workspacemanager/WorkspaceManagerPresenterTest.py
```diff
@@ -0,0 +1,173 @@
...
+        self.workspace_provider.getWorkspaceNames = mock.Mock(side_effect=[['file1'], ['file1', 'file2'],
+                                                                           ['file1', 'file2', 'file3']])
+        for i in range(3):
+            self.presenter.notify(Command.LoadWorkspace)
+        load_calls = [call(Filename='C:\\path\\to\\file1.nxs', OutputWorkspace='file1'),
+                      call(Filename='C:\\path\\to\\any\\file2.nxs', OutputWorkspace='file2'),
+                      call(Filename='C:\\path\\to\\file3.nxs', OutputWorkspace='file3')]
+        self.workspace_provider.Load.assert_has_calls(load_calls)
+        #The test should just finish here
+        # <Not Maybe not useful> TODO askOwen case1
```
OwenArnold: Either possibility is fine that you mention here. Lets keep the full option (verbose) for now.

workspacemanager/WorkspaceManagerPresenterTest.py
```diff
@@ -0,0 +1,173 @@
...
+        display_workspace_calls = [call(['file1']), call(['file1', 'file2']), call(['file1', 'file2', 'file3'])]
+        self.view.display_loaded_workspaces.assert_has_calls(display_workspace_calls)
+        # </Maybe not useful>
+
+    # TODO make sure that workspace provider is not called on failure cases
+    def test_rename_workspace(self):
```
OwenArnold: You have a lot of "Happy Path" test cases, but very few "Destructive Test Cases". Destructive test cases are more useful. Rename your existing test cases to things like `test_rename_workspace_succeeds` and then add tests for cases in which the operation cannot succeed for various reasons and add test cases for those `test_rename_workspace_throws_when_X` where X is some failure scenario. I estimate that you should have about a 2:1 ratio (or greater) between destructive and happy-path test cases.

workspacemanager/WorkspaceManagerPresenterTest.py
```diff
@@ -0,0 +1,173 @@
...
+        self.presenter = WorkspaceManagerPresenter(self.view, self.workspace_provider)
+        self.view.get_workspace_selected = mock.Mock(return_value=[])
+
+        self.presenter.notify(Command.RemoveSelectedWorkspaces)
+        self.view.get_workspace_selected.assert_called_once_with()
+        self.view.error_select_one_or_more_workspaces.assert_called_once_with()
+        self.workspace_provider.DeleteWorkspace.assert_not_called()
+        self.view.display_loaded_workspaces.assert_not_called()
+
+
```
OwenArnold: Generally. The test scenarios here are very good. Well Done!

OwenArnold:
@EltayebAhmed. Please make the fixes here that I've outlined. I'll do anther review of this PR when you are done.

EltayebAhmed:
I have addressed all issues except the readme

Tests/PowderProjectionPresenterTest.py
```diff
@@ -0,0 +1,62 @@
...
+from mainview import MainView
+from MainPresenter import MainPresenter
+
+
+class PowderProjectionPresenterTest(unittest.TestCase):
+    def setUp(self):
+        # Set up a mock view, presenter, main view and main presenter
+        self.powder_view = mock.create_autospec(PowderView)
+        self.projection_calculator = mock.create_autospec(ProjectionCalculator)
+        self.main_presenter = mock.create_autospec(MainPresenter)
```
OwenArnold: Is this a copy-paste error? Why do you need the `MainPresenter`?

Tests/SlicePlotterPresenterTest.py
```diff
@@ -0,0 +1,83 @@
...
+        self.main_view.get_presenter = mock.Mock(return_value=self.main_presenter)
+        # Setting up view for slice plotter window
+        self.slice_plotter = mock.create_autospec(SlicePlotter)
+        self.slice_view = mock.create_autospec(SlicePlotterView)
+
+    def test_constructor_success(self):
+        self.slice_plotter_presenter = SlicePlotterPresenter(self.main_view, self.slice_view, self.slice_plotter)
+
+    def test_constructor_invalid_main_view_failure(self):
+        self.assertRaises(TypeError, SlicePlotterPresenter, self.slice_view, self.slice_view, self.slice_plotter)
```
OwenArnold: `ValueError` is probably better than `TypeError` unless you are really checking types (and you probably shouldn't be).

OwenArnold:
OK. I haven't run the code, but the changes look good. This looks like good MVP to me.

As something to strongly consider going forward, I'm still a little unhappy about the title of the `PR` because it gives me no sense of what is being delivered in terms of functionality. "run MainQuickView.py" is too loose. I suggest that you now work with @NickDraper to identify "stories" in terms of feature points that need to be achieved. @NickDraper can help you extract one-or two of these at a time from the requirements starting from the simplest. I can then work with you to help you build these stories into the code.

I'm happy to merge this current PR.

PR #3, opened 2016-08-04
Restructured code

EltayebAhmed:
Restructured the code
And partially hooked up workspace manager presenter to gui


martyngigg:
@EltayebAhmed Generally this looks good. At some point we will need to find homes for things like the startup scripts etc but that can wait for now.

A couple of things I notice:
* both `pylint-check` & `pylint_check` exist in the main directory - I assume we only want one of these files? Can we also move it into a `tools` subdirectory?
* I can see there is a `Tests` directory and also a `tests` directory. Please move what is in `Tests` to `tests` and use lower case names for all of the filenames

@OwenArnold Any more thoughts on the structure?

OwenArnold:
* I would prefer that `widgets` only contained concrete pyqt4 `views`
* I would prefer to see a `views` directory that contained only the abstract views
* I would prefer to see all `presenters` in a presenters directory
* I would prefer to see all of your models like `MantidWorkspaceProvider` in a `models` directory

martyngigg:
@OwenArnold Is this a more standard layout for MVP-style approaches?

martyngigg:
@EltayebAhmed Can you make the updates that both Owen and I have suggested?

EltayebAhmed:
@OwenArnold  Where should the command files (the ones where i put the constants for presenter.notify()) go

OwenArnold:
@EltayebAhmed. Well those need to be accessible to the presenters but I think they're private to the views. So they could possibly live with the presenters. I'm not too worried about where they end up living.

EltayebAhmed:
Ready to merge?

OwenArnold:
@EltayebAhmed I would say no. Your tests are failing on Travis CI, lets always make sure that we have a green light on the CI before we merge anything.

OwenArnold:
When you get a chance, could you add a more verbose output to the test execution. Also inclusion of coveralls would be good.

EltayebAhmed:
@OwenArnold Could you point me to an example of some verbose tests?

EltayebAhmed:
@OwenArnold Coveralls is now working

mainview.py
```diff
@@ -1,5 +1,5 @@
-from workspacemanager.WorkspaceView import WorkspaceView
-from MainPresenter import MainPresenter
+from main_presenter import MainPresenter
```
OwenArnold: Looks like MainPresenter lives outside the reset of the presenters?

Also. I'm still seeing a lot of inconsistencies with the casing. You use lower case for modules in some places but not in others.

models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -1,17 +1,22 @@
-from WorkspaceProvider import WorkspaceProvider
+from workspace_provider import WorkspaceProvider
 from mantid.simpleapi import mtd,Load,DeleteWorkspace,Load,GroupWorkspaces,RenameWorkspace,SaveNexus
 
 class MantidWorkspaceProvider(WorkspaceProvider):
+
     def getWorkspaceNames(self):
```
OwenArnold: There's problems with case consistency here too.

mainview.py
EltayebAhmed: all modules are now lowercase

OwenArnold:
Great. Thanks for adding this.

1. Create "badges" for the master build and coverage output and add them to the .README file. Have a look at how this has been done with Plankton.
2. Have a serious look at any areas that are not adequately being covered by your tests. I think 90% coverage would be a good target rule/guideline. However, you should read the report and make common sense judgements (is this area going to be prone to errors).

models/workspacemanager/mantid_workspace_provider.py
EltayebAhmed: I have changed the first functions name, since this module is such a thin wrapper around mantid i have made the function names as similar as possible to the mantid algorithms, should i make the names PEP compliants?

OwenArnold: Absolutely. Remember, from the `presenters` point of view, the `MantidWorkspaceProvider`, and what it runs under-the-hood (manitd.simpleapi) is implementation detail. Follow the PEP8 standard there's no reason to let the `mantid.simpleapi` dictate the casing.

models/workspacemanager/workspace_provider.py
```diff
@@ -2,7 +2,7 @@
 
 class WorkspaceProvider:
     @abc.abstractmethod
-    def getWorkspaceNames(self):
+    def mtd_getObjectNames(self):
```
OwenArnold: Don't call it `mtd`. This supposes the use of Mantid. Remember the Presenter is not supposed to know what concrete models it's using. You're leaking dependency information into the presenter with this. Just call it `getObjectNames`.

.travis.yml
```diff
@@ -0,0 +1,13 @@
+language: python
+python:
+  - "2.7"
+# command to install dependencies
+install: 
+  pip install -r pip-requirements.txt
+
+# command to run tests
+script: 
+  nosetests --with-coverage --cover-package=presenters,plotting.FigureManager
```
OwenArnold: Is this actually running your unit tests? I'm not convinced it is? All unit tests must be run as part of the CI. We need to get the outputs of the executed tests in the output console.

models/workspacemanager/workspace_provider.py
EltayebAhmed: @OwenArnold  I made the names case compliant and abandoned mimicking the underlying algorithm's casing

.travis.yml
EltayebAhmed: @OwenArnold i have now ammended that

EltayebAhmed:
@OwenArnold did you have a paid subscription with coveralls for plankton? because I can't find the badge

models/slice/matplotlib_slice_plotter.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_aspect_ratio(self, workspace):
+        return 'auto'
+
+
+class MatplotlibSlicePlotter(SlicePlotter):
+    def display_slice(self, selected_workspace, x_axis, y_axis, smoothing, intensity_start, intensity_end, norm_to_one,
+                      colourmap):
+        workspace = mtd[selected_workspace]
+        plot_method = 'imshow'
```
martyngigg: Is there a reason we have `plot_method` hard coded but then `if` blocks for other methods? We shouldn't add more code than we need at the current time.

OwenArnold:
@EltayebAhmed No. You might want to ask @MichaelWedel what he did as he did that configuration.

models/slice/matplotlib_slice_plotter.py
```diff
@@ -0,0 +1,47 @@
+from slice_plotter import SlicePlotter
+from mantid.simpleapi import mtd
+from plotting import pyplot as plt
+import numpy as np
+
+
+def get_aspect_ratio(self, workspace):
+        return 'auto'
```
martyngigg: This is indented too much.

models/slice/matplotlib_slice_plotter.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class MatplotlibSlicePlotter(SlicePlotter):
+    def display_slice(self, selected_workspace, x_axis, y_axis, smoothing, intensity_start, intensity_end, norm_to_one,
+                      colourmap):
+        workspace = mtd[selected_workspace]
```
martyngigg: This is quite picky but can we avoid using `mtd` and using `AnalysisDataService` instead? I think libraries such as this should be verbose as possible and `mtd` reserved for short hand in user space.

EltayebAhmed:
@martyngigg 
These are alternatives that may be needed in the future . The other methods are much slower but take xdata into consideration. Using imshow puts in the assumption that the bin centers are equidistant and identical throughout the workspace, Nick said that would be fine for now. So i left them as a reminder. But I can remove them if that would be better?

models/slice/matplotlib_slice_plotter.py
OwenArnold: 👍 That's important. We'll build the gui up as per the features required at the time that the features are specified. Don't put code in thats not currently needed.

.travis.yml
OwenArnold: @EltayebAhmed Yes, travis output Looks good.

mslice.py
```diff
@@ -0,0 +1,24 @@
...
+        super(MsliceGui,self).__init__()
+        self.setupUi(self)
+        workspace_presenter = self.wgtWorkspacemanager.get_presenter()
+        self._presenter = MainPresenter(self,workspace_presenter)
+
+        self.wgtWorkspacemanager.set_main_window(self)
+        self.wgtSlice.set_main_window(self)
+
+if __name__ == "__main__":
+    from build_all_ui import build_all_ui_files,script_folder
```
martyngigg: @EltayebAhmed I can't find a reference to `build_all_ui` anywhere? Where does this come from?

EltayebAhmed:
@martyngigg sorry the file was matching the gitignore pattern. I have now renamed it

MichaelWedel:
When I look at the plankton coveralls page (https://coveralls.io/github/DMSC-Instrument-Data/plankton?branch=master), there is a box on the upper right, which says "badge" and has a small drop-down menu "embed". From there you can links in different formats.

build_all_ui_files.py
```diff
@@ -0,0 +1,46 @@
+BUILDER = r'C:\Users\OMi32458\Workspace\mantid\external\src\ThirdParty\lib\python2.7\pyuic4.bat'
```
martyngigg: @EltayebAhmed We are going to have to rework this at some point so it both works on other machines and on other platforms. We probably want to look at using the `PyQt4.uic` http://pyqt.sourceforge.net/Docs/PyQt4/designer.html#the-uic-module. This PR is getting quite large though so let's leave it as a future issue.

EltayebAhmed:
@MichaelWedel Thanks, it is now up

MichaelWedel:
Are you sure you want to point at that branch instead of master? After merging the badge should be alright for master and the coverage for PRs is indicated in the "All checks have passed"-thing.

build_all_ui_files.py
OwenArnold: @EltayebAhmed. I added macros to mantid so that the cmake scripts automatically create the UI files. You might want to look into that to see how I got this working cross-platform. But as Martyn said `PyQt4.uic` is probably the right approach.

pip-requirements.txt
```diff
@@ -0,0 +1,2 @@
+mock>=2.0.0
+coveralls
```
OwenArnold: pip requirements missing `nose`. You get away with this because both `mock` and `nose` come pre-installed, but you should never the less have it specified in your requirements for other developers.

.travis.yml
```diff
@@ -0,0 +1,13 @@
+language: python
+python:
+  - "2.7"
+# command to install dependencies
+install: 
+  pip install -r pip-requirements.txt
+
+# command to run tests
+script: 
+  nosetests -v --with-coverage --cover-package=presenters,plotting.figuremanager
```
OwenArnold: I strongly suggest fixing the mimimum required coverage to be at your current coverage level (at least) so that if there's a drop in coverage the CI will break on future PR requests. 
nose tests has a `--cover-min-percentage` option that you can specify. 

This will be crude mechanism of ensuring that future developers and maintainers are forced to keep the test coverage at a mimimum standard.

I also think that the `--cover-branches` option should be added, but that's probably best left for another PR.

pip-requirements.txt
OwenArnold: Also, I think something to bear in mind soon is that this `requirements.txt` covers the bare essentials for running the tests (which is good). But there's not enough here to actually run something mslice like. You might consider very soon (different PR) adding a separate script to pull in `mantid`. @martyngigg what do you think?

martyngigg: If this is a completely separate thing then it should probably just rely on `mantid` being installed and available on the python path. Or do you mean from a developer perspective?

OwenArnold: There's several issues.

1. We should attempt to verify that Mantid is installed. Although we don't want this to interfere with our ability to run the tests against the presenters. 
1. mslice may only work with a specific version of Mantid, so I think it would be important to verify that you have the version of mantid that you expect before mslice actually tries to do anything.
1. If we are relying on an external mantid install, we should document that in the README

martyngigg: @OwenArnold Agreed. These all seem like they don't need to be solved in this PR though? It is getting rather large.

.travis.yml
martyngigg: Once we've added this I'm going to suggest we merge this?

MichaelWedel: For the local tests that is good, but for PRs you can set limits in coveralls. You can specify that the build should fail if coverage goes down at all, or a certain percentage.

OwenArnold: @MichaelWedel I don't mind which mechanism is followed, but I think for these MVP approaches, robustness from the get-go is so important, I think enforcing (and in fact becoming gradually stricter) on test coverage is required. Could you help @EltayebAhmed configure it in coveralls if that is the preferred approach? I've never set coveralls up myself.

pip-requirements.txt
OwenArnold: @EltayebAhmed. Please make github issues for your repository about the improvements that are not going to make this PR. @martyngigg has highlighted a few that are important but are not required for this PR.

EltayebAhmed:
@martyngigg @martyngigg 
I have created an issue for the dependency resolution and set 80% minimum for build to pass

.travis.yml
```diff
@@ -7,7 +7,7 @@ install:
 
 # command to run tests
 script: 
-  nosetests -v --with-coverage --cover-package=presenters,plotting.figuremanager
+  nosetests -v --with-coverage --cover-min-percentage=80 --cover-package=presenters,plotting.figuremanager
```
OwenArnold: 👍

martyngigg:
Is this ready for merging?

EltayebAhmed:
@martyngigg 
I have either addressed or opened issues for everything you and @OwenArnold have mentioned so far.

OwenArnold:
@martyngigg. I think this is ready to merge. Do you want to do the honours?

martyngigg:
Nice work, @EltayebAhmed.

PR #4, opened 2016-08-10
Add routines for plotting 1D and 2D

AndreiSavici:
Added a new directory for random stuff. I started with my scripts to do some simple 1D and 2D plots


martyngigg:
Sorry to be a pain @AndreiSavici. Can you rename the directory `misc` instead? It's now more consistent with the restructure.

martyngigg:
Thanks!

PR #7, opened 2016-08-11
Implement zoom

EltayebAhmed:
Refactored PlotFigure class into a base_plot_figure_class and a base_qt_plot_figure and a concrete class
Implemented zoom in and zoom out functionality.

**To test:**
Bring up any plotting window (either by plotting a slice in mslice or using the plotting.pyplot library and click the zoom in button. 
Then draw using click and drag draw a rectangle over an area to zoom in to that area. It will add the previous view to a an internal stack
Clicking zoom out will undo the last zoom and remove it from the zoom stack. If the stack is empty zoom out will do nothing.

### DIFF
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -1,15 +1,16 @@
-
-from mantid.simpleapi import AnalysisDataService,ConvertToMD
-
+from mantid.simpleapi import ConvertToMD
 from models.projection.powder.projection_calculator import ProjectionCalculator
 from models.workspacemanager.mantid_workspace_provider import MantidWorkspaceProvider
 
+MD_SUFFIX = '_MD'
+
 
 class MantidProjectionCalculator(ProjectionCalculator):
-    def calculate_projection(self, input_workspace, output_workspace, axis1, axis2):
-        import random,string
-        letters = list(string.ascii_lowercase)
-        random.shuffle(letters)
-        input_workspace = MantidWorkspaceProvider.get_workspace_handle(input_workspace)
-        x = ConvertToMD(input_workspace,'|Q|','Direct')
-        raise NotImplementedError('Implement me')
+    def calculate_projection(self, input_workspace, axis1, axis2):
+        output_workspace = input_workspace + MD_SUFFIX
+        if axis1 == '|Q|' and axis2 == 'Energy':
+            ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=output_workspace, QDimensions='|Q|',
+                        PreprocDetectorsWS='-')
+        else:
+            raise NotImplementedError('MSlice currently only supports projection to Energy vs |Q|')
+
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -1,22 +1,15 @@
-from math import sqrt,cos
 
-from mantid.simpleapi import SofQWNormalisedPolygon,mtd
+from mantid.simpleapi import AnalysisDataService,ConvertToMD
 
 from models.projection.powder.projection_calculator import ProjectionCalculator
+from models.workspacemanager.mantid_workspace_provider import MantidWorkspaceProvider
 
 
 class MantidProjectionCalculator(ProjectionCalculator):
-    def calculate_projection(self, input_workspace, output_workspace, qbinning, axis1, axis2):
-        if axis1 == 'Energy' and axis2 == '|Q|':
-            #TODO is EMode always direct ?
-            SofQWNormalisedPolygon(InputWorkspace=input_workspace, OutputWorkspace=output_workspace,
-                                                                QAxisBinning=qbinning,EMode='Direct')
-
-    def calculate_suggested_binning(self, input_workspace):
-        return '0,.1,10'
-        #TODO AskInstrument scientist about calculating scattering angles
-        input_workspace = mtd[input_workspace]
-        ei = float(input_workspace.getRun().getLogData('Ei').value)
-        ki = sqrt(ei/2.07)
-        kf = sqrt((ei-0.0)/2.07)
-        Qx=ki*1-cos(minTheta)*kf
+    def calculate_projection(self, input_workspace, output_workspace, axis1, axis2):
+        import random,string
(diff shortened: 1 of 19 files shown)


plotting/plot_window/base_qt_plot_window.py
```diff
@@ -0,0 +1,92 @@
...
+                                   QtGui.QSizePolicy.Expanding)
+        FigureCanvas.updateGeometry(self)
+        self._rect_cid = None
+        self._motion_cid = None
+        self._new_axes_limits = [[None,None],[None,None]] # For zooming [[x_lower,x_upper],[y_lower,y_upper]]
+        # TODO implement zoom history stack
+        self._zoom_history_stack = []
+        self._zoom_rect = None
+
+    def zoom_out(self):
```
EltayebAhmed: Is putting the zoom related functions in the canvas ok? I did that because most tools i needed were attributes of the canvas, although it would not be difficult to move this code to the window

plotting/plot_window/base_qt_plot_window.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _redraw_zoom_rect(self,mouse_event):
+        self._new_axes_limits[0][1] = mouse_event.xdata
+        self._new_axes_limits[1][1] = mouse_event.ydata
+        if self._zoom_rect:
+            self._zoom_rect.remove()
+        x,y = self._new_axes_limits[0][0], self._new_axes_limits[1][0]
+        width, height = map(lambda x: x[1]-x[0], self._new_axes_limits)
+        self._zoom_rect = Rectangle((x, y), width, height, fill=False)
+        self.axes.add_patch(self._zoom_rect)
```
EltayebAhmed: Is using matplotlib to draw the zoom rectangle (shown while selecting region to zoom in on) OK? 
it comes with much less code overhead compared to using a qpainter and overriding the paintEvent function

EltayebAhmed:
@martyngigg

models/projection/powder/mantid_projection_calculator.py
```diff
@@ -1,22 +1,15 @@
...
-        return '0,.1,10'
-        #TODO AskInstrument scientist about calculating scattering angles
-        input_workspace = mtd[input_workspace]
-        ei = float(input_workspace.getRun().getLogData('Ei').value)
-        ki = sqrt(ei/2.07)
-        kf = sqrt((ei-0.0)/2.07)
-        Qx=ki*1-cos(minTheta)*kf
\ No newline at end of file
+    def calculate_projection(self, input_workspace, output_workspace, axis1, axis2):
+        import random,string
```
martyngigg: Is there a reason this has been modified when the PR is related to implementing zooming? This seems completely unrelated.

models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -19,4 +19,9 @@ def rename_workspace(self, selected_workspace, newName):
         return RenameWorkspace(selected_workspace,newName)
 
     def save_nexus(self, workspace, path):
-        SaveNexus(workspace,path)
\ No newline at end of file
+        SaveNexus(workspace,path)
+
+    @staticmethod
+    def get_workspace_handle(workspace_name):
```
martyngigg: What's the reason that this is a static method when all other methods are instance methods?

plotting/plot_window/base_qt_plot_window.py
martyngigg: How is this handled by the standard canvas? I would suggest following a similar style.

plotting/plot_window/base_qt_plot_window.py
martyngigg: That seems fine. As long as the interactive performance is okay then this seems reasonable.

plotting/pyplot.py
```diff
@@ -6,6 +6,7 @@ def draw_if_interactive():
     # We will always draw because mslice might be running without matplotlib interactive
     for figure in FigureManager._figures.values():
         figure.canvas.draw()
+        figure.canvas.draw()
```
martyngigg: Does there need to be 2 draw calls here?

EltayebAhmed:
@martyngigg Issues resolved

EltayebAhmed:
Moving mouse out of bounds while zooming breaks zooming

plotting/plot_window/base_qt_plot_window.py
```diff
@@ -45,10 +45,16 @@ def _start_draw_zoom_rect(self,mouse_event):
 
 
     def _redraw_zoom_rect(self,mouse_event):
+        cache = [self._new_axes_limits[0][:],self._new_axes_limits[1][:]]
         self._new_axes_limits[0][1] = mouse_event.xdata
         self._new_axes_limits[1][1] = mouse_event.ydata
         rect = zip(*self._new_axes_limits)
-        rect = self.axes.transData.transform(rect)
+        try:
```
EltayebAhmed: If user releases mouse while out of bounds will result in strange zooming bounds, but if you bring the mouse back in before releasing it will zoom as expected.
Is that acceptable behaviour?

martyngigg: @EltayebAhmed I don't think that is acceptable unless there's absolutely on way around it.

martyngigg:
This seems to work as expected. One problem I have though is that I can't really see the zoom rectangle as on my system it is black on dark blue with the current colour map.

EltayebAhmed:
@martyngigg  Is it poorly visible
or totally invisible?

martyngigg:
@EltayebAhmed As we discussed we will leave this entirely for `matplotlib` to handle.

The branch has also developed a conflict.

EltayebAhmed:
@martyngigg After merging in master and removing the custom zoom implementation this branch now effectively contains a refactor of the plot_figure to a base class and concrete class.
It now candidate for review and merging as is, Should i rename the pull request? or open a new one from the same branch

martyngigg:
@EltayebAhmed Is it intentional that the zooming no longer works?

EltayebAhmed:
@martyngigg  yes this branch is now effectively just a refactor and not an implement zoom branch, Should i open a new pull request?

martyngigg:
Yes, please close this & rename the branch then open a new pull request on that branch.

PR #10, opened 2016-08-12
Fixed Broken import

EltayebAhmed:
Fixed import broken due to file changing location
@martyngigg

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,3 @@ ENV/
 
 # Rope project settings
 .ropeproject
-
-#Files i create just for debugging
-debugbug.py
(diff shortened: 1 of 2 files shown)


martyngigg:
@EltayebAhmed Thanks. Just for future reference you don't always need to reference people in the initial comment. If you need someone specific to look at this then consider assigning them using the "Assignees" box on the right.

.gitignore
```diff
@@ -98,3 +98,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+#Files i create just for debugging
+debugbug.py
```
martyngigg: @EltayebAhmed The global .gitignore should not be used for masking personal files. There is a `.git/info/exclude` file that can be used for that.

EltayebAhmed:
@martyngigg 
I have reverted the gitignore editing commit

martyngigg:
Thanks.

PR #13, opened 2016-08-12
Implement calculate projections

### LINKED ISSUE #11: Implement Calculate projections
Minimal functionality to convert the currently selected workspace to an MD workspace.
Will initially only support projecting to DeltaE vs |Q|

EltayebAhmed:
Will transform the current selected workspace to MD workspace if user clicks on calculate powder projection

**To Test:**
Load a workspace 2d into mslice, select the checkbox next to it in the workspace manager
And click on the calculate projections button on the powder tab
A new workspace should appear in the in the workspacemanager tab with the suffix 'MD'

resolves #11

### DIFF
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -1,11 +1,13 @@
 from mantid.simpleapi import ConvertToMD
 from models.projection.powder.projection_calculator import ProjectionCalculator
 
+MD_SUFFIX = '_MD'
+
 
 class MantidProjectionCalculator(ProjectionCalculator):
     def calculate_projection(self, input_workspace, output_workspace, qbinning, axis1, axis2):
         if axis1 == '|Q|' and axis2 == 'Energy':
-            ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=input_workspace+'_MD',QDimensions='|Q|',
+            ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=input_workspace + MD_SUFFIX, QDimensions='|Q|',
                         PreprocDetectorsWS='-')
         else:
             raise NotImplementedError('MSlice currently only supports projection to Energy vs |Q|')
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -5,7 +5,8 @@ from models.projection.powder.projection_calculator import ProjectionCalculator
 class MantidProjectionCalculator(ProjectionCalculator):
     def calculate_projection(self, input_workspace, output_workspace, qbinning, axis1, axis2):
         if axis1 == '|Q|' and axis2 == 'Energy':
-            ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=input_workspace+'_MD',QDimensions='|Q|')
+            ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=input_workspace+'_MD',QDimensions='|Q|',
+                        PreprocDetectorsWS='-')
         else:
             raise NotImplementedError('MSlice currently only supports projection to Energy vs |Q|')
 
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -5,9 +5,10 @@ MD_SUFFIX = '_MD'
 
 
 class MantidProjectionCalculator(ProjectionCalculator):
-    def calculate_projection(self, input_workspace, output_workspace, qbinning, axis1, axis2):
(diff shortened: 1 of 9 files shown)


EltayebAhmed:
@martyngigg

martyngigg:
@EltayebAhmed Thanks for the pull request.

Could you update the description and title to follow the guidelines [here](http://www.mantidproject.org/Git_Workflow#Pull_Requests). https://github.com/mantidproject/mantid/pull/17231 is an example of a good pull request that shows a description and testing instructions.

widgets/projection/powder/powder.py
```diff
@@ -1,8 +1,44 @@
...
 from PyQt4.QtGui import QWidget
+from presenters.powder_projection_presenter import PowderProjectionPresenter
+from models.projection.powder.mantid_projection_calculator import MantidProjectionCalculator
+from views.powder_projection_view import PowderView
+from command import Command
 
 
-class PowderWidget(QWidget,Ui_Form):
+class PowderWidget(QWidget,Ui_Form,PowderView):
+    """This widget is not usable without a main window which implements mainview"""
```
martyngigg: @EltayebAhmed Does this mean that this cannot be embedded within another widget?

widgets/projection/powder/powder.py
```diff
@@ -1,8 +1,44 @@
...
     def __init__(self,*args, **kwargs):
         super(PowderWidget,self).__init__(*args, **kwargs)
         self.setupUi(self)
+        self.btnPowderCalculateProjection.clicked.connect(self._btn_clicked)
+
+    def set_main_window(self, main_window):
```
martyngigg: What's the reason that the widget needs to know about the main window? What if it were to be used by another widget that does not have access to a main window?

EltayebAhmed: The main window must implement mainview. Any information flowing between tabs goes via the the mainviews presenter

widgets/projection/powder/powder.py
EltayebAhmed: No problem doing that. 
However the widget is not in a usable state (no presenter) until someone calls set_main_window with something that implements main view

models/projection/powder/mantid_projection_calculator.py
```diff
@@ -1,22 +1,11 @@
...
 
 
 class MantidProjectionCalculator(ProjectionCalculator):
     def calculate_projection(self, input_workspace, output_workspace, qbinning, axis1, axis2):
-        if axis1 == 'Energy' and axis2 == '|Q|':
-            #TODO is EMode always direct ?
-            SofQWNormalisedPolygon(InputWorkspace=input_workspace, OutputWorkspace=output_workspace,
-                                                                QAxisBinning=qbinning,EMode='Direct')
+        if axis1 == '|Q|' and axis2 == 'Energy':
+            ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=input_workspace+'_MD',QDimensions='|Q|')
```
martyngigg: I think it would be good to pull out the suffix string, `_MD`, as a module-level variable so that it is clearer where it would be changed, e.g.
```python
MD_SUFFIX = '_MD'
```
above the class definition similar to this type of thing: https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/backends/backend_qt5.py#L73

martyngigg:
Once we make the change to pull out the suffix string as a module variable then I think this is ready to merge.

models/projection/powder/mantid_projection_calculator.py
```diff
@@ -1,22 +1,14 @@
...
-from mantid.simpleapi import SofQWNormalisedPolygon,mtd
-
+from mantid.simpleapi import ConvertToMD
 from models.projection.powder.projection_calculator import ProjectionCalculator
 
+MD_SUFFIX = '_MD'
+
 
 class MantidProjectionCalculator(ProjectionCalculator):
     def calculate_projection(self, input_workspace, output_workspace, qbinning, axis1, axis2):
```
martyngigg: I've just noticed we have an `output_workspace` variable that is not used. Does that need to be removed?

EltayebAhmed:
@martyngigg  Done it.

martyngigg:
@EltayebAhmed Thanks. There's just the comment about the `ouput_workspace` variable left...

models/projection/powder/mantid_projection_calculator.py
EltayebAhmed: The I was told i would be given a naming convention for the projections, I will ask about that today. 
it may be needed in the future

martyngigg:
@EltayebAhmed It looks good now but the coveralls coverage has decreased.

martyngigg:
:+1:

PR #17, opened 2016-08-16
Implement MDworkspace slicing

### LINKED ISSUE #12: Support Plotting of 2D MDworkspace in slice viewer

### LINKED ISSUE #30: Implement Input validation ,autocomplete and error display for slicing
If no information is supplied the slice plotter should display a slice spanning the whole data set
If invalid data is supplied the slice is supplied the signal should emit an error signal.

EltayebAhmed:
**To test:**
Load an MD workspace into mslice
specify all 3 binning parameters for for both x and y in slice tab
Then click display slice

Fixes #12 
Fixes #30

### DIFF
--- a/models/slice/mantid_slice_algorithm.py
+++ b/models/slice/mantid_slice_algorithm.py
@@ -36,7 +36,7 @@ class MantidSliceAlgorithm(SliceAlgorithm):
             self._workspace_provider.delete_workspace(slice)
 
         else:
-            raise NotImplementedError('Implement Using Rebin2D')
+            raise NotImplementedError('Slicing workspace2d is not implemented')
             plot_data = []
             for i in range(workspace.getNumberHistograms()-1,-1,-1):
                 plot_data.append(workspace.readY(i))
@@ -56,6 +56,6 @@ class MantidSliceAlgorithm(SliceAlgorithm):
 
     def _fill_in_missing_input(self,axis,workspace):
         if axis.end is None or axis.start is None or axis.step is None:
-            raise NotImplementedError()
+            raise NotImplementedError("Autocompletion of slice params is not completed")
 
 
--- a/models/slice/mantid_slice_algorithm.py
+++ b/models/slice/mantid_slice_algorithm.py
@@ -56,7 +56,6 @@ class MantidSliceAlgorithm(SliceAlgorithm):
         return int(max(1, floor(axis.end - axis.start)/axis.step))
 
     def _fill_in_missing_input(self,axis,workspace):
-        """Deduct Values not supplied by user from workspace"""
-        pass
+        raise NotImplementedError()
 
 
--- a/models/slice/mantid_slice_algorithm.py
+++ b/models/slice/mantid_slice_algorithm.py
@@ -1,0 +1,62 @@
+from slice_algorithm import SliceAlgorithm
+from mantid.simpleapi import BinMD
+from mantid.api import IMDEventWorkspace
+from models.workspacemanager.mantid_workspace_provider import MantidWorkspaceProvider
+from math import floor
+import numpy as np
+
+
+class MantidSliceAlgorithm(SliceAlgorithm):
+    def __init__(self):
+        self._workspace_provider = MantidWorkspaceProvider()
+
+    def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing):
+        workspace = self._workspace_provider.get_workspace_handle(selected_workspace)
+        if isinstance(workspace,IMDEventWorkspace):
+            # TODO implement axis swapping
+            # TODO implement input validation and return appropriate error codes
+            # Deduct values not supplied by user from workspace
+            self._fill_in_missing_input(x_axis, workspace)
+            self._fill_in_missing_input(y_axis, workspace)
+
+            n_x_bins = self._get_number_of_steps(x_axis)
+            n_y_bins = self._get_number_of_steps(y_axis)
+            x_dim = workspace.getDimension(0)
+            y_dim = workspace.getDimension(1)
+            xbinning = x_dim.getName() + "," + str(x_axis.start) + "," + str(x_axis.end) + "," + str(n_x_bins)
+            ybinning = y_dim.getName() + "," + str(y_axis.start) + "," + str(y_axis.end) + "," + str(n_y_bins)
+            slice = BinMD(InputWorkspace=workspace, AxisAligned="1", AlignedDim0=xbinning, AlignedDim1=ybinning)
+            # perform number of events normalization then mask cells where no data was found
+            with np.errstate(invalid='ignore'):
+                plot_data = slice.getSignalArray() / slice.getNumEventsArray()
+            plot_data = np.ma.masked_where(np.isnan(plot_data), plot_data)
+            # The flipud is because mantid plots first row of array at top of plot
+            # rot90 switches the x and y axis to to plot what user expected.
+            plot_data = np.rot90(plot_data)
+            plot_data = np.flipud(plot_data)
+            self._workspace_provider.delete_workspace(slice)
+
+        else:
+            raise NotImplementedError('Implement Using Rebin2D')
+            plot_data = []
+            for i in range(workspace.getNumberHistograms()-1,-1,-1):
+                plot_data.append(workspace.readY(i))
+                x_left = workspace.readX(0)[0]
+                x_right = workspace.readX(0)[-1]
+                y_top = workspace.getNumberHistograms() - 1
+                y_bottom = 0
+
+        return plot_data
+
+    def get_labels(self, workspace, x_axis, y_axis):
+        return 'X', 'Y'
+
+
+    def _get_number_of_steps(self, axis):
+        return int(max(1, floor(axis.end - axis.start)/axis.step))
+
+    def _fill_in_missing_input(self,axis,workspace):
+        """Deduct Values not supplied by user from workspace"""
+        pass
+
+
--- a/models/slice/mantid_slice_algorithm.py
+++ b/models/slice/mantid_slice_algorithm.py
@@ -39,14 +39,6 @@ class MantidSliceAlgorithm(SliceAlgorithm):
 
         else:
             raise NotImplementedError('Slicing workspace2d is not implemented')
-            plot_data = []
-            for i in range(workspace.getNumberHistograms()-1,-1,-1):
-                plot_data.append(workspace.readY(i))
-                x_left = workspace.readX(0)[0]
-                x_right = workspace.readX(0)[-1]
-                y_top = workspace.getNumberHistograms() - 1
-                y_bottom = 0
-
         return plot_data
 
     def get_available_axis(self, selected_workspace):
--- a/models/slice/mantid_slice_algorithm.py
+++ b/models/slice/mantid_slice_algorithm.py
@@ -13,9 +13,6 @@ class MantidSliceAlgorithm(SliceAlgorithm):
     def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing):
         workspace = self._workspace_provider.get_workspace_handle(selected_workspace)
         if isinstance(workspace,IMDEventWorkspace):
-            # TODO implement axis swapping
-            # TODO implement input validation and return appropriate error codes
-            # Deduct values not supplied by user from workspace
             self._fill_in_missing_input(x_axis, workspace)
             self._fill_in_missing_input(y_axis, workspace)
 
@@ -32,7 +29,6 @@ class MantidSliceAlgorithm(SliceAlgorithm):
             with np.errstate(invalid='ignore'):
                 plot_data = slice.getSignalArray() / slice.getNumEventsArray()
             plot_data = np.ma.masked_where(np.isnan(plot_data), plot_data)
-            # The flipud is because mantid plots first row of array at top of plot
             # rot90 switches the x and y axis to to plot what user expected.
             plot_data = np.rot90(plot_data)
             self._workspace_provider.delete_workspace(slice)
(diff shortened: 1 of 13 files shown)


martyngigg:
@EltayebAhmed Does #13 replace this pull request? Can we delete the branch?

EltayebAhmed:
No discovered a bug, I am re opening it now,

martyngigg:
If I specify all 3 slice parameters for both axes then it works fine.

If I leave one of the slice boxes empty then I only receive a Python error message in the terminal. Is there a plan for how we display these errors to users?

martyngigg:
This works well when you use it exactly as you should.

For the cases of bad input, will this be fixed in another issue?

EltayebAhmed:
@martyngigg I intend on doing that once implement #19  so i don't have to change/update it afterwards.
Just opened an issue for it now.
https://github.com/mantidproject/mslice/issues/30

models/slice/matplotlib_slice_plotter.py
```diff
@@ -1,25 +1,97 @@
...
+            # TODO implement axis swapping
+            # TODO implement input validation and return appropriate error codes
+            # Deduct values not supplied by user from workspace
+            self._fill_in_missing_input(x_axis, workspace)
+            self._fill_in_missing_input(y_axis, workspace)
+            error_code = self._validate_input(x_axis,y_axis,intensity_start,intensity_end)
+            if error_code:
+                return error_code
+            # TODO make shown workspaces refresh after this is called
+            n_x_bins = self._get_number_of_steps(x_axis)
```
OwenArnold: I'm not sure about this. There's a lot of massaging of inputs happening outside of our testable zone. There's probably a missing seam in this design.

models/slice/matplotlib_slice_plotter.py
```diff
@@ -1,25 +1,97 @@
...
+            n_y_bins = self._get_number_of_steps(y_axis)
+            x_dim = workspace.getDimension(0)
+            y_dim = workspace.getDimension(1)
+            xbinning = x_dim.getName() + "," + str(x_axis.start) + "," + str(x_axis.end) + "," + str(n_x_bins)
+            ybinning = y_dim.getName() + "," + str(y_axis.start) + "," + str(y_axis.end) + "," + str(n_y_bins)
+            slice = BinMD(InputWorkspace=workspace, AxisAligned="1", AlignedDim0=xbinning, AlignedDim1=ybinning)
```
OwenArnold: Why are we using `BinMD` here and not `CutMD`. I think `CutMD` will treat partial bins correctly. Is this down to a dimensionality issue?

models/slice/matplotlib_slice_plotter.py
```diff
@@ -1,25 +1,97 @@
...
+        return float(x)
+    else:
+        return None
 
 class MatplotlibSlicePlotter(SlicePlotter):
+    def __init__(self):
+        self._workspace_provider = MantidWorkspaceProvider()
+        self._colormaps = ['viridis', 'jet', 'summer', 'winter']
+
     def display_slice(self, selected_workspace, x_axis, y_axis, smoothing, intensity_start, intensity_end, norm_to_one,
```
OwenArnold: I think having a `SlicePlotter` with a `display_slice` is a good design. However this is not just `displaying` the slice it's doing a whole load of computation including manipulating the inputs. I think manipulation of inputs should be done in the presenter or some auxillary method/class in that testable zone.

How about following the `WorkspaceProvider` approach and having a `SlicingAlgorithm` of which your `MantidSlicingAlgorithm` actually runs `BinMD`?

models/slice/matplotlib_slice_plotter.py
EltayebAhmed: @OwenArnold  Yes this is due to dimensionality, Martin said he would have a look at CutMD to see if the issue is resolvable, I am using SliceMD in the mean time

OwenArnold: Ok. That sounds reasonable. We do not want `mslice` to do the job that algorithms should be doing. So manipulation of data should be something that we always push down into algorithms.

models/slice/matplotlib_slice_plotter.py
EltayebAhmed: I am have moved all input validation to the presenter and I am working on the SlicingAlgorithm refactor now

models/slice/slice_algorithm.py
```diff
@@ -0,0 +1,8 @@
+
+
+class SliceAlgorithm(object):
+
+    def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing):
```
OwenArnold: Good

models/slice/matplotlib_slice_plotter.py
```diff
@@ -1,25 +1,22 @@
...
-
-def get_aspect_ratio(workspace):
-    return 'auto'
+from mantid_slice_algorithm import MantidSliceAlgorithm
+from slice_plotter import SlicePlotter
 
 
 class MatplotlibSlicePlotter(SlicePlotter):
+    def __init__(self):
+        self._slice_algorithm = MantidSliceAlgorithm()
```
OwenArnold: I would inject this rather than hard-coding it.

EltayebAhmed:
@martyngigg  is this ready for merging?

EltayebAhmed:
@martyngigg  is this ready for merging now?

models/slice/mantid_slice_algorithm.py
```diff
@@ -0,0 +1,80 @@
...
+                plot_data = slice.getSignalArray() / slice.getNumEventsArray()
+            plot_data = np.ma.masked_where(np.isnan(plot_data), plot_data)
+            # The flipud is because mantid plots first row of array at top of plot
+            # rot90 switches the x and y axis to to plot what user expected.
+            plot_data = np.rot90(plot_data)
+            self._workspace_provider.delete_workspace(slice)
+
+        else:
+            raise NotImplementedError('Slicing workspace2d is not implemented')
+            plot_data = []
```
martyngigg: We should remove the unreachable code or it could cause confusion

presenters/slice_plotter_presenter.py
```diff
@@ -1,13 +1,27 @@
...
 from models.slice.slice_plotter import SlicePlotter
 from views.slice_plotter_view import SlicePlotterView
 from widgets.slice.command import Command
 from validation_decorators import require_main_presenter
 from interfaces.slice_plotter_presenter import SlicePlotterPresenterInterface
 from interfaces.main_presenter import MainPresenterInterface
 
-Axis = namedtuple('Axis', ['units', 'start', 'end', 'step'])
+
+class Axis:
```
martyngigg: This should be a new-style class

models/slice/matplotlib_slice_plotter.py
```diff
@@ -1,25 +1,26 @@
...
-
-
-
-
+            plot_data = self._slice_algorithm.compute_slice(selected_workspace, x_axis, y_axis, smoothing)
+            boundaries = [x_axis.start, x_axis.end, y_axis.start, y_axis.end]
+            if norm_to_one:
+                plot_data = self._slice_algorithm._norm_to_one(plot_data)
+            norm = Normalize(vmin=intensity_start, vmax=intensity_end)
+            plt.imshow(plot_data, extent=boundaries, interpolation='none', aspect='auto', cmap=colourmap, norm=norm)
```
martyngigg: I'm not sure the model is the correct place to perform the display operations? It would seem to me that the we have the part to do with the the data, boundaries etc that belong to a model but the display should be performed by the presenter?

mslice.py
```diff
@@ -6,16 +6,21 @@
...
 
 class MsliceGui(QMainWindow,Ui_MainWindow,MainView):
+
     def __init__(self):
         super(MsliceGui,self).__init__()
         self.setupUi(self)
 
         workspace_presenter = self.wgtWorkspacemanager.get_presenter()
         slice_presenter = self.wgtSlice.get_presenter()
         powder_presneter = self.wgtPowder.get_presenter()
```
martyngigg: There is a minor spelling error here `powder_presneter` -> `powder_presenter`

presenters/slice_plotter_presenter.py
```diff
@@ -27,6 +42,8 @@ def __init__(self, slice_view, slice_plotter):
         self._slice_view = slice_view
         self._main_presenter = None
         self._slice_plotter = slice_plotter
+        colormaps = self._slice_plotter.get_available_colormaps()
```
martyngigg: Good style here using the modelto provide the data that is pushed in to the view.

models/slice/matplotlib_slice_plotter.py
EltayebAhmed: @martyngigg  so the actual call to the plotting library should be in the presenter?

martyngigg: I think so yes. You had a good example in `slice_plotter_presenter.py` where you pull the list of colour maps from the model and then push them to the view with the presenter.

EltayebAhmed: would it be ok to import the plotting library inside the function call ? otherwise running the unittests will require pyqt @martyngigg

EltayebAhmed: is passing the plotting method into the presenters contructor a reasonable workaround

martyngigg: I think I'd be happy with the plotting method passed in as a parameter. That way it can be mocked for the test?

tests/slice_plotter_presenter_test.py
```diff
@@ -17,34 +17,35 @@ def setUp(self):
         # Setting up view for slice plotter window
         self.slice_plotter = mock.create_autospec(SlicePlotter)
         self.slice_view = mock.create_autospec(SlicePlotterView)
+        self.plotting_module = mock.Mock(spec=["imshow","xlabel","ylabel"])
```
EltayebAhmed: This highlights there is no formally defined interface for the module (i am passing in the module because I need 4 functions, and those would be cumbersome to pass is as arguments)
Is this ok? 
Should I create a class containing a subset of the api needed? 
Or would passing in a tuple of parameters suffice?

martyngigg: I think the module can be our interface for now. It allows us to mock out the functions so should be sufficient for testing

EltayebAhmed:
@martyngigg  is this ready for merging?

models/slice/mantid_slice_algorithm.py
martyngigg: There shouldn't be code underneath the exception being raised. It is not clear whether the code is not meant to be there or the exception was accidentally left in place. Can we remove it, pleaes/

martyngigg:
Apart from the point about the doe left around after the exception I think the code looks good.

I'll do some functional testing now.

martyngigg:
I've found an issue that is worth addressing now:
* load a file
* calculate the projection
* select the `_md` workspace
* click display without filling in the parameters
* an error is displayed in the status bar but an empty plot window is generated also

martyngigg:
In addition:
* just filling out the slice parameters still produces an empty plot

Should the intensity be required? I would think we can use the default range of the data for that. Can we pre-populate these fields based on the workspace?

martyngigg:
I think "Norm to 1" should be a check box not a radio button.

martyngigg:
I can't seem to provide any numbers to get it to display a slice. It just keeps saying "invalid parameters for the intensity of the slice"

EltayebAhmed:
@martyngigg which file are you testing with? If no params are supplied it should display a slice with a 100 bins along each axis , each axis spanning exactly the data, and The intensity should go from min(data) to max(data)

EltayebAhmed:
The problem is that you probably have some white space in the intensity boxes (doing so will cause it fail on machine).  I will push a fix for that now

presenters/slice_plotter_presenter.py
```diff
@@ -19,74 +33,138 @@
...
+                                                                                      smoothing, intensity_start,
+                                                                                      intensity_end, norm_to_one,
+                                                                                      colourmap)
+        except NotImplementedError as e:
+            self._slice_view.error(e.message)
+            return
+        try:
+            self._plot_module.imshow(plot_data, extent=boundaries, cmap=colourmap, aspect='auto', norm=norm,
+                                interpolation='none')
+        except ValueError:
```
EltayebAhmed: Apparently matplotlib is throwing a value error
I thought the only case that would happen was if intensity_min> data_max
could you send me the file you are testing with?
@martyngigg

martyngigg:
Just for reference it turned out that I had matplotlib < 1.5 so I did not have the `viridis` colourmap.

Upgrading to mpl 1.5 solved the issue.

Once we implement the features of pre-populating the range text boxes then I think this is done

EltayebAhmed:
@martyngigg  The reason for the existence for a MatplotlibSlicePlotter and a MantidSlice algorithm is because @OwenArnold  suggested it to keep the Mantid stuff and the Matplotlib stuff seperate, now the plotting is done in the presenter, is there any need for two modules? The MatplotlibSlicePlotter is turning into a (needless) wrapper around the MantidSliceAlgorithm

EltayebAhmed:
@martyngigg  auto-populating the intensity boxes would require me to convert to an MDEventsHistogram workspace and carry out numEventsNormalization, Doing so is kind of costly would it be worth caching it? 
Is there a cheaper way of doing this?

EltayebAhmed:
@martyngigg  will now refuse to plot if any of the 6 axis parameters are not specified

martyngigg:
Thanks. Just for information, there's no need to mention the username after the initial mention as I become a participant and will be notified about changes.

It looks like the CI build has failed.

martyngigg:
We're nearly there now. Just the test coverage to sort out and it's ready to merge.

EltayebAhmed:
yes i almost finished

EltayebAhmed:
done

martyngigg:
:+1:

PR #31, opened 2016-08-18
Implement workspace selection changed notification system.

### LINKED ISSUE #19: Broadcast selected workspace
When the selected workspace changes the workspace manager must broadcast the change to all subscribed listeners. This will allow any tab to repopulate its combo boxes with options more suitable for the selected workspace if needed.
This is currentl needed by the slice and cut tabs

### LINKED ISSUE #33: Test Coverage
The main view presenter has a test coverage much less than the threshold 80%. The main view presenter should have very good coverage based on how important it is to the operation of mslice.

Also, can you start slowly increasing the test coverage limits. Raise it as you improve the test coverage.

EltayebAhmed:
Implements a system that when the currently selected workspace is changed will notify all subscribed 'listeners'
This allows things like the tabs to update the contents of its combo boxes according to the currently selected worskpace

**To test:**

Load (preferably multiple workspaces) to the work space manager. Any changes to the currently selected worskpaces will be printed to the console. This print call is from a subscribed listener (the slice presenter)

Fixes #19 .
Fixes #33

### DIFF
--- a/mslice.py
+++ b/mslice.py
@@ -9,10 +9,10 @@ class MsliceGui(QMainWindow,Ui_MainWindow,MainView):
     def __init__(self):
         super(MsliceGui,self).__init__()
         self.setupUi(self)
+        self.wgtWorkspacemanager.set_client_handler(self)
         workspace_presenter = self.wgtWorkspacemanager.get_presenter()
         self._presenter = MainPresenter(self,workspace_presenter)
 
-        self.wgtWorkspacemanager.set_main_window(self)
         self.wgtSlice.set_workspace_selector(self)
         self.wgtPowder.set_workspace_selector(self)
 
--- a/mslice.py
+++ b/mslice.py
@@ -9,11 +9,12 @@ class MsliceGui(QMainWindow,Ui_MainWindow,MainView):
     def __init__(self):
         super(MsliceGui,self).__init__()
         self.setupUi(self)
-        self._presenter = MainPresenter(self)
 
-        self.wgtWorkspacemanager.get_presenter().register_master(self)
-        self.wgtSlice.get_presenter().register_master(self)
-        self.wgtPowder.get_presenter().register_master(self)
+        workspace_presenter = self.wgtWorkspacemanager.get_presenter()
+        slice_presenter = self.wgtSlice.get_presenter()
+        powder_presneter = self.wgtPowder.get_presenter()
+
+        self._presenter = MainPresenter(self, workspace_presenter, slice_presenter, powder_presneter)
 
     def get_presenter(self):
         return self._presenter
--- a/mslice.py
+++ b/mslice.py
@@ -9,12 +9,14 @@ class MsliceGui(QMainWindow,Ui_MainWindow,MainView):
     def __init__(self):
         super(MsliceGui,self).__init__()
         self.setupUi(self)
-        self.wgtWorkspacemanager.set_client_handler(self)
-        workspace_presenter = self.wgtWorkspacemanager.get_presenter()
-        self._presenter = MainPresenter(self,workspace_presenter)
+        self._presenter = MainPresenter(self)
 
-        self.wgtSlice.set_workspace_selector(self)
-        self.wgtPowder.set_workspace_selector(self)
+        self.wgtWorkspacemanager.get_presenter().register_master(self)
+        self.wgtSlice.get_presenter().register_master(self)
+        self.wgtPowder.get_presenter().register_master(self)
+
+    def get_presenter(self):
+        return self._presenter
 
 if __name__ == "__main__":
     qapp = QApplication([])
--- a/presenters/main_presenter.py
+++ b/presenters/main_presenter.py
@@ -1,4 +1,4 @@
-class MainPresenter:
+class MainPresenter(object):
     def __init__(self,main_view,workspace_presenter):
         self._mainView = main_view
         self._workspace_presenter = workspace_presenter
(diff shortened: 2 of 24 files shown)


presenters/main_presenter.py
```diff
@@ -1,7 +1,8 @@
-class MainPresenter():
+class MainPresenter:
```
martyngigg: All classes should now be so-called "new-style classes" and so should always provide a base class. In the "empty" case then classes should inherit from `object`

EltayebAhmed:
This pull request when merged should close the issue #33

martyngigg:
@EltayebAhmed In order for issues to be closed automatically on merging the issue reference has to be in the initial description so can edit that to include the new issue reference?

EltayebAhmed:
This needs to be merged for the slice and cut implementations to be fully implemented.

EltayebAhmed:
@martyngigg Is this ready for merging ?
I checked the unit tests and all of them are already using mock.create_spec to mock out other presenters needed for tests, Is it still ne

mslice.py
```diff
@@ -9,12 +9,14 @@ class MsliceGui(QMainWindow,Ui_MainWindow,MainView):
...
         super(MsliceGui,self).__init__()
         self.setupUi(self)
-        workspace_presenter = self.wgtWorkspacemanager.get_presenter()
-        self._presenter = MainPresenter(self,workspace_presenter)
+        self._presenter = MainPresenter(self)
 
-        self.wgtWorkspacemanager.set_main_window(self)
-        self.wgtSlice.set_workspace_selector(self)
-        self.wgtPowder.set_workspace_selector(self)
+        self.wgtWorkspacemanager.get_presenter().register_master(self)
```
martyngigg: This isn't quite right. It should be the presenters that are registered and not the view and `self` here refers to the view. 

How about passing the 3 presenters, manager, slice & powder, to the `MainPresenter` constructor and then having that call `register_master(self)`?

presenters/main_presenter.py
```diff
@@ -12,4 +12,23 @@ def update_displayed_workspaces(self):
...
         workspace"""
-        self._workspace_presenter.update_displayed_workspaces()
\ No newline at end of file
+        self._workspace_presenter.update_displayed_workspaces()
+
+    def broadcast_selection_changed(self):
+        for listener in self._selected_workspace_listener:
+            listener.workspace_selection_changed()
+
+    def notify_selection_changed(self):
```
martyngigg: This is good

presenters/powder_projection_presenter.py
```diff
@@ -1,25 +1,28 @@
...
-        if not isinstance(main_view,MainView):
-            raise TypeError("main_view is not of type MainView")
 
-        self._main_view = main_view
         #Add rest of options
         self._powder_view.populate_powder_u1(['|Q|'])
         self._powder_view.populate_powder_u2(['Energy'])
+        self._main_presenter = None
+
+    def register_master(self, main_view):
```
martyngigg: I think this argument should be the `presenter` not the `view`. @OwenArnold, do you agree?

presenters/validation_decorators.py
```diff
@@ -0,0 +1,11 @@
+from functools import wraps
+
+
+def require_main_presenter(function):
```
martyngigg: This is nice.

mslice.py
EltayebAhmed: ok

presenters/validation_decorators.py
EltayebAhmed: Thanks

EltayebAhmed:
@martyngigg  ready now?

martyngigg:
Code wise this looks good now. Unfortunately I can't seem to see the notification messages when I load and select the different workspaces.

EltayebAhmed:
@martyngigg 
It works on my machine and I have no idea why, do have any idea what could be the issue?

martyngigg:
@EltayebAhmed It's not obvious why it is not working. I have checked out `master` and merged your branch in locally. Do you perhaps have some other changes mixed in?

EltayebAhmed:
@martyngigg  I don't think so,
That would show up on `git status` if that were the case right?

EltayebAhmed:
@martyngigg 

`git merge master` reports 
`Already up-to-date`

`git status` reports
 ```
On branch 19_broadcast_change_in_selected_workspaces
nothing to commit, working directory clean
```

martyngigg:
If other branches were merged in the `git status` would still show as clean.

I have figured out the problem and I can see the messages now. I wrongly thought that just highlighting the workspaces was enough trigger the notification and I didn't realise I had to check the boxes. Checking/unchecking the boxes correctly emits the messages now.

I think this points to a usability issue here where you can highlight an item but the selection status doesn't change. Do we need the check boxes? Can we just do use the highlight state? This is how MantidPlot works and may be more intuitive?

EltayebAhmed:
@martyngigg  Setting checkboxes allows for multiple workspaces to be selected, And it would confuse users to if we use highlighted selection as a fallback if none are,
Should I open and issue and axe the checkboxes now or should I wait for user feedback on that?

martyngigg:
You can implement multiple selections without check boxes, we do this in the MantidPlot dock widget.

As the current implementation works I think we should leave these set of changes as is but open another issue to remove the check boxes and just use the highlight state and allow `ctrl+click` for multiple selections: http://doc.qt.io/qt-4.8/qabstractitemview.html#selectionMode-prop

PR #41, opened 2016-08-23
Implement Cut Tab

### LINKED ISSUE #27: Implement Cut
Implement the cut model for the cut . The user should be able to take cuts for the currently selected workspace

### LINKED ISSUE #44: Plot cuts with error bars
Cuts should be plotted with error bars. The option to turn them off should be in the plot options pop-up box.

### LINKED ISSUE #45: Cut naming schemes
cuts should be generated with names that follow
< original_workspace_name>_cut _< index>
where index starts at 1 and increments to prevent name conflicts

### LINKED ISSUE #46: Add the ability to plot cuts that have saved to workspaces
The input fields of the cut box should be autopopulated to the information contained in the cut but greyed out. plot and plot over should behave as if the original event workspace was selected with the values in the input field selected.

### LINKED ISSUE #47: Add cut legends
The cut legend should be the workspace name + range the range and axis of the cut i.e
MAR232422   -2<e<1

EltayebAhmed:
Description of work.

**To test:**
Select an MD workspace in the worskpace manager,  fill in the required parameters for the cut and click plot/ plot_over should bring up plots

clicking on save cut to workspace will save cuts to workspaces and show them in workspace manager. Selecting any one of these cuts will disable the cut tab except plot/plot over and intensity range and the show the cuts parameters.

Specifying a width will divide the integration range into blocks of specified width and will then cause multiple cuts to be saved/plotted

Fixes #27 
Fixes #45 
Fixes #44 
Fixes #47 
Fixes #46


EltayebAhmed:
Currently, only plot/plot_over are supported, intensity ranges are ignored. This is WIP.

EltayebAhmed:
@martyngigg this branch is now ready for review

martyngigg:
It's good that this is ready to go.

For future reference can we try an avoid fixing many issues in a single PR. 1 or 2 is generally okay but 6 starts to feel like too much work for one PR - it get's hard to review this way.

martyngigg:
My first test case seems to cause a looping call to `BinMD`:
* load `MAR21335_Ei60.00meV.nxs`
* click calculate projection to convert to an MD workspace
* select the cut tab
* enter
  * 0-12 in step 0.1 along cut axis
  * 0-30 in step 0.1 integration
* click plot -> application locks up and the console prints that a loop of `BinMD` calls is being performed

martyngigg:
I see what is happening now. The integration width is very small and it's trying to generate many plots. If I try something more sensible like a width of 15 then I get two curves as expected.

martyngigg:
I've looked over the code and it's following a similar model to the slice implementation, which is very good. 

Tests have been added and the code coverage has been increased slightly.

martyngigg:
We have just had a demo with the scientists were the cut features were shown and they were happy with the functionality so I will not test any further at this point. It was agreed that we would get out a pilot version for scientists to use.

PR #54, opened 2016-08-31
Refactored workspace provider widget to emit errors as QSignals

### LINKED ISSUE #24: Refactor widgets to emit error messages as signals
Currently the QWidgets attempt to display error messages themselves by directly passing them on to the main windows status.
This behaviour should be changed to the widgets  emitting the error strings as a QSignal to increase code ease of code reuse

EltayebAhmed:
Refactored workspacemanager widget to emit errors as signals rather than to handle them on the widgets

**To test:**
Open MSlice and make the workspacemanager show an error on the statusbar.
For example click Remove/Save/Rename without having a workspace selected

Fixes #24


martyngigg:
This is a good decoupling of behaviour so now the host container for the workspace list can decide what to do with the errors.

Functionally it all checks out too and the user still sees the appropriate errors.

PR #59, opened 2016-09-02
Startup scripts

### LINKED ISSUE #58: Create startup Scripts
Create startup scripts that will start up mslice and accompanying terminal

EltayebAhmed:
Added startup Scirpts and Instructions to run them

**To test:**
Read the `readme.md` and try to follow the instructions to get MSlice up and running.
Fixes #58


martyngigg:
It was agreed on slack that these are in a good enough state for scientific testing.

Future work will need a different approach to startup and packaging but that is outside of the scope of current work.

PR #68, opened 2016-09-08
Refactor figure manager

### LINKED ISSUE #56: Refactor Figuremanager.py to work out kinks that arise when mixing cli use and gui use
Inconsistent behaviour is observed when using the `pyplot.figure` while using the mslice gui.
The Figuremanager also has alot of redudant code to a misunderstanding during implementation . It now currently keeps track of `kept` , `current` and `active` even though the latter two are equivelant

EltayebAhmed:
Refactored figure manager which keeps track of current and kept figures. It should now be more robust and should make getting a working cli interface easier.

**To test:**
Start mslice and verify that keep and make current behave as expected.

Fixes #56


plotting/figuremanager.py
```diff
@@ -1,8 +1,57 @@
-from __builtin__ import staticmethod
-
+# This is the class responsible for handling which figures are current (will receive plotting commands) and which are
```
martyngigg: Great documentation!

plotting/figuremanager.py
```diff
@@ -1,8 +1,57 @@
...
+#
+
+
+from functools import wraps
 from plotting import get_figure_class
-PlotFigureManager = get_figure_class( )
-NO_FIGURE = -1723
+
+
+class NoFigure(object):
```
martyngigg: Good use of a null object type of pattern

tests/figure_manager_test.py
```diff
@@ -117,3 +117,66 @@ def test_close_non_existant_window_fail(self,mock_figure_class):
 
         self.assert_(fig1 == mock_figures[0])
         self.assert_(fig2 == mock_figures[0])
+
+    @mock.patch(plot_window_class)
```
martyngigg: Good testing here. It's nice that the coverage is incrementally improving

PR #83, opened 2016-09-22
Remove misleading references of energy from mslice.

### LINKED ISSUE #82: Remove misleading use of 'Energy' from mslice
The GUI in mslice refers to Energy and DeltaE interchangably. This is misleading as they are different scientific concepts and should not be confused.

EltayebAhmed:
In the projections tab the term energy was used an input even though output workspace was in DeltaE vs |Q|. Although the resulting plots did have correct legends this may become a source of confusion in MSlice if more units are added.

**To test:**
1. Open Mslice. 
2. Verify that the projections tab dropdown menu has a `|Q|` entry and a `DeltaE` entry and that is does not have an `Energy` entry.

Fixes #82

### DIFF
--- a/widgets/projection/powder/powder.ui
+++ b/widgets/projection/powder/powder.ui
@@ -33,33 +33,7 @@
          </widget>
         </item>
         <item row="0" column="1">
-         <widget class="QComboBox" name="cmbPowderU1">
-          <item>
-           <property name="text">
-            <string>Energy</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
-            <string>|Q|</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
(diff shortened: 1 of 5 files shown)


widgets/projection/powder/powder.ui
```diff
@@ -33,33 +33,7 @@
...
-           <property name="text">
-            <string>Det Group Number</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
-            <string>Time Channel</string>
-           </property>
-          </item>
-         </widget>
```
EltayebAhmed: The .ui file was found to contain a list of initial entries. On contruction the presenter populates the drop downs with values stored in the presenter, effectively making this *Dead Code*

PR #86, opened 2016-09-22
Fix grouping workspaces

### LINKED ISSUE #85: Group workspaces does not work
Group workspaces does not work. Selecting multiple workspaces (via `ctrl` + `click`) then clicking on group workspaces will cause an exception to be raised

EltayebAhmed:
The Group workspaces was not working. This pull requests fixes it

**To test:**
1.Load multiple workspaces into mslice
2. Select them all via using `ctrl` + `click`
3. Click on the `Group Selected` button in the **Workspace Manager**
4. Verify that a new group has appeared.

Fixes #85


mducle:
Moved commits to #91 to get rid of docs commits.

PR #91, opened 2016-09-29
Fix group workspaces

### LINKED ISSUE #85: Group workspaces does not work

mducle:
The Group workspaces was not working. This pull requests fixes it

**To test:**
1.Load multiple workspaces into mslice
2. Select them all via using `ctrl` + `click`
3. Click on the `Group Selected` button in the **Workspace Manager**
4. Verify that a new group has appeared.

Fixes #85

### DIFF
--- a/models/workspacemanager/mantid_workspace_provider.py
+++ b/models/workspacemanager/mantid_workspace_provider.py
@@ -16,7 +16,7 @@ class MantidWorkspaceProvider(WorkspaceProvider):
         return Load(Filename=Filename, OutputWorkspace=OutputWorkspace)
 
     def group_workspaces(self, InputWorkspaces, OutputWorkspace):
-        return GroupWorkspaces(InputWorkspaces,OutputWorkspace)
+        return GroupWorkspaces(InputWorkspaces, OutputWorkspace=OutputWorkspace)
 
     def rename_workspace(self, selected_workspace, newName):
         return RenameWorkspace(selected_workspace,newName)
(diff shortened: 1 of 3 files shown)


martyngigg:
I can see that the grouping does now create a new group but the "member" workspaces are also still in the list, which seems confusing. There's nothing to tell me at a glance that operations on the new group will do anything different to clicking one of the other entries.

Is the grouping behaviour still a required feature? If so I would say we need to look at improving the display of them first. What do you think?

mducle:
I think that if we can do operations on multiple workspaces without grouping them in the Mantid sense, then groups will not be necessary - e.g. in the MSlice GUI, if you shift-select several workspaces and click "Calculate Projection" then it operates on all the workspaces (by looping over them and applying the algorithm to each in turn). I guess this is the same as what group workspace does but seems to me to be more flexible (e.g. if you just want to create the same cuts for a set of workspaces but then want to do different things to each of the cuts, you don't have to first group the workspaces, create the cuts and then ungroup the workspaces and cuts). However, this doesn't work for all the use cases at the moment (e.g. cuts / slices of workspaces with the same axes).

If we go with this option, we can then just remove the "Group Selected" button.

martyngigg:
@mducle Personally I would prefer your second option of multiple selections too - I think it is more intuitive based on how many other GUIs would function and I don't see why we can't make it work.

Is this something worth putting on the list for the new developer to look at? It could be a good way for them to get into the code a bit more.

mducle:
Ok. Then let's remove the "Group Selected" button and the grouping functionality, and do multiple operations on multiply selected workspaces. Some of it's already in the project list, so yeah, it would be really good if the new developer can have a look at it.

martyngigg:
@mducle I'll close this pull request and open another to remove the grouping behaviour.

PR #93, opened 2016-09-30
Implement wavenumber conversion

### LINKED ISSUE #63: Implement energy unit conversions
In the calculate projections tab there is an option for different energy units. This is one of the key requirements of the Molecular Spectroscopy group because they tend to use *wavenumbers* `cm-1` rather than `meV`. 

The conversion factor is 1meV = 8.06573cm-1

mducle:
Added wavenumbers `cm-1` as a possible energy unit. 

**To test:**

Load a dataset and select `cm-1` from the `units` combo box, and calculate projection. Check that the resulting MD workspace has energy which is scaled by 8.066 compared to the MD workspace created with `units=meV`. The new workspaces should have a `_cm` suffix.

Fixes #63.

### DIFF
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -21,15 +21,14 @@ class MantidProjectionCalculator(ProjectionCalculator):
     def calculate_projection(self, input_workspace, axis1, axis2, units):
         """Calculate the projection workspace AND return a python handle to it"""
         emode = self._workspace_provider.get_workspace_handle(input_workspace).getEMode().name
-        if emode == 'Elastic': emode = 'Direct'
         if axis1 == MOD_Q_LABEL and axis2 == DELTA_E_LABEL:
             scale = [1, 8.06554]
-            output_workspace = input_workspace + '_QE'
+            output_workspace = input_workspace + '_QE' + ('_cm' if units == WAVENUMBER_LABEL else '')
             ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=output_workspace, QDimensions=MOD_Q_LABEL,
                         PreprocDetectorsWS='-', dEAnalysisMode=emode)
         elif axis1 == DELTA_E_LABEL and axis2 == MOD_Q_LABEL:
             scale = [8.06554, 1]
-            output_workspace = input_workspace + '_EQ'
+            output_workspace = input_workspace + '_EQ' + ('_cm' if units == WAVENUMBER_LABEL else '')
             ConvertToMD(InputWorkspace=input_workspace, OutputWorkspace=output_workspace, QDimensions=MOD_Q_LABEL,
                         PreprocDetectorsWS='-', dEAnalysisMode=emode)
             output_workspace_handle = self._workspace_provider.get_workspace_handle(output_workspace)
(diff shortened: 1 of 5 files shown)


models/projection/powder/mantid_projection_calculator.py
```diff
@@ -33,7 +40,13 @@ def calculate_projection(self, input_workspace, axis1, axis2):
...
+            SliceMD(InputWorkspace=output_workspace, OutputWorkspace=output_workspace, AlignedDim0=dim0,
+                    AlignedDim1=dim1)
         else:
             raise NotImplementedError("Not implemented axis1 = %s and axis2 = %s" % (axis1, axis2))
+        # Now scale the energy axis if required - ConvertToMD always gives DeltaE in meV
+        if units == WAVENUMBER_LABEL:
+            TransformMD(InputWorkspace=output_workspace, OutputWorkspace=output_workspace, Scaling=scale)
+            self._workspace_provider.get_workspace_handle(output_workspace).setComment('MSlice_in_wavenumber')
+        elif units != MEV_LABEL:
+            raise NotImplementedError("Unit %s not recognised. Only 'meV' and 'cm-1' implemented." % (units))
```
martyngigg: Is this method missing a `return` statement. The previous version returned from the `ConvertToMD`/`SliceMD` call but nothing is returned here.

I thinks this only affects the `cli`. If I run
```python

import cli
w1 = cli.Load('MAR21335_Ei60.00meV.nxs')
w2 = cli.get_projection(w1, '|Q|', 'DeltaE')
```

I get an error:
```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-db82bf2cbb25> in <module>()
----> 1 w2 = cli.get_projection(w1, '|Q|', 'DeltaE')

/mnt/data1/source/github/mantidproject/mslice/cli/_mslice_commmands.py in get_projection(input_workspace, axis1, axis2, units)
     83     if len(names) > 1:
     84         raise Exception('Too many left hand side arguments, %s' % str(names))
---> 85     RenameWorkspace(InputWorkspace=output_workspace, OutputWorkspace=names[0])
     86     return output_workspace
     87 

/opt/mantidnightly/bin/mantid/simpleapi.pyc in RenameWorkspace(InputWorkspace,[OutputWorkspace],[True||False], ****kwargs)
    490     _set_logging_option(algm, kwargs)
    491     for key, val in arguments.items():
--> 492         algm.setProperty(key, val)
    493 
    494     algm.execute()

ValueError: When converting parameter "InputWorkspace": Unable to extract boost::shared_ptr<Workspace> from Python object
```

The GUI operations are fine though.

mducle: Ah oops. Yes. Sorry, I should at a return value. I was only checking it with the GUI.

martyngigg:
I can see the graph axis update to the new scale correctly. The same operations in the `cli` also function as expected.

Should we add unit labels to the axis title? It's not obvious from the plot what scale the axis has.

mducle:
Yes. It would be a good idea to have unit labels in the axis title. 

About refactoring - I think that might not be a bad idea - because another open pull request is to add `Theta` as a possible axis in addition to `|Q|` - and for future, for single crystal mode, we'd also have to add `Q_h`, `Q_k`, `Q_l` or `Q_1`, `Q_2`, `Q_3`.

martyngigg:
@mducle That sounds sensible. I generally like the concept of having a separate method once the body of an `if` clause gets over about 2-3 lines.

martyngigg:
Would you want to do that here or should we get the functionality in so it can be tested?

mducle:
Refactoring should probably be done in conjunction with the `Theta` issue - #94 - which also adds to the projections code. So, maybe if we merge this one, I can modify the other PR to refactor the code there?

martyngigg:
That sounds reasonable.

Issue #38, opened 2016-08-19
Implement error handling for powder projections


EltayebAhmed:
Currently there is no error handling in the projections tab.
Error handling to show an error message if no workspace is selected or the algorithm fails should be implemented.

PR #94, opened 2016-10-01
Theta axis implementation

### LINKED ISSUE #92: Add support for 2Theta (scattering angle) axis
In addition to `|Q|` and `DeltaE`, `Theta` (the scattering angle) should also be an available axis in powder mode in mslice.

This requires modification of the projections functions - to apply a `ConvertSpectrumAxis` before a `ConvertToMD` with the `CopyToMD` method.

mducle:
This PR adds a `2Theta` axis to the list of possible *x*-, or *y*-axes (`u1` or `u2`). The implementation uses `ConvertSpectrumAxis` on the input workspace and the `ConvertToMD` with the `CopyToMD` method. (Note: For indirect geometry instrument data, `PreprocessDetectorsWS` which is run by `ConvertToMD` doesn't work on the output of `ConvertSpectrumAxis` for some reason - probably a bug. However, if it's run first on the original input workspace, then the detector workspace is used in `ConvertToMD`, a valid `MDWorkspace` is created. This work-around is used in this PR).

**To test:**

In the mslice GUI, load a file, and select `2Theta` as `u1` and `DeltaE` as `u2` (or vice versa). Calculate projection and plot the slice - check that the `2Theta` axis is in `Degrees` and the data looks reasonable.

Fixes #92.

### DIFF
--- a/models/projection/powder/mantid_projection_calculator.py
+++ b/models/projection/powder/mantid_projection_calculator.py
@@ -1,9 +1,6 @@
-<<<<<<< HEAD
 import tempfile
-from mantid.simpleapi import ConvertToMD, SliceMD, ConvertSpectrumAxis, PreprocessDetectorsToMD, DeleteWorkspace
-=======
-from mantid.simpleapi import ConvertToMD, SliceMD, TransformMD
->>>>>>> master
+from mantid.simpleapi import ConvertToMD, SliceMD, TransformMD, ConvertSpectrumAxis, PreprocessDetectorsToMD
+from mantid.simpleapi import RenameWorkspace, DeleteWorkspace
 from models.projection.powder.projection_calculator import ProjectionCalculator
 from models.workspacemanager.mantid_workspace_provider import MantidWorkspaceProvider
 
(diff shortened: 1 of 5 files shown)


martyngigg:
I've tested this with the MAR21335 & IRIS26176 test data set and the projections in their various combinations look as expected, i.e the elastic line is in the right place!

I also tried selecting various combinations that should not be allowed and as expected the GUI does not allow this.

There is a more general issue at the moment that if you click "Calculate Projection" on one of the MD workspaces then you see

```
  File "/mnt/data1/source/github/mantidproject/mslice/widgets/projection/powder/powder.py", line 42, in _btn_clicked
    self._presenter.notify(Command.CalculatePowderProjection)
  File "/mnt/data1/source/github/mantidproject/mslice/presenters/powder_projection_presenter.py", line 34, in notify
    self._calculate_powder_projection()
  File "/mnt/data1/source/github/mantidproject/mslice/presenters/powder_projection_presenter.py", line 53, in _calculate_powder_projection
    self._projection_calculator.calculate_projection(workspace, axis1, axis2, units)
  File "/mnt/data1/source/github/mantidproject/mslice/models/projection/powder/mantid_projection_calculator.py", line 69, in calculate_projection
    emode = self._workspace_provider.get_workspace_handle(input_workspace).getEMode().name
AttributeError: 'MDEventWorkspace2D' object has no attribute 'getEMode'
```
in the IPython console but no error on the GUI is shown. This should be rectified by #38.

mducle:
Yes. We should probably sort out the error handling better - a job for Simon I think!

PR #95, opened 2016-10-03
Fixed landscape errors and some warnings.

mducle:
Fixes some errors highlighted by landscape.io

**To test:**

Just check that everything still runs as it should (data loads, can plot slices and make cuts).

No issue number.


plotting/pyplot.py
```diff
@@ -1858,7 +1858,7 @@ def matshow(A, fignum=None, **kw):
         ax = gca()
     else:
         # Extract actual aspect ratio of array and make appropriately sized figure
-        fig = figure(fignum, figsize=figaspect(A))
+        fig = Figure(fignum, figsize=figaspect(A))
```
martyngigg: This was a good thing to catch!

PR #97, opened 2016-10-05
Implemenation of automatic workspace selection after Load and Calculate Projection.

### LINKED ISSUE #96: Streamline GUI workflow by autoselecting output workspaces.
A user has noted that at present, the workflow is a bit convoluted: It involves:
1. Load data
2. Clicking on the loaded workspace in the workspace manager
3. Clicking calculate projection
4. Clicking on the projected workspace in the workspace manager
5. Clicking plot

He suggested to eliminate step 2 and 4 by making the output of `Load` and `Calculate Projection` the currently selected workspace in the workspace manager.

mducle:
This PR changes the GUI behaviour to set the selected workspace(s) in the workspace manager to be the result of the `Load` or `Calculate Projection` actions after they complete.

**To test:**

Run the GUI, and load a data file. On success check that the loaded workspace is selected in the workspace manager. Now click `Calculate Projection` - check the created `_QE` or `_EQ` workspace is now selected. (Also try it with multiple selection for `Calculate Projection` and check that all the created workspaces are selected).

Fixes #96.

Unit tests to follow...


models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -42,3 +42,9 @@ def get_workspace_handle(self, workspace_name):
         if isinstance(workspace_name, Workspace):
             return workspace_name
         return AnalysisDataService[workspace_name]
+
+    def get_workspace_name(self, workspace):
```
martyngigg: To be consistent with other areas we should add a method with this name to the `models/workspace_provider.py` interface

mducle: Right... I didn't because I saw that `get_workspace_handle` didn't have a declaration. I'll just add both these methods to `models/workspacemanager/workspace_provider.py`

martyngigg: That must have slipped through. The problem with Python and not having a compiler!

martyngigg:
Functionally it all checks out too. I see the expected behaviour as described.

PR #110, opened 2016-11-29
Move to setuptools project

### LINKED ISSUE #102: Migrate to a setuptools project
This is the standard way Python projects should be created. See all of the well-known projects such as `numpy`, `scipy`, `matplotlib`.

[spyder-ide](https://github.com/spyder-ide/spyder/blob/master/setup.py) has a nice setup.py as an example to follow. This will involve placing most of the current directories under an `mslice` package.

Other tasks:

* generate `*_ui.py` files as part of the build;
* generate startup scripts that include starting it through IPython
* include flake8 and unit testing and update the Travis configuration
* update README for developers to get started

This can then form the basis for future work to create packages for users: #5

martyngigg:
This work implements the standard setuptools structure to make both development and external packaging easier for the future. Building of the `.ui` files is now automated and a manual build step is not required.

**To test:**

The instructions in the README have been updated. Please read these and try to follow them.

Fixes #102


mducle:
`mslicedevel.bat` works on Windows for me.

`python setup.py install` works on Red Hat 7 for me, but `mslicedevel.sh` gives an error when running the app, because it can't find `mainwindow_ui` in line 5 of `mainwindow.py` - I think the paths are not set correctly for some reason.

martyngigg:
Thanks, @mducle. It looks like there is a slightly different build layout on RHEL/Ubuntu. It should be fixed now.

mducle:
@martyngigg Yep works fine now. :shipit:

PR #111, opened 2016-12-05
Colorbar range and scale type editable 

### LINKED ISSUE #80: Implement option to change colorbar limits in plot option
In the plot options dialog it was requested that it should be possible to change the intensity scale minimum and maximum values (the start and end values for the colorbar). It was also required to add an option to change the colorbar scale to logarithmic.
(relative source files are in `plotting/plot_window`)

simonfernandes:
The colorbar scale minimum and maximum values can now be changed via the plot options dialog's corresponding text boxes.

Additionally, the colorbar scale can also be changed to logarithmic by ticking a new "logarithmic" checkbox in the plot options dialog.

**To test:**

Ensure that the minimum and maximum scale can be adjusted, as well as the colorbar and graph change to logarithmic when selected.

Fixes #80


mducle:
Seems to work ok... 

Except I seemed to have found a very peculiar bug on RH7 (Linux). If I plot a slice, then got to plot options and check `logarithmic` and then press `OK` it closes the main window (no error message is displayed). If I then open it up again with `show_gui()` and try to plot with `logarithmic` unchecked (e.g. normal again), it closes the main window again. This also happens if I just click on `keep` and try to plot another slice from the main window. However, then reopening the main window with `show_gui()` and plotting the slice again, everything works as expected (e.g. the main window no longer closes when you plot `logarithmic`.

On Windows, everything is fine. 

I have no idea why the Linux version is behaving like this - could there be a name collision somewhere?

martyngigg:
I just tried this out on Ubuntu and it seems to behave as expected. I'll try my RHEL 7 VM

martyngigg:
I see the same thing on RHEL7 with `master` too. If I plot a slice, click `Keep` and then plot another slice the main gui disappears.

Something strange is going on there but I don't think it's related to these changes. I wonder if it's GNOME 3 related. I'll try and take a look at this in #112.

martyngigg:
The only thing that I think needs addressing at the moment is that if you try and use plot options if you have plotted a cut then you get an error as there is no image on the canvas:
```
  File "/mnt/data1/source/github/mantidproject/mslice/build/lib.linux-x86_64-2.7/mslice/plotting/plot_window/plot_figure.py", line 51, in _plot_options
    config = self._get_plot_description()
  File "/mnt/data1/source/github/mantidproject/mslice/build/lib.linux-x86_64-2.7/mslice/plotting/plot_window/plot_figure.py", line 182, in _get_plot_description
    mappable = current_axis.get_images()[0]
IndexError: list index out of range
```

mslice/plotting/plot_window/plot_options.py
```diff
@@ -2,13 +2,31 @@
...
 
+
 class PlotOptionsDialog(QtGui.QDialog, Ui_Dialog):
     def __init__(self, current_config):
         super(PlotOptionsDialog, self).__init__()
         self.setupUi(self)
-        self.lneFigureTitle.setText(current_config.title)
-        self.lneXAxisLabel.setText(current_config.xlabel)
-        self.lneYAxisLabel.setText(current_config.ylabel)
+        if self.lneFigureTitle is not None:
```
martyngigg: Is there a reason that we need these `is not None` checks. I think we can safely assume that the Qt designer has done its job here

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -61,6 +62,33 @@ def _apply_config(self, plot_config):
         current_axis.set_xlim(*plot_config.x_range)
         current_axis.set_ylim(*plot_config.y_range)
 
+        images = current_axis.get_images()
+        if len(images) != 1:
```
martyngigg: I think we need a similar `if not current_axis.get_images()` check here.

At the moment If I plot a cut, click plot options, uncheck show error bars and then click OK I get an error about applying the settings

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -61,6 +62,34 @@ def _apply_config(self, plot_config):
         current_axis.set_xlim(*plot_config.x_range)
         current_axis.set_ylim(*plot_config.y_range)
 
+        if current_axis.get_images():
+            images = current_axis.get_images()
+            if len(images) != 1:
+                raise RuntimeError("Expected single image on axes, found " + str(len(images)))
```
martyngigg: There's quite a lot of whitespace here. Can we remove the double-empty lines?

mslice/plotting/plot_window/plot_options.py
```diff
@@ -23,17 +41,11 @@ def __init__(self, current_config):
...
-            self.lneXMax.setText(str(current_config.x_range[1]))
-        if None not in current_config.y_range:
-            self.lneYMin.setText(str(current_config.y_range[0]))
-            self.lneYMax.setText(str(current_config.y_range[1]))
 
     @staticmethod
     def get_new_config(current_config):
         dialog = PlotOptionsDialog(current_config)
-        dialog_accepted= dialog.exec_()
+        dialog_accepted=dialog.exec_()
```
martyngigg: This is really picky but as I have other formatting comments then can we follow pep8 and out a spaces around the `=` here?

mslice/plotting/plot_window/plot_options.py
```diff
@@ -50,6 +62,15 @@ def get_new_config(current_config):
...
 
+        try:
+            cmin = float(str(dialog.lneCMin.text()))
+            cmax = float(str(dialog.lneCMax.text()))
+            colorbar_range = (cmin, cmax)
+        except ValueError:
+            colorbar_range = (None, None)
+
+        logarithmic = dialog.chkLogarithmic.isChecked()
+
```
martyngigg: No need for this blank line I don't think.

martyngigg:
This is good to go now. :+1:

PR #113, opened 2016-12-20
Keep reference to MainWindow alive

### LINKED ISSUE #112: RHEL7: Main window disappears after plotting second slice and keeping old window
While testing #111 it was noted that on RHEL7 there is an issue with the main Mslice window disappearing after certain operations are performed. Calling `show_gui` from IPython seems to resolve the issue but it is unclear what is causing this to happen.

martyngigg:
Description of work.

Fixes lifetime issue with the reference to the `MainWindow` being cleaned up before the program had exited. 

**To test:**

This only appeared to be an issue on RHEL7 with Gnome but it should have been general.

* start mslice with `./mslicedevel`
* load and calculate a projection
* plot a slice and keep the window
* change some slice parameters and plot another slice
* a new window should appear and the main GUI should stay visible.

Before this fix the main mslice window disappeared on RHEL7. 

Fixes #112

### DIFF
--- a/mslice/app/__init__.py
+++ b/mslice/app/__init__.py
@@ -32,4 +32,3 @@ def startup(with_ipython):
         qapp = QApplication([])
         show_gui()
         qapp.exec_()
-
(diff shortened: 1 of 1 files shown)


mducle:
Works, fixes problem, and code looks good. :shipit:

PR #115, opened 2017-01-15
Save ASCII

### LINKED ISSUE #108: Enable `Save Ascii` functionality in cuts.
A `Save Ascii` button exists in the cut tab but currently does not do anything. A very rough implementation is present in the `demo3` branch - this should be cleaned up and have tests written for it.

mducle:
Enables the saving of cuts to text files in a three-column `x`, `y`, `e` format. 

If multiple cuts area selected, each will be saved to a separate file with the same starting prefix.

**To test:**

Open a data file, make a cut and click `Save ASCII` in the `Cut` tab. Select a file name and check that the correct data (and header, indicating the limits of the cut and original workspace name) is written to that file. Select multiple cuts and check that the correct number of files are written with the correct data. 

Fixes #108.

### DIFF
--- a/misc/MDUtils.py
+++ b/misc/MDUtils.py
@@ -10,13 +10,13 @@ def dim2array(d):
     """
     Create a numpy array containing bin centers along the dimension d
     input: d - IMDDimension
-    return: numpy array, from min+st/2 to max-st/2 with step st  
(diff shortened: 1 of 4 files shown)


martyngigg:
This looks fine now.

PR #116, opened 2017-03-03
Remember cut parameters

### LINKED ISSUE #106: Remember cut `param` values for different directions
This basically means storing previous parameters for cuts along `E` and `Q` separately.

mducle:
Stores previously entered parameters for cuts (along which axis, from, to, step, integration range, smoothing level, whether to normalise to unity) in an attribute of `CutPresenter`. When the user changes back to a previously defined cut/workspace pair (or a new workspace with the same axes), will repopulate the fields in the `Cut` tab with the previously defined values.

**To test:**

Load several data files and calculate several sets of projections (convert to MDWorkspace) for each dataset, ensuring that you have at least two of the same type (e.g. two `_QE` and one `_EQ` or whatever). Make some cuts, and check that when you select another workspace or axis, and then select the previous one again, that the correct values for the previous cut reappear.

Fixes #106

### DIFF
--- a/mslice/presenters/cut_presenter.py
+++ b/mslice/presenters/cut_presenter.py
@@ -19,6 +19,9 @@ class CutPresenter(object):
         self._cut_plotter = cut_plotter
         self._acting_on = None
         self._cut_view.disable()
+        self._saved_parameters = dict()
+        self._previous_cut = None
+        self._previous_axis = None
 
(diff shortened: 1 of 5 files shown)


martyngigg:
I tried taking several cuts and switching between them in various scenarios: correct parameters, incorrect parameters and empty parameters. All circumstances behaved as expected.

Issue #114, opened 2017-01-15
Set E/Q mininum bin size


mducle:
E/Q bins should not be set smaller than in original data.

At present they can be set arbitrarily small, which can lead to memory problems, and also means the represented data is not accurate, because bins which have been subdivided will appear as a pixel with counts in the centre and `NAN`s outside.

Ideally, `mslice` should snap to smallest bin in the data file.

mducle:
Duplicate of issue #48. Will be addressed with changes in issue #118

mducle:
Fixed in PR #120

Issue #48, opened 2016-08-24
Improve Energy axis binning suggestion


EltayebAhmed:
Currentl MSlice will come up with suggestions for plotting a slice when the user clicks on a sliceable workspace. The suggestions for the range of an azis are to the span of the data along that axis. The suggestions for the bin size is calculated by `range/100`. For the `|Q|` Axis this is OK. For the energy axis a more intelligent guess can be made. 
To be more specific, energy axis binning should suggested as the coarsest bin in the original data (if this data is not available the workspace it is known as a fact that it is in the .nxs file)

mducle:
Fixed in PR #120

Issue #28, opened 2016-08-18
Warn users before attempting to slice to extreme binnings


EltayebAhmed:
I tried to display a slice with a total number of bins in the range of 10^8
As expected the computer came to total freeze (had to forcefully shutdown from power button).
It would be a good idea to warn the users before trying plot unreasonable (slices/plots)
@martyngigg any suggestions on the number of bins where we should start trying to warn the user?

EltayebAhmed:
@martyngigg  any input? I think should be addressed before people start installing and testing mslice on there machines.

martyngigg:
I agree that it makes sense to warn users of the danger. It's probably worth doing a couple of tests to see on your own machine where the choke point is. I would hazard a guess that something like warning if the number of bins > 10^5 would be more than enough.

mducle:
After discussions with Ross and Helen, we think the best approach is if `mslice` can detect the smallest bin size in the data and not allow binning smaller than this value (e.g. if the user demands a smaller bin size, it should return a cut with the minimum bin size and (re)set the value in the input box to this smallest bin size).

mducle:
Fixed by PR #120

PR #120, opened 2017-03-23
Bin step size limits and using SofQW3 to Q conversion

### LINKED ISSUE #118: Using SofQW3 and CopyToMD in calculate projections
At the moment `ConvertToMD` is used to convert the spectrum to |Q|, and the `BinMD` is used to generate cuts. However, this combination is akin to using centre-point rebinning (in the `BinMD` step) because `ConvertToMD` creates only a single event with coordinates at the centre of the bin of the original histogram data. This is not ideal because the rectangular bins in angle-energy space becomes trapezoids in |Q|-energy space. `SofQWNormalisedPolygons` (`SofQW3`) on the other hand rebins the data weighted by the fraction of the old bins falling into the new bins. Thus a more accurate workflow is to convert the data from scattering angle to |Q| using `SofQW3` first with a very fine Q-Energy grid. Then to convert this to an `MDEventWorkspace` using the `CopyToMD` option of `ConvertToMD` for subsequent rebinning. This would also provide a good indication of the minimum step size for issue #114 and #48 .

mducle:
Implements using `SofQWNormalisedPolygons` (`SofQW3`) to rebin data (usually from MARI or with a "rings" mapping) where pixel sizes are large before conversion to MDWorkspace with the `CopyToMD` method. For large datasets (and for indirect geometry data), the direct conversion to `|Q|` using `ConvertToMD` is still used because the centre-point rebinning of `BinMD` is not a problem in those cases.

Also, implemented a calculation of the `|Q|` limits and steps from the detector angles in the datafile at loading, which is attached to the `WorkspaceProvider` as a property. This required changing the `MainWindow` code to ensure that all widgets and presenters run the same instance of `WorkspaceProvider`. 

**To test:**

Load data (single or multiple files). If the data is direct inelastic from MARI, check that when "calculate projection" is run that `SofQWNormalisedPolygons` is called. If the data is direct inelastic with "one2one" mapping from Merlin or LET, check that `SofQWNormalisedPolygons` is *not* called - and same for Iris or Osiris data.

When plotting a slice, put in a step size of `0` or a very small number (e.g. `1e-6`) and check that it resets it back to the `optimum` step size, and displays an error message in the status bar.

Fixes #118, #28, #48, #114

### DIFF
--- a/mslice/models/workspacemanager/mantid_workspace_provider.py
+++ b/mslice/models/workspacemanager/mantid_workspace_provider.py
@@ -28,6 +28,8 @@ class MantidWorkspaceProvider(WorkspaceProvider):
 
     def delete_workspace(self, workspace):
         ws = DeleteWorkspace(Workspace=workspace)
+        if workspace in self._EfDefined.keys():
+            del self._EfDefined[workspace]
         if workspace in self._limits.keys():
             del self._limits[workspace]
         return ws
@@ -49,7 +51,10 @@ class MantidWorkspaceProvider(WorkspaceProvider):
         # For cases, e.g. indirect, where EFixed has not been set yet, return calculate later.
         try:
             efix = ws_h.getEFixed(ws_h.getDetector(0).getID())
-        except RuntimeError:
+        except RuntimeError:   # Efixed not defined
+            return
+        except AttributeError: # Wrong workspace type (e.g. cut)
+            self._limits[ws_name] = {}
             return
         # Checks that loaded data is in energy transfer.
         enAxis = ws_h.getAxis(0)
@@ -92,6 +97,8 @@ class MantidWorkspaceProvider(WorkspaceProvider):
         ws = RenameWorkspace(InputWorkspace=selected_workspace, OutputWorkspace=new_name)
         if selected_workspace in self._limits.keys():
             self._limits[new_name] = self._limits.pop(selected_workspace)
+        if selected_workspace in self._EfDefined.keys():
+            self._EfDefined[new_name] = self._EfDefined.pop(selected_workspace)
         return ws
 
     def save_nexus(self, workspace, path):
@@ -148,4 +155,7 @@ class MantidWorkspaceProvider(WorkspaceProvider):
 
     def propagate_properties(self, old_workspace, new_workspace):
         """Propagates MSlice only properties of workspaces, e.g. limits"""
-        self._limits[new_workspace] = self._limits[old_workspace]
+        if old_workspace in self._EfDefined.keys():
+            self._EfDefined[new_workspace] = self._EfDefined[old_workspace]
+        if old_workspace in self._limits.keys():
+            self._limits[new_workspace] = self._limits[old_workspace]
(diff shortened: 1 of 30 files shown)


mducle:
@martyngigg I'll look into the check of the line edit box, and delay it to either after the user finishes editing or when they click `Display`.

mducle:
@martyngigg I just had a chat with some other users and they also don't like the editing behaviour. They've suggested that a "magical" value of zero should result in the step size being changed to the default, but other values should be allowed. I've set a limit of 1/10000 of the minimum step size for which it will calculate a plot to avoid out-of-memory errors, but they prefer to give users the freedom to be stupid...

martyngigg:
@mducle Okay. It's probably best to get the change out there so it can be used to see if it really makes sense.

martyngigg:
@mducle The minimum bin size now works in a more understandable way I think so that's good.

Are there any files to test the bad nxspe issue?

martyngigg:
Verified that the new check for missing instrument names for nxspe files works. `master` version is unable to load the problematic files. This version loads and displays them correctly.

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -36,14 +44,67 @@ def _processEfixed(self, workspace):
...
+        m2A = 1.e10                                       # metres to Angstrom
+        if ws_name not in self._limits.keys():
+            self._limits[ws_name] = {}
+        th = np.array([np.min(theta), np.max(theta), np.mean(np.diff(theta))])
+        # Use |Q| at elastic line to get minimum and step size
+        qmin, qmax, qstep = tuple(np.sqrt(E2q * 2 * efix * (1 - np.cos(th)) * meV2J) / m2A)
+        # Use minimum energy (Direct geometry) or maximum energy (Indirect) to get qmax
+        emax = -np.min(en) if (str(ws_h.getEMode()) == 'Direct') else np.max(en)
+        qmax = np.sqrt(E2q * (2 * efix + emax - 2 * np.sqrt(efix * (efix + emax)) * np.cos(th[1])) * meV2J) / m2A
+        self._limits[ws_name]['MomentumTransfer'] = [qmin - qstep, qmax + qstep, qstep]
```
martyngigg: I think it might be nice to define all of these string contants in one place, probably a separate module.

I wonder if this might be a nice refactoring task for someone new coming in to get them used to the codebase?

mducle: Actually, they are defined as globals in https://github.com/mantidproject/mslice/blob/master/mslice/models/projection/powder/mantid_projection_calculator.py#L9 - however it's just file. We would need a C type `include`  file - a class of definitions maybe?

One thing which I found is that using `SofQW3` to convert the data from `Theta` to Q changes the Q unit to `MomentumTransfer`. However, using `ConvertToMD` with the `|Q|` option changes the unit to `|Q|`. Hence in the code, I had to make two entries in the dictionary... I think it would need some changes to the Mantid code to rationalise this, if it's considered a priority. At the moment it's just an annoyance really.

martyngigg: Hmm, that is annoyingly inconsistent. This seems like the difference between what we use as a key in our convert units dictionary vs display text. As you say, a minor annoyance at present.

PR #130, opened 2017-05-04
Change init to allow starting within MantidPlot as an interface

### LINKED ISSUE #121: Modify startup.py to allow use as an interface in MantidPlot
Currently, adding `start_mslice.py` as a custom interface in MantidPlot crashes MantidPlot when it is run, because it restarts `QApplication`. A check should be added to see if an instance of QApplication already exists, and if so not to restart it.

mducle:
Changes the `app/__init__.py` script to detect whether the `QApplication` main loop is running or not and as a result to start it up or not. Tries to detect common use cases:

  - In MantidPlot, `QApplication` is running - don't start it.
  - In IPython, `QApplication` *should* be running - but in some cases (e.g. isiscompute) does not. Detects both cases and if `QApplication` is not running, tries to use IPython methods to start it running in the background, so that scripting can also occur. If this also fails, prints a warning and start `QApplication` manually (blocking).
  - In normal Python - `QApplication` is not usually running. If so, start it.

**To test:**

Check that you can run `start_mslice.py` as an interface in MantidPlot. Check that running the commandline (`mslicedevel.sh`/`.bat` or `mslice`/`.bat`) still works.

Fixes #121.


martyngigg:
The push error is unrelated and needs changes from `master`.

Issue #133, opened 2017-05-05
Use LaTeX in axes titles


mducle:
Axes titles are just the name of the axes at the moment, which is a bit ugly - and also because of the change to using `SofQW3` + `CopyToMD` for some cases and `ConvertToMD` with `|Q|` in others causes the momentum transfer axes to change from `MomentumTransfer` to `|Q|`. 

It would be good to have uniform labels for the common axes, in LaTeX (so the Angstrom symbol renders correctly).

mducle:
Resolved by #134

Issue #132, opened 2017-05-05
Error in slice step calculation


mducle:
There is a bug in the calculation of the step size for slices.

The code is in `models/slice/mantid_slice_algorithm.py` line 55. It is currently:

```
        return int(max(1, floor(axis.end - axis.start)/axis.step))
```

it should be:

```
        return int(max(1, (axis.end - axis.start)/axis.step))
```

e.g. `floor` is in the wrong place - it should either be: `int(max(1, floor( (axis.end - axis.start)/axis.step) ))` or `floor` should be removed (we use `int` which rounds down anyway).

mducle:
Resolved by #134

Issue #131, opened 2017-05-05
Legacy matplotlib support


mducle:
Some instrument computers at ISIS are still using Red Hat 6. This means that they are restricted to using Python 2.6, and thus matplotlib 1.3.1. MSlice generally requires matplotlib >= 1.5.0, however the API changes are small, and no core feature of MSlice (apart from the `viridis` colormap) requires 1.5. 

It would thus be good if MSlice can detect the matplotlib versions and disable >=1.5 specific code, so it can still be run on older machines (even though Mantid is not officially supported on them anymore).

mducle:
Resolved by #134

Issue #122, opened 2017-03-28
Plot window focus


mducle:
After clicking `Plot` in the `Cut` tab, the focus should change to the plot window.

Presently, if a new plot window is create it inherits focus, but if the window already exists and the data is overplotted or replotted, then it does not get focus.

Ideally, this behaviour should be modifiable by an option somewhere.

mducle:
Can manually raise the window using:

```
import mslice.cli
import sys
window = mslice.cli.gcf().canvas.parent()
if 'Linux' in sys.platform.platform():
    window.showMaximised()
    window.showNormal()
window.raise_()
window.activateWindow()
```

The `showMaximised()` and `showNormal()` calls seem to be needed on RHEL7 (not tested on Ubuntu), but not on Windows (not tested on OSX).

mducle:
Resolved by #134

PR #134, opened 2017-05-07
Small fixes

mducle:
A series of small fixes:

  - Re issue #42 - Prints a warning on start up if matplotlib < 1.5 detected.
  - Re issue #103 - The legends box in cuts is now draggable
  - Re issue #105 - Line plots now have `o` markers.
  - Re issue #122 - now plot windows will inherit focus when `Display` or `Plot` is clicked
  - Re issue #131 - Mslice will run (but not well) under matplotlib 1.3.1 / python 2.6
  - Re issue #132 - slice step sizes now correctly calculated (removed `floor`).
  - Re issue #133 - Axes labels now in LaTeX.  

**To test:**

Check all tests pass, and no functionality has been affected


mslice/models/slice/matplotlib_slice_plotter.py
```diff
@@ -6,7 +6,11 @@
 class MatplotlibSlicePlotter(SlicePlotter):
     def __init__(self, slice_algorithm):
         self._slice_algorithm = slice_algorithm
-        self._colormaps = ['viridis', 'jet', 'summer', 'winter', 'coolwarm']
+        import matplotlib
```
martyngigg: Can we move the import and version check to the module-level near the top of the file and use [`distutils.version.LooseVersion`](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/plugins/algorithms/StringToPng.py#L37) instead? If we need to have a check anywhere else in the file in the future it would be better to do it once and set a flag such as `MPL_COMPAT`?

mslice/plotting/pyplot.py
```diff
@@ -2572,15 +2590,18 @@ def imshow(X, cmap=None, norm=None, aspect=None, interpolation=None, alpha=None,
     if hold is not None:
         ax.hold(hold)
     try:
+        if float('.'.join(matplotlib.__version__.split('.')[:2])) >= 1.5:
```
martyngigg: Same comment as above. Can we move the version check and only do it once on import of the module?

mslice/plotting/pyplot.py
```diff
@@ -2603,7 +2624,7 @@ def loglog(*args, **kwargs):
 
 # This function was autogenerated by boilerplate.py.  Do not edit as
 # changes will be lost
-@_autogen_docstring(Axes.magnitude_spectrum)
+#@_autogen_docstring(Axes.magnitude_spectrum)
```
martyngigg: Does `autogen_docstring` no work on older Matplotlib versions?

mducle: The decorator works fine. It's just that some of the properties don't exist in older versions of Matplotlib so it gives an error. I don't quite know how to disable the decorator for specific versions. I'll see if google or stack overflow comes up with something.

martyngigg: @mducle We'll probably need to wrap it with our own definition and check internally on the version: http://stackoverflow.com/questions/10724854/how-to-do-a-conditional-decorator-in-python-2-6

mducle: Ok. So it turned out I just need to put the whole of those blocks in an `if` - don't need to do anything fancy with conditional decorators.

mslice/plotting/pyplot.py
```diff
@@ -3041,26 +3068,27 @@ def triplot(*args, **kwargs):
...
-                            showextrema=showextrema, showmedians=showmedians,
-                            points=points, bw_method=bw_method, data=data)
-    finally:
-        ax.hold(washold)
-
-    return ret
+if not MPL_COMPAT:
+    @_autogen_docstring(Axes.violinplot)
+    @script_log(module_name)
+    def violinplot(dataset, positions=None, vert=True, widths=0.5, showmeans=False,
```
martyngigg: Should the whole function definition be indented here?

mducle: Yes. I think so, it gives an `IndentationError` otherwise. 

Actually, `Axes.violinplot`, `Axes.angle_spectrum`, `Axes.phase_spectrum` and `Axes.magnitude_spectrum` were all introduced in matplotlib 1.4, so don't strictly need matplotlib 1.5 - but I'm testing it on a system with matplotlib 1.3.1. Should we be a bit more fine grained in the check?

martyngigg: I don't have a problem with those only being available on 1.5. I can't imagine them being heavily used...

Issue #109, opened 2016-11-17
Option to toggle lines on/off for cut plots


martyngigg:
Legend titles can currently be toggled on or off in plot options. It should be possible to also toggle lines on/off.

Issue #107, opened 2016-11-17
Figure title (in task bar) should be editable


martyngigg:
This appears to be achievable using the http://matplotlib.org/api/backend_bases_api.html?highlight=set_window_title#matplotlib.backend_bases.FigureCanvasBase.set_window_title method on the canvas.

This makes most sense to again be in the plot options

Issue #104, opened 2016-11-17
Ability to customise style of line plots


martyngigg:
In the plot options menu we neeed to allow 1D line plots to be customised. This should include:
* line colour
* line style
* line width 
* marker type

These options should only be shown if the plot is of the appropriate type (1D line plot).

PR #136, opened 2017-08-01
Plot Options Refactor 

MatthewBowles:
Refactored PlotOptionsDialog class in preparation for issues #104, #107 and #109
-  Added PlotOptionsPresenter to use model-view-presenter pattern
- Modified PlotOptionsDialog and PlotFigureManager to use new presenter
- Divided PlotOptionsDialog into subclasses depending on plot type (slice or cut)
- Added unit tests for PlotOptionsPresenter

**To test:**
Code review, run unit tests. No changes in UI yet.

### DIFF
--- a/mslice/plotting/plot_window/plot_figure.py
+++ b/mslice/plotting/plot_window/plot_figure.py
@@ -6,9 +6,10 @@ import matplotlib.colors as colors
 from PyQt4.QtCore import Qt
 import numpy as np
 
+from mslice.presenters.plot_options_presenter import PlotOptionsPresenter, PlotConfig, LegendDescriptor
 from .plot_window_ui import Ui_MainWindow
 from .base_qt_plot_window import BaseQtPlotWindow
-from .plot_options import PlotOptionsDialog, LegendDescriptor, PlotConfig
+from .plot_options import PlotOptionsDialog
 
 
 class PlotFigureManager(BaseQtPlotWindow, Ui_MainWindow):
@@ -50,7 +51,7 @@ class PlotFigureManager(BaseQtPlotWindow, Ui_MainWindow):
 
     def _plot_options(self):
         config = self._get_plot_description()
-        new_config = PlotOptionsDialog.get_new_config(config)
+        new_config = PlotOptionsPresenter(PlotOptionsDialog(config)).get_new_config()
         if new_config:
             self._apply_config(new_config)
 
--- a/mslice/plotting/plot_window/plot_figure.py
+++ b/mslice/plotting/plot_window/plot_figure.py
@@ -51,7 +51,8 @@ class PlotFigureManager(BaseQtPlotWindow, Ui_MainWindow):
 
     def _plot_options(self):
         config = self._get_plot_description()
-        new_config = PlotOptionsPresenter(PlotOptionsDialog(config)).get_new_config()
+        view = PlotOptionsDialog(config) # tmp - remove config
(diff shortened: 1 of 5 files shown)


mslice/plotting/plot_window/plot_figure.py
```diff
@@ -49,12 +52,17 @@ def _display_status(self,status):
             self.actionKeep.setChecked(False)
 
     def _plot_options(self):
-        config = self._get_plot_description()
-        new_config = PlotOptionsDialog.get_new_config(config)
+        if self.canvas.figure.gca().get_images():
```
martyngigg: It feels as if we need a helper function somewhere such as `is_cut_figure()` and `is_slice_figure()` so that we can encapsulate this check and make the code more readable. What do you think?

We could also use common python approach to reduce the amount of similar code we type:
```
if self.canvas.figure.gca().get_images():
    view_class, presenter_class = SlicePlotOptions, SlicePlotOptionsPresenter
else:
    view_class, presenter_class = CutPlotOptions, CutPlotOptionsPresenter
new_config = presenter_class(view_class(), self).get_new_config()
...
```

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -65,74 +73,50 @@ def _get_min(self, data, absolute_minimum=-np.inf):
...
-            elif not plot_config.logarithmic and type(mappable.norm) != colors.Normalize:
-                mappable.colorbar.remove()
-                norm = colors.Normalize(vmin, vmax)
-                mappable.set_norm(norm)
-                self.canvas.figure.colorbar(mappable)
+        if xy_config['x_log']:
+            xdata = [ll.get_xdata() for ll in current_axis.get_lines()]
+            xmin = self.get_min(xdata, absolute_minimum=0.)
+            current_axis.set_xscale('symlog', linthreshx=pow(10, np.floor(np.log10(xmin))))
+            if self.get_min(xdata) > 0:
```
martyngigg: Is there a reason to do this calculation again rather than just using the value calculated 2 lines above?

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -65,74 +73,50 @@ def _get_min(self, data, absolute_minimum=-np.inf):
...
-        # the plot (in terms of having/not having errorbars)
-        if legend_config.visible:
-            leg = current_axis.legend()
-            leg.draggable()
+            current_axis.set_xscale('linear')
+        if xy_config['y_log']:
+            ydata = [ll.get_ydata() for ll in current_axis.get_lines()]
+            ymin = self.get_min(ydata, absolute_minimum=0.)
+            current_axis.set_yscale('symlog', linthreshy=pow(10, np.floor(np.log10(ymin))))
+            if self.get_min(ydata) > 0:
```
martyngigg: Is there a reason to do this calculation again rather than just using the value calculated 2 lines above?

martyngigg:
@MatthewBowles The code looks pretty good now. Thanks for making those changes.

The final thing worth looking at is see if the test coverage warning above needs addressing. If you click on the "coverage 94%" badge it shows that the presenter has lower than 80% coverage. It's mainly missing things around the legends. I think it could be worth addressing that in the tests.

martyngigg:
As discussed the legend code is going to be updated when the new options are added and further testing can be added there. Good to go now.

PR #138, opened 2017-08-03
Set Window Title on Plot Figure

### LINKED ISSUE #107: Figure title (in task bar) should be editable

MatthewBowles:
It makes sense to set the window title to be the same as the figure title, so I've added a setWindowTitle call to the title setter.
Unfortunately this means that the window title will be 'Figure x' as before until the title is modified in plot options, so will not match the figure title on launch. Fixing this seems non-trivial since the figure title isn't decided until sometime after PlotFigureManager's init method.

**To test:**
-Open any plot and go to plot options
-Change figure title
-new figure title should also be displayed as the window title (replacing 'Figure 1')

Fixes #107

### DIFF
--- a/mslice/models/cut/matplotlib_cut_plotter.py
+++ b/mslice/models/cut/matplotlib_cut_plotter.py
@@ -20,6 +20,7 @@ class MatplotlibCutPlotter(CutPlotter):
         leg.draggable()
         plt.xlabel(self._getDisplayName(cut_axis.units, self._cut_algorithm.getComment(selected_workspace)))
         plt.ylabel(INTENSITY_LABEL)
+        plt.gcf().canvas.set_window_title(selected_workspace)
         plt.autoscale()
         plt.ylim(intensity_start, intensity_end)
         plt.draw_all()
--- a/mslice/models/cut/matplotlib_cut_plotter.py
+++ b/mslice/models/cut/matplotlib_cut_plotter.py
@@ -20,7 +20,6 @@ class MatplotlibCutPlotter(CutPlotter):
         leg.draggable()
         plt.xlabel(self._getDisplayName(cut_axis.units, self._cut_algorithm.getComment(selected_workspace)))
         plt.ylabel(INTENSITY_LABEL)
-        plt.gcf().canvas.set_window_title(selected_workspace)
         plt.autoscale()
         plt.ylim(intensity_start, intensity_end)
         plt.draw_all()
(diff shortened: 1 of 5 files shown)


martyngigg:
@MatthewBowles This is a good start but as a user it would seem odd to me that the title is known when we plot but we can't get it to the window until later. Perhaps we have to have an additional call to set it when we create the plot in `mslice.models.slice.matplotlib_slice_plotter.py`? and also keep the option from the dialog (http://matplotlib.1069221.n5.nabble.com/Window-title-of-a-figure-td10697.html)?
Can you look into this?

MatthewBowles:
I'll take another look at this. I'm not sure how to get a reference to the window (and not just the figure) from the slice_plotter, but it should be possible.

MatthewBowles:
@martyngigg In most recent commit, I added the code from the link to slice_plotter without changing it, but it does not work. (Just displays new figure with title 'Figure 2', not the title I provided).
This feels like it should be simple but I'm not sure why it doesn't work or what else to try.

martyngigg:
@MatthewBowles I finally figured out why the window titles weren't working. Our version of the FigureManager was missing the required methods and our canvas was also missing a reference to the
manager.

Can you test this one to see if it does what you expect?

MatthewBowles:
@martyngigg Yeah that's good, should be done now.

mslice/models/cut/matplotlib_cut_plotter.py
```diff
@@ -20,7 +20,6 @@ def plot_cut(self, selected_workspace, cut_axis, integration_start, integration_
         leg.draggable()
         plt.xlabel(self._getDisplayName(cut_axis.units, self._cut_algorithm.getComment(selected_workspace)))
         plt.ylabel(INTENSITY_LABEL)
-        plt.gcf().canvas.set_window_title(selected_workspace)
```
martyngigg: @MatthewBowles Did you mean to remove this line?

MatthewBowles: by default, the cut plots have no title. I was just sticking to this, so it would be figure 1 by default. Can change it if you would prefer.

martyngigg: Okay, that's fine. At least we know how to change it now.

Issue #140, opened 2017-08-09
Bug in logarithmic color plot options


mducle:
There is a bug in converting from a log color scale back to linear. To reproduce:

- Load a dataset, calculate projections and display a slice
- Open the `Plot Options` menu and check the `Logarithmic` box in the `Colorbar` section, and click `OK`
- Open `Plot Options` again and set `Min` to `0`, uncheck `Logarithmic` and click `OK`

It should show the data again on a linear colour scale, with minimum zero, but instead gives a `ZeroDivisionError`. This is due to line 108 of `mslice/plotting/plot_window/plot_figure.py` - the call to  `mappable.set_clim`. This call is needed if the user just changes the values of `Min` and/or `Max` but does not change the value of `Logarithmic`.

A fix would be to extend the `if` block to put the call to `mappable.set_clim` as another `else` or `elif` block - `elif` to check if `mappable.norm` is `colors.LogNorm` and then to give an error if `Min` is zero (and not to implement the change).

martyngigg:
Closing manually as the PR was missing the "fixes #" comment.

PR #142, opened 2017-08-10
Logarithmic color plot bug

MatthewBowles:
Fixed bug by moving call to mappable.set_clim() to after the log conversion.

**To test:**
See issue #140


martyngigg:
Fixes #140

Issue #51, opened 2016-08-31
MSlice will not recognise a cut when the parent is not in the ADS


EltayebAhmed:
A possible solution would be to reimplement CutAlgorith.is_cut and CutAlgorithm.get_cut_params without depending on the workspace history.
MSlice currently checks to see if a workspace is a cut by looping through the workspace history and searching for a call to BinMD.
MSlice extracts the parameters by looking at the properties passed to the BinMD algorithm.
This is superfluous and will fail if the parent algorithm does not exist in the ADS. 
To be exact 

```
        history = cut_workspace.getHistory()
        for i in range(history.size()-1,-1,-1):
            algorithm = history.getAlgorithm(i) #<- RuntimeError will be raised here 
```

The above snippet will fail if run on a workspace in ipython in mantid on workspace
 even though right-clicking the smae workspace and clicking show History will bring up the entire workspaces history

It is possible to deduct the cut axis specs by doing something along the lines of

```
dim = workspace.getDimension(0)
name = dim.getName()
min = dim.getMinimum()
max = dim.getMaximum()
step = dim.getX(1) - dim.getX(0)
```

All that is left is to find a way to find the integration limits without failing if parent has been removed from the ADS

martyngigg:
@EltayebAhmed Please provide more details of what the actual issue is. Most of what we discussed on slack would be good to put in the description for reference. If you don't get to it we need to preserve the information.

EltayebAhmed:
If you right-click on a workspace in mantidPlot and click show history it will  bring up the entire history regardless of if any of the parents are in the ADS or not

MatthewBowles:
Fixed in #222

PR #147, opened 2017-08-15
Fix AttributeError when loading cut

### LINKED ISSUE #146: Loading cut from file causes exception
Cuts can be saved to a workspace, and then to a file. Attempting to load this file causes an AttributeError.
```
  workspace_manager_presenter.py, line 26, in notify
    self._load_workspace()
  workspace_manager_presenter.py, line 65, in _load_workspace
    self._work_spaceprovider.load(filename=workspace_to_load[ii], output_workspace=ws_name)
  mantid_workspace_provider.py, line 97, in load
    if self.get_emode(output_workspace) == 'Indirect':
  mantid_workspace_provider.py, line 136, in get_emode
    emode = workspace_handle.getEMode().name
AttributeError: 'MDHistoWorkspace' object has no attribute 'getEMode'
```

MatthewBowles:
Loading a cut workspace used to cause an AttributeError, now it loads correctly. 

**To test:**
- Load file, calculate projections, input cut parameters
- Save cut to workspace
- Save cut to file
- Load cut

Loading should succeed, and the cut should be plottable (as long as parent workspace is in ADS - see #51 )

Fixes #146


mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -132,7 +132,10 @@ def get_emode(self, workspace):
             workspace_handle = self.get_workspace_handle(workspace)
         else:
             workspace_handle = workspace
-        emode = workspace_handle.getEMode().name
+        try:
+            emode = workspace_handle.getEMode().name
+        except AttributeError:
+            return None
```
martyngigg: I'm not sure about silently swallowing this. If we can't retrieve it then the caller needs to know. How about we model it on Python's [`dict.get()` method](https://docs.python.org/2/library/stdtypes.html#dict.get)? The caller can choose whether to see an exception or not and then handle it appropriately.

martyngigg:
@MatthewBowles While this now loads the file I can't seem to do anything with it afterwards. Is that the intention?

martyngigg:
I think the issue in this case is that an `MDHistoWorkspace` is being loaded but the `get_efixed` method can't handle this. The `MDHistoWorkspace` inherits from [MultipleExperimentInfos](http://doxygen.mantidproject.org/nightly/d7/d1f/classMantid_1_1API_1_1MultipleExperimentInfos.html). The `ExperimentInfo` class then as the `getEmode` method. I think we need to improve this method so it can look at the multiple `ExperimentInfo` classes and return the emode if they are all the same.

MatthewBowles:
@martyngigg As discussed, a cut cannot be plotted without the parent due to #51.
Load can now access the `ExperimentInfo` classes from a `MDHistoWorkspace` correctly.

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -61,9 +61,9 @@ def _processLoadedWSLimits(self, workspace):
             else:
                 self._limits[ws_name] = None
                 return
-        except AttributeError: # Wrong workspace type (e.g. cut)
-            self._limits[ws_name] = {}
-            return
+        # except AttributeError: # Wrong workspace type (e.g. cut)
```
martyngigg: Can we also remove the commented out code?

MatthewBowles:
@martyngigg :+1:

martyngigg:
@MatthewBowles Loading of the original woskspace is working again :+1: 

Unfortunately after performing a cut, saving to a file then trying to load that cut again I get:

```
  File "/media/data1/source/github/mantidproject/mslice/mslice/widgets/workspacemanager/workspacemanager.py", line 45, in _btn_clicked
    self._presenter.notify(command)
  File "/media/data1/source/github/mantidproject/mslice/mslice/presenters/workspace_manager_presenter.py", line 27, in notify
    self._load_workspace()
  File "/media/data1/source/github/mantidproject/mslice/mslice/presenters/workspace_manager_presenter.py", line 65, in _load_workspace
    self._workspace_provider.load(filename=workspace_to_load[ii], output_workspace=ws_name)
  File "/media/data1/source/github/mantidproject/mslice/mslice/models/workspacemanager/mantid_workspace_provider.py", line 98, in load
    if self.get_EMode(output_workspace) == 'Indirect':
  File "/media/data1/source/github/mantidproject/mslice/mslice/models/workspacemanager/mantid_workspace_provider.py", line 137, in get_EMode
    emode = self._get_ws_EMode(workspace_handle).name
AttributeError: 'NoneType' object has no attribute 'name'

If you suspect this is an IPython bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@scipy.org

You can print a more detailed traceback right now with "%tb", or use "%debug"
to interactively debug it.

Extra-detailed tracebacks for bug-reporting purposes can be enabled via:
    %config Application.verbose_crash=True
```

MatthewBowles:
@martyngigg The calculate projections button is now disabled and a message is displayed if the emode cannot be retrieved from the selected workspace. This maintains compatibility with any loaded MDWorkspaces which do have an emode set.

martyngigg:
@MatthewBowles This looks really nice. One tiny thing would be if we could use the message that Duc provided previously? I think that would make more sense to users. After that I think it's good to go.

MatthewBowles:
@martyngigg :+1:

MatthewBowles:
@martyngigg I've fixed the formatting and the error when no workspaces are selected. The build seems unhappy for some reason though.

martyngigg:
@MatthewBowles It seemed to be a issue unrelated to the code and a rebuild solved it.

Issue #129, opened 2017-04-08
Option to change resolution (DPI) of save figure


mducle:
MSlice currently uses the standard matplotlib `save_figure` toolbar option, which does not allow a resolution to be specified. It instead uses a per-user default (the `savefig.dpi` property in `.matplotlibrc`) which is set to 100dpi unless the user changes it. This is quite low resolution for publication.

So, there should be an option to change the default resolution of figures, which can be implemented separately (e.g. programatically setting `savefig.dpi` through `plt.rcParams` so that `save_figure` can still be used), or a custom save figure function / dialog can be implemented to ask the user for the resolution (and then using the matplotlib `savefig` method directly.

PR #150, opened 2017-08-17
Add print button to plot figure

### LINKED ISSUE #149: Ability to print (to hardcopy)
Add print option to slices and cuts, perhaps in plot_figure window.

MatthewBowles:
Added a 'Print Plot' button to the plot_figure window.

**To test:**
- Plot any slice or cut
- Click 'Print Plot'
- Choose options in the print dialog
- If 'Print' is clicked, the plot is then printed out. 

Fixes #149


martyngigg:
@MatthewBowles This looks much better now and the options are being respected. The only thing is that the resolution is a little poor. Can we look if we can make it any better, perhaps with http://doc.qt.io/qt-5/qprinter.html#setResolution ?

MatthewBowles:
@martyngigg setResolution doesn't seem to work for some reason - I'll try to find an alternative

MatthewBowles:
@martyngigg it's better than it was before. There seems to be a limit to how high the resolution can go though, which I'm guessing is the resolution the figure is displayed at. Might be changeable in `.matplotlibrc` like #129.
Anyway, this should be good enough for now.

PR #152, opened 2017-09-05
Wrappers for Mantid workspace classes

### LINKED ISSUE #151: Add thin wrappers to Mantid workspace classes to give a consistent interface
The `mantid` workspace types all have very different interfaces. It would be more helpful for `mslice` users if there was a consistent interface to these types to allow easier manipulation on the command line.

Implement a set of wrappers around `MatrixWorkspace`, `MDHistoWorkspace` and `MDEventWorkspace` to provide this interface. The class names should be:
  - `Workspace` wraps a `MatrixWorkspace`
  - `HistogramWorkspace` wraps an `MDHistoWorkspace`  
  - `PixelWorkspace` wraps an `MDEventWorkspace`  
  
The wrapper classes should all have the following methods:

- `get_coordinates` - returns the ND coordinates as an N-element `dict`.
- `get_signal` - returns a `numpy` ND array of the signal. If applied to a `pixel_workspace` 
  implicitly returns rebinned data (converts to `histogram_workspace` internally).
- `get_error` - returns a `numpy` ND array of the standard errors. Rebins `pixel_workspace` as for 
  `get_signal`.
- `get_variance` - returns a `numpy` ND array of the variance (error^2). Rebins as noted above for
  `pixel_workspace`
---

The classes should also overload these operators:

- `operator(workspace, workspace)` - performs a binary operator (+, -, *, /) on two workspaces of 
  the same type, dimensionality and binning, by applying the operator to each pixel / bin.
- `operator(workspace, number)` - performs a binary operator (+, -, *, /) on a workspace and a 
  scalar constant, by adding / subtracting / multiplying / dividing the signal array by the constant
- `operator(workspace, list)` - performs a binary operator (+, -, *, /) on a workspace and a `list`
  or 1D `numpy` array with the same number of elements as an axis of the workspace by recursively 
  using `replicate` on the `list` / `numpy` array and then applying `operator(workspace,workspace)`.
- `__pow__(workspace, number) - raises each intensity value to the given power. e.g. `ws2 = ws**2`
- `__neg__(workspace)` - takes the negative of each intensity value. e.g. `msw = -ws`

The wrapper classes should go into a new subpackage: `mslice.workspace`.

cc @mducle

MatthewBowles:
Added workspace wrappers according to description in #151.

**To test:**
Run unit tests

Fixes #151 

@mducle @martyngigg


mslice/workspace/workspace.py
```diff
@@ -0,0 +1,77 @@
...
+import operator
+from mantid.simpleapi import CreateWorkspace
+
+
+class Workspace(object):
+
+    def __init__(self, matrix_workspace):
+        self.inner_workspace = matrix_workspace
+
+    def get_coordinates(self):
```
martyngigg: Can we have some method documentation here and on the other methods, please? See https://www.python.org/dev/peps/pep-0257/

mslice/tests/workspace_test.py
```diff
@@ -0,0 +1,217 @@
+import unittest
```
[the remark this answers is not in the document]
MatthewBowles: `PixelWorkspaceTest` definitely belongs in its own file, but I'm unsure about the others.

The tests for `Workspace` and `HistogramWorkspace` are basically the same, so they inherit from `BaseWorkspaceTest` (which doesn't actually have any tests on its own, just provides common implementation to the two subclasses). We might want to keep these in the same file since they're so closely related?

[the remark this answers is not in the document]
martyngigg: @MatthewBowles I think it would still be good to split them up. If I'm looking for the `HistogramWorkspace` test then it's not that obvious that it's with the workspace one. Can you just import the `BaseWorkspaceTest` class in the `Histogram` test?

mslice/workspace/base.py
```diff
@@ -0,0 +1,48 @@
+import abc
+
+
+class WorkspaceBase(object):
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, mantid_workspace):
```
martyngigg: I think we would be better leaving `__init__` out completely as this is now an interface so it would be better that it has no implementation.

mslice/tests/workspace_test.py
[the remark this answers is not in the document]
MatthewBowles: Good point, will do.

mducle:
@martyngigg @MatthewBowles It looks fine to me. I've just run a quick check and everything seems to work as advertised (except when I purposefully try to break it by constructing things from a non-workspace type (or the wrong workspace type).

mslice/workspace/workspace.py
```diff
@@ -0,0 +1,103 @@
...
+import operator
+from base import WorkspaceBase
+from mantid.simpleapi import CreateWorkspace
+
+
+class Workspace(WorkspaceBase):
+    """workspace wrapper for MatrixWorkspace"""
+
+    def __init__(self, mantid_workspace):
+        self._raw_ws = mantid_workspace
```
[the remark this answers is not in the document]
martyngigg: Thinking about this I wonder if we might be better with a `mixin-style` hierarchy (https://www.ianlewis.org/en/mixins-and-python). This would allows us to decouple the types in the inheritance hierarchy so that, for example, `PixelWorkspace` and `Workspace` do not inherit each other but both inherit `WorkspaceBase` and then we mixin other classes to add the functionality.

MatthewBowles:
@martyngigg I've implemented the mixin classes. I tried to keep them loosely-coupled and added a `rewrap` method to help with this. Changed the `__pow__` methods to be more efficient as well.
The CI build is giving 404 errors for some reason though.

martyngigg:
@MatthewBowles There was an issue with the nightly deployment of the mantid build. I've fixed that and restarted it.

mslice/workspace/histo_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+                                    SignalInput=other, ErrorInput=other, NumberOfBins=str(size),
+                                    Names=self._raw_ws.getDimension(0).getName(), Units='MomentumTransfer')
+        try:
+            replicated = ReplicateMD(self._raw_ws, ws)
+            return operator(self._raw_ws, replicated)
+        except RuntimeError:
+            raise RuntimeError("List or array must have same number of elements as an axis of the workspace")
+
+    def __pow__(self, exponent):
+        pow_alg = AlgorithmManager.createUnmanaged('PowerMD')
```
martyngigg: Can we do the same with `CreateMDHistoWorkspace` and `ReplicateMD` above? Maybe a utility function for running an  algorithm in this way that accepts `(name, *args, **kwargs)`?

martyngigg:
@MatthewBowles The builds look to be fixed now.

mslice/workspace/histo_mixin.py
MatthewBowles: @martyngigg This seems like something that would be useful in Mantid also - this is a common pattern for running algorithms but it's quite verbose. Maybe it would be better to implement something in Mantid that Mslice could import.

martyngigg: @MatthewBowles Agreed. @gvardany is working on something now where you would be able to do `ws = PowerMD(..., RunAsChild=True)` that would achieve the same result. I don't think we should wait for that.

martyngigg:
@MatthewBowles Could you make those minor changes to the way the algorithms are run? We can have a future maintenance task once Mantid has a sensible generic implemetation for this.

MatthewBowles:
@martyngigg I've made the changes but the build keeps giving seg faults?

PR #154, opened 2017-10-10
Python 3 compatibility

### LINKED ISSUE #153: Python 3 support
Mantid itself is very close to having complete support for Python 3. We should add this support to `mslice`.

This will involve:

* converting code to use constructs that both Python 2/Python 3 can understand: https://www.mantidproject.org/Python_3, using `six` where appropriate
* setting up build support to enable us to test this so that we don't introduce regressions in the future

MatthewBowles:
This PR aims to convert the code to be compatible with both Python 2 and 3.

**To do:**
Build support to prevent future regressions

Fixes #153


mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -78,7 +78,7 @@ def _processLoadedWSLimits(self, workspace):
         E2q = 2. * constants.m_n / (constants.hbar ** 2)  # Energy to (neutron momentum)^2 (==2m_n/hbar^2)
         meV2J = constants.e / 1000.                       # meV to Joules
         m2A = 1.e10                                       # metres to Angstrom
-        if ws_name not in self._limits.keys():
+        if ws_name not in list(self._limits.keys()):
```
martyngigg: Same with `list` and `keys()`

MatthewBowles:
@martyngigg The new build script setup seems to be working with 2.7.

martyngigg:
@MatthewBowles Is this ready to review again now?

MatthewBowles:
@martyngigg Not really, the tests pass but the python 3 version still has a few bugs

martyngigg:
@MatthewBowles Okay, thanks. Are the bugs purely in the GUI layer or is it possible to add tests to catch them?

MatthewBowles:
@martyngigg Looks like it was just a problem with launching a new window. Should be good to go now.

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -166,9 +168,9 @@ def get_limits(self, workspace, axis):
 
     def propagate_properties(self, old_workspace, new_workspace):
         """Propagates MSlice only properties of workspaces, e.g. limits"""
-        if old_workspace in self._EfDefined.keys():
+        if old_workspace in self._EfDefined:
             self._EfDefined[new_workspace] = self._EfDefined[old_workspace]
-        if old_workspace in self._limits.keys():
+        if old_workspace in self._limits:
```
martyngigg: @MatthewBowles Do you know why the `.keys` error here that was fixed in the last commit wasn't picked up by a test?

MatthewBowles: @martyngigg Looks like the only classes that use it are mocked out in the tests

martyngigg: I see. Would it be hard to write some tests for it?

MatthewBowles: Seems doable. Should I open a separate issue?

martyngigg: I think that would be a good plan. You can never have too much test coverage!

martyngigg:
@MatthewBowles I'm assuming this is ready to go now?

MatthewBowles:
@martyngigg Yes

martyngigg:
:+1:

PR #157, opened 2017-10-16
MantidWorkspaceProvider unit tests

### LINKED ISSUE #156: Unit tests for MantidWorkspaceProvider
The MantidWorkspaceProvider class currently has no unit tests.

MatthewBowles:
Added unit tests for MantidWorkspaceProvider class.

Fixes #156

### DIFF
--- a/mslice/models/workspacemanager/mantid_workspace_provider.py
+++ b/mslice/models/workspacemanager/mantid_workspace_provider.py
@@ -18,6 +18,7 @@ from .workspace_provider import WorkspaceProvider
 # Classes and functions
 # -----------------------------------------------------------------------------
 
+
 class MantidWorkspaceProvider(WorkspaceProvider):
     def __init__(self):
         # Stores various parameters of workspaces not stored by Mantid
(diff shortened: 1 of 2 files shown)


MatthewBowles:
@martyngigg Here's a start. I'm not sure if I can/should test some of the other stuff (e.g. the other routes through `_processLoadedWSLimits`) since they seem tied up in the mantid workspace code. Thoughts?

martyngigg:
This is a good start. I do think tests for `get_limits` (so ultimately  `processLoadedWSLimits`) are required as it's quite complicated and thus something could feasibly go wrong with a calculation and it could go unnoticed.

MatthewBowles:
@martyngigg I've added some tests for `get_limits`.

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -60,9 +74,6 @@ def _processLoadedWSLimits(self, workspace):
             else:
                 self._limits[ws_name] = None
                 return
-        except AttributeError: # Wrong workspace type (e.g. cut)
```
martyngigg: I think we still need these lines? If we don't have them then the nice work from #147 breaks.

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -60,9 +74,6 @@ def _processLoadedWSLimits(self, workspace):
             else:
                 self._limits[ws_name] = None
```
martyngigg: I know you didn't change this but could we use `self._limits[ws_name] = {}` here instead for consistency with other places here?

MatthewBowles:
@martyngigg OK. I've tried to tidy up `_processLoadedWSLimits` while I was there since it was way too long.

PR #159, opened 2017-10-19
Add error messages to calculate projection

### LINKED ISSUE #38: Implement error handling for powder projections

MatthewBowles:
Previously, misuse of the calculate projections button would give errors in the console but not report anything to the user through the GUI. Now, messages are displayed if no workspace is selected or a workspace is the wrong type.

**To test:**
- Click calculate projection without selecting a workspace
- Observe error message

And:

- Select a workspace that is not a workspace2D (e.g. _QE)
- Click calculate projection
- Observe error message

Fixes #38


mslice/models/projection/powder/mantid_projection_calculator.py
```diff
@@ -103,3 +104,8 @@ def calculate_projection(self, input_workspace, axis1, axis2, units):
 
     def set_workspace_provider(self, workspace_provider):
         self._workspace_provider = workspace_provider
+
+    def validate_workspace(self, ws):
+        ws_type = self._workspace_provider.get_workspace_handle(ws).id()
+        if ws_type != 'Workspace2D':
```
martyngigg: @MatthewBowles This is much better than seeing the error in the console. I'm not sure whether the error message would make sense to users though. Generally they will not be aware of the subtle differences between the mantid workspace types.

@mducle Could you suggest a more user-friendly error message for this scenario?

mducle: Something like: `Input workspace for projection calculation must be a reduced data workspace`?

However, I think we should be using duck-typing instead because I think one of the axes should be in energy transfer (e.g. if you loaded a raw datafile, the workspace will be `Workspace2D` but the bin axis will be time-of-flight. I think `convertToMD` might still work on this (or on workspaces converted into units of wavelengths of d-spacing or whatever) but we might not want those workspaces for MSlice.

So maybe:

```
def validate_workspace(self, ws):
    try:
        axes = [self._workspace_provider.get_workspace_handle(ws).getAxis(0),
                self._workspace_provider.get_workspace_handle(ws).getAxis(1)]
        if not all([ax.isSpectra() or ax.getUnit.unitID() == 'DeltaE' for ax in axes]):
            raise AttributeError
    except (AttributeError, IndexError):
        raise TypeError('Input workspace for projection calculation must be a reduced '
                        'data workspace with a spectra and energy transfer axis.')
```

?

martyngigg: Yes, that's a good point.

martyngigg:
@MatthewBowles  Just a few flake8 warnings to fix and then we can merge this.

MatthewBowles:
@martyngigg Done

PR #166, opened 2017-11-06
Add Chi''(Q,E) and Chi''(Q,E) magnetic display options

### LINKED ISSUE #164: Intensity display on slices: show chi''(q,w)
There should be an option to display the intensity (colour) of slices as the dynamical susceptibility `chi''(Q,E)` instead of the scattering function `S(Q,E)`. The two are related by the fluctuation-dissipation theorem, which is given [here ](http://docs.mantidproject.org/nightly/algorithms/ApplyDetailedBalance-v1.html) as:

```
chi''(Q,E) = PI * (1/<n+1>) * S(Q,E)
```

for neutron energy loss (positive energy transfer), where the thermal population factor `1/<n+1>`  is `(1 - exp(-|E|/(kB.T)))`. For neutron energy gain (negative energy transfer), instead

```
chi''(Q,E) = PI * (1/<n>) * S(Q,E)
```

should be used where `1/<n>=(exp(|E|/(kB.T))-1)` [note the positive sign in the exponent and also that the modulus of the energy should be used (because we are thinking in terms of the energy of the excited magnon or phonon quasiparticles and not the neutron's energy)]. E.g.

```
chi''(Q,E) = PI * (1-exp(-E/(kB.T))) * S(Q,E)       for E>=0
chi''(Q,E) = PI * (exp(abs(E)/(kB.T))-1) * S(Q,E)       for E<0
```

`S(Q,E)` is usually in units of `milibarn/sr/meV/f.u.` (where `f.u.` is a formula unit of material studied). Since the other factors are dimensionless, `chi''(Q,E)` in this case is also in the same units as `S(Q,E)`. Although the stated use of this conversion is to compare with theoretical calculations in absolute units, this is almost only ever done for magnetic scattering (see below). For lattice dynamics, the calculations will need to be scaled by some factor to match the `chi''(Q,E)` shown here, and rather the use of this conversion is mainly instead to remove the temperature dependence of the scattering (thermal population factor of the phonons). If the color axis is labelled, it should be labelled with `(arb. units)`. 

For magnetic scattering, however, because there is a total magnetic cross-section which depends on the magnetic moment, this can be factored out, yielding a dynamical susceptibility in absolute units. These units are `mu_B^2/meV/sr/f.u.`. The conversion in this case is:

```
chi''(Q,E) * |F(Q)|^2 = [1/((gamma * r_e / 2)^2) * PI * (1/<n+1/2+/-1/2>) * S(Q,E) ] * (uB^2)
```

where `<n+1/2+/-1/2>` is the factor `<n+1>` or `<n>` described above for the cases of `E>=0` or `E<0`; `u_B` is the Bohr magneton, and the quantity `(gamma * r_e / 2)^2=(-0.54e-14 m)^2 = 291 milibarns` is the total neutron cross-section for a moment of one bohr magneton (note `gamma=-3.826` is the nuclear gyromagnetic ratio and `r_e=2.818e-15 m` is the classical electron radius).

We keep the factor (uB^2) in the units, so the only change from the non-magnetic case is to divide the intensity further by 291 milibarns. The axis should now be labelled `\chi''(Q,E) |F(Q)|^2 (\mu_B meV^{-1} sr^{-1} f.u.^{-1}`.

---

The conversion should be applied to the plotted slice only (can be done directly on the numpy array, or on the `MDHistoWorkspace`), but it needs access to the energy transfer information to calculate the thermal population factor. 

There should be a menu option `Intensity` with a set of radio options. The default is `S(Q,E)` which means that the workspace should be plot as it is. This issue adds to other options: 

1. `chi''(Q,E)`
2. `chi''(Q,E) magnetic`

If the user selects one of these, unless the intensity data is already in this form, `mslice` should convert the intensity using the above relationships, and then modify the plot and the color axis label.

MatthewBowles:
Added options under 'Intensity' to change the slice display to dynamical susceptibility `chi''(Q,E)`.
The temperature is retrieved by asking the user to choose the correct entry in the sample log. Their response is cached and they are not asked again unless the entry is not found.

**To test:**
- Load a workspace
- Calculate projections 
- Display slice
- Click 'Intensity' and select a new option

Fixes #164


MatthewBowles:
@mducle Could you please check that the output of this is as expected?

mducle:
It's not quite right... the exponent should be the (modulus) energy transfer (axis) divided by `k_B T` - rather than the intensity divided by `k_B T` as currently implemented.

Also, could you store the temperature with the workspace (maybe in the `SlicePlotterPresenter` like the cache of cut parameters in the `CutPresenter`) so that you only need to input the temperature once rather each time you plot a slice? (And if you can defer asking the temperature until the user selects either of the `chi` options rather than when it's initially plotted [e.g. the default `S(q,w)` doesn't need the temperature information] - that would be great.)

MatthewBowles:
Yeah no problem, I'm still working on the temperature input. 
I'll take another look at the calculation.

mslice/models/slice/matplotlib_slice_plotter.py
```diff
@@ -40,6 +44,77 @@ def _getDisplayName(self, axisUnits, comment=None):
...
+
+    def update_sample_temperature(self, workspace):
+        temp = self.sample_temperature(str(workspace))
+        self.slice_cache[workspace]['sample_temp'] = temp
+        self.slice_cache[workspace]['plot_data'][1] = self._slice_algorithm.compute_chi(
+            self.slice_cache[workspace]['plot_data'][0], temp, self.slice_cache[workspace]['y_axis'])
+        self.slice_cache[workspace]['plot_data'][2] = self._slice_algorithm.compute_chi_magnetic(
+            self.slice_cache[workspace]['plot_data'][1])
+
+    def sample_temperature(self, ws_name):
```
martyngigg: I think this method should move to the `mantid_slice_algorithm` class and then the `AnalysisDataService` dependency can be removed from here.

Can we also add a comment that we have to go back to the unprojected workspace because mantid drops the log information on the way from the reduced data to the projection when going via `SofQW`.

mslice/models/slice/matplotlib_slice_plotter.py
```diff
@@ -40,6 +44,77 @@ def _getDisplayName(self, axisUnits, comment=None):
...
+            pass
+        else:
+            return sample_temp
+        if isinstance(sample_temp, str):
+            sample_temp = self.get_sample_temperature_from_string(sample_temp)
+        if isinstance(sample_temp, np.ndarray) or isinstance(sample_temp, list):
+            sample_temp = np.mean(sample_temp)
+        return sample_temp
+
+    def get_sample_temperature_from_string(self, string):
```
martyngigg: Same with this method - I think we should move it to `mantid_slice_algorithm`

mslice/plotting/plot_window/plot_window.ui
```diff
@@ -153,6 +159,39 @@
...
+  </action>
+  <action name="actionChi_Q_E">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="checked">
+    <bool>false</bool>
+   </property>
+   <property name="text">
+    <string>chi''(Q,E)</string>
```
martyngigg: This is minor but for consistency can we capitalize this to `Chi''`

mslice/plotting/plot_window/plot_window.ui
```diff
@@ -153,6 +159,39 @@
...
+  </action>
+  <action name="actionChi_Q_E_magnetic">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="checked">
+    <bool>false</bool>
+   </property>
+   <property name="text">
+    <string>chi''(Q,E) magnetic</string>
```
martyngigg: Similar, this is minor but for consistency can we capitalize this to `Chi''`

mslice/presenters/slice_plotter_presenter.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, slice_view, slice_plotter):
         self._slice_plotter = slice_plotter
         colormaps = self._slice_plotter.get_available_colormaps()
         self._slice_view.populate_colormap_options(colormaps)
+        self._sample_temp_fields = []
```
martyngigg: I don't think this is used?

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -32,14 +31,35 @@ def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing, norm_to_o
...
+        plot = [None, None, None]
+        plot[0] = plot_data
+        plot[1] = self.compute_chi(plot_data, sample_temp, y_axis)
+        plot[2] = self.compute_chi_magnetic(plot[1])
+        return plot, boundaries
+
+    def compute_chi(self, scattering_data, sample_temp, y_axis):
+        if sample_temp is None:
+            return None
+        kBT = sample_temp * constants.value('Boltzmann constant in eV/K') * 1000  # convert to meV/K
```
martyngigg: I would suggest extracting `constants.value('Boltzmann constant in eV/K') * 1000` out to a module-level variable to avoid `scipy` repeatedly looking up the value in its dictionary. Something like

```
KB_MEV = constants.value('Boltzmann constant in eV/K') * 1000
```

at the top of the module would suffice.

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -32,14 +31,35 @@ def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing, norm_to_o
...
-        plot_data = np.ma.masked_where(np.isnan(plot_data), plot_data)
         # rot90 switches the x and y axis to to plot what user expected.
         plot_data = np.rot90(plot_data)
         self._workspace_provider.delete_workspace(thisslice)
         boundaries = [x_axis.start, x_axis.end, y_axis.start, y_axis.end]
         if norm_to_one:
             plot_data = self._norm_to_one(plot_data)
-        return plot_data, boundaries
+        plot = [None, None, None]
+        plot[0] = plot_data
```
martyngigg: For larger datasets it might be a problem that we are calculating all 3 up front - this would be a waste if they are never used. Can we calculate the `chi` values upon first use and then cache?

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -32,14 +31,35 @@ def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing, norm_to_o
...
+        plot[1] = self.compute_chi(plot_data, sample_temp, y_axis)
+        plot[2] = self.compute_chi_magnetic(plot[1])
+        return plot, boundaries
+
+    def compute_chi(self, scattering_data, sample_temp, y_axis):
+        if sample_temp is None:
+            return None
+        kBT = sample_temp * constants.value('Boltzmann constant in eV/K') * 1000  # convert to meV/K
+        energy_transfer = np.arange(y_axis.start, y_axis.end, y_axis.step)
+        signs = np.sign(energy_transfer)
```
martyngigg: This is nice but do we have a problem if the `energy transfer==0`? The `numpy` docs say this returns `0` but we would like it to follow the `>0` case.

MatthewBowles:
@martyngigg Those are fixed now.

mducle:
Thanks for this @MatthewBowles but I think that the expression for the energy transfer you used (using the axis start, stop, and step size) is somehow not correct - it seems to put the zero energy transfer in the wrong position - in the picture below, the dark stripe should be at 0meV

mar21335_ei60 00mev_qe

MatthewBowles:
@mducle Like this?
mar21335_ei60 00mev_qe

martyngigg:
@MatthewBowles I was just having another look and I noticed that any some plot options, particularly log scaling on the color axis don't get preserved when flicking between the intensity types. If you set a log scale for the colour bar and then flick to `Chi''` you'll be reverted to a linear scale. Can we fix that?

mducle:
Yes. That looks much better!

MatthewBowles:
@martyngigg The colorbar log and the x/y axis ranges are preserved between intensity types now.

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -48,6 +52,74 @@ def is_slice_figure(self):
...
+            self.change_slice_plot(self.colorbar_range, cbar_log)
+            self.x_range = x_range
+            self.y_range = y_range
+            self.title = title
+            self.canvas.draw()
+        else:
+            action.setChecked(True)
+
+    def temp_handler(self, slice_plotter_method, previous):
+            try:
```
martyngigg: The indentation is off for this method

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -48,6 +52,74 @@ def is_slice_figure(self):
...
+                slice_plotter_method(self.ws_title)
+            except ValueError:  # sample temperature not yet set
+                try:
+                    field = self.ask_sample_temperature_field(str(self.ws_title))
+                except RuntimeError:  # if cancel is clicked, go back to previous selection
+                    self.intensity_selection(previous)
+                    return 'cancelled'
```
martyngigg: I think using a string like this could make this a little brittle as the string is defined in 2 places. How about just using `True`/`False` return values for the caller to represent whether the operation was a success?

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -48,6 +52,74 @@ def is_slice_figure(self):
...
+
+    def show_intensity_plot(self, action, slice_plotter_method, temp_dependent):
+        if action.isChecked():
+            previous = self.intensity_selection(action)
+            cbar_log = self.colorbar_log
+            x_range = self.x_range
+            y_range = self.y_range
+            title = self.title
+            if temp_dependent:
+                if self.temp_handler(slice_plotter_method, previous) is 'cancelled':
```
martyngigg: When comparing values it is better to use `==` rather than `is` in Python.

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -48,6 +52,74 @@ def is_slice_figure(self):
     def is_cut_figure(self):
         return not bool(self.canvas.figure.gca().get_images())
 
+    def add_slice_plotter(self, slice_plotter):
+        self.slice_plotter = slice_plotter
+        self.menuIntensity.setDisabled(False)
+        self.ws_title = self.title
+        self.actionS_Q_E.triggered.connect(partial(self.show_intensity_plot, self.actionS_Q_E,
```
martyngigg: Nice use of `partial`!

MatthewBowles:
@martyngigg Fixed :+1:

PR #168, opened 2017-11-09
Intensity - Add d2sigma/dOmega.dE option

### LINKED ISSUE #167: Intensity - d2sigma/dOmega.dE
Add new option to Intensity menu for slices.

d2sigma/dOmega.dE = multiply the S(q,w) intensity by ( kf / ki ) where:
ki is the incident wavevector: `ki==sqrt(Ei*2*m_n)/hbar` where m_n is the neutron mass
kf is obtained from the energy transfer `kf==sqrt((Ei-EN)*2*m_n)/hbar`

MatthewBowles:
**Not to be merged until after #166**

Adds an additional option to the Intensity menu for slices.

**To test:**
- Display slice
- Click 'Intensity'
- Click `D2sigma/dOmega.dE`

Fixes #167

### DIFF
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -1,6 +1,5 @@
 from __future__ import (absolute_import, division, print_function)
 import numpy as np
-from scipy import constants
 
 from mantid.simpleapi import BinMD
 from mantid.api import IMDEventWorkspace
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -49,11 +49,11 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
         if sample_temp is None:
             return None
         kBT = sample_temp * KB_MEV
-        energy_transfer = np.arange(y_axis.end, y_axis.start, -y_axis.step)
+        energy_transfer = np.linspace(y_axis.end, y_axis.start, self._get_number_of_steps(y_axis))
         return np.exp(-energy_transfer / kBT)
 
     def compute_chi(self, scattering_data, boltzmann_dist, y_axis):
-        energy_transfer = np.arange(y_axis.end, y_axis.start, -y_axis.step)
+        energy_transfer = np.linspace(y_axis.end, y_axis.start, self._get_number_of_steps(y_axis))
         signs = np.sign(energy_transfer)
         signs[signs == 0] = 1
         chi = (signs + (boltzmann_dist * -signs))[:, None]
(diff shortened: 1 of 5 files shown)


martyngigg:
@MatthewBowles We've got some conflicts on this one now.

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -65,6 +65,19 @@ def compute_chi_magnetic(self, chi):
         chi_magnetic = chi / 291
         return chi_magnetic
 
+    def compute_d2sigma(self, scattering_data, workspace, y_axis):
+        Ei = self._workspace_provider.get_EFixed(self._workspace_provider.get_workspace_handle(workspace))
```
martyngigg: This feels messy and we're also doing similar things with the temperature so it would be good to clean this up. Maybe we need some centralized functions?

MatthewBowles:
@martyngigg It's not zeroes, the values are just very small (and therefore not displayed...)

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -65,15 +66,24 @@ def compute_chi_magnetic(self, chi):
         chi_magnetic = chi / 291
         return chi_magnetic
 
+    def compute_d2sigma(self, scattering_data, workspace, y_axis):
+        Ei = self._workspace_provider.get_EFixed(self._workspace_provider.get_parent_by_name(workspace))
+        if Ei is None:
+            return None
+        ki = np.sqrt(Ei*2*constants.neutron_mass) / HBAR_MEV
+        energy_transfer = np.arange(y_axis.end, y_axis.start, -y_axis.step)
+        kf = np.sqrt(((Ei - energy_transfer)*2*constants.neutron_mass) / HBAR_MEV)[:, None]
```
martyngigg: I'm not sure this is right. I don't think the `sqrt` should encompass the `hbar`? I suspect this is the reason the values are so small. I would expect that `kf/ki` would be `1` `deltaE=0` and then decrease as `deltaE` increases.

Further to this I think we might be as well to define a conversion constant for `E`->`K` at module level so that we do that calculation once, i.e.

```
E_TO_K = np.sqrt(2*constants.neutron_mass)/HBAR_MEV
```

Then

```
ki = np.sqrt(Ei)*E_TO_K
kf = np.sqrt(Ei - deltaE)*E_TO_K
etc
```

mducle:
@martyngigg @MatthewBowles The displayed data looks good to me - also the "Symmetrised S(Q,E)" option, which has already been merged.

However, I have found another problem with the calculation of the energy transfer - there is an off-by-one error using the attached nxs file and the default bins where the calculated `energy_transfer` used in the calculation of the Boltzmann factor has one more element than the actual slice. This is because `np.arange` is used with a step size to calculate the `energy_transfer` from `y_axis` (which has the start, end and step values). However, the actual slice uses `BinMD` which takes as its input the start and end values and the number of steps. The number of steps is calculated from the step size using `int()` which implicitly rounds down, whereas I guess `arange` implicitly rounds up.

[MAR22017_Ei10.00meV.nxs.zip](https://github.com/mantidproject/mslice/files/1471324/MAR22017_Ei10.00meV.nxs.zip)

MatthewBowles:
I've swapped `arange` for `linspace` to get around this problem.

mducle:
Cool. Everything LGTM now. :shipit:

Issue #163, opened 2017-10-24
Informational overplot - recoil and powder lines


mducle:
This should populate the `Information` menu option with two submenus:

1. Recoil lines
2. Bragg peaks

The first should lead to a submenu with options:

1. Hydrogen
2. Deuterium
3. Helium
4. Arbitrary nuclei

The second should have the following submenu:

1. Aluminium
2. Copper
3. Niobium
4. Tantalum
5. CIF file

----

## Recoil

In addition to the scattering processes which the neutrons excite collective motions of an atomic or electronic lattice (phonons or magnons) or excite electronics between different energy levels, the neutron may also elastically "recoil" from an atom in the sample. All its kinetic energy and momentum is transfered to the atom, so that (using the quantum mechanical expression for the kinetic energy), the energy transfer and momentum transfer is: `hbar**2 q**2 / 2 m_a` where `m_a` is the mass of the atom from which the neutron is recoiling. This can be expressed approximately as `m_r m_n`, where `m_r` is the relative mass (==1 for Hydrogen, ==2 for Deuterium, ==4 for Helium), with `m_n` being the mass of a nucleon (neutron or proton).

Recoil gives a signal at high momentum and high energy transfers which may be mistaken for an excitation of interest (a molecular vibration or high energy electronic transition), so it would be useful to be able to plot where the recoil line should occur.

----

## Powder lines

In addition to the sample of interest, there is often other materials in the neutron beam which can generate a signal in the data. These are usually part of the sample environment and/or sample containers. Most often these will be made of Aluminium, Vanadium or Copper. These materials will mostly contribute Bragg scattering at the elastic line, but it would be useful to be able to distinguish these peaks from that of the sample.

Vanadium scatters mostly incoherently so does not contribute much Bragg scattering. In addition, some high temperature furnaces use parts made from Niobium or Tantalum which have higher melting points.

The d-spacing for a given crystal structure (e.g. from a CIF) can be computed in Mantid using:

http://docs.mantidproject.org/nightly/concepts/CrystalStructureAndReflections.html

[...]

PR #177, opened 2017-11-14
Informational overplot - recoil lines

MatthewBowles:
Plot recoil lines over slices (from issue #163).

Added options to the Information -> Recoil lines menu on slices to toggle Hydrogen, Deuterium or Helium recoil lines. There is also an option to plot any other atom by entering the relative mass.

### DIFF
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -120,10 +120,10 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
         sample_temp = float(''.join(c for c in k_string if c.isdigit()))
         return sample_temp
 
-    def compute_recoil_line(self, x_axis, y_axis, relative_mass=1):
+    def compute_recoil_line(self, x_axis, relative_mass=1):
         momentum_transfer = np.arange(x_axis.start, x_axis.end, x_axis.step)
-        line = np.square(momentum_transfer * 1.e10 * constants.hbar ) / (2 * relative_mass * constants.neutron_mass) / \
-               (constants.elementary_charge / 1000)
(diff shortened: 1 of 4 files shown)


martyngigg:
@MatthewBowles We've got a few conflicts here now.

martyngigg:
@MatthewBowles I think this is a good start. I think, though, that we will need someway of marking what the lines mean. If I had all 3 hydrogen isotope lines I just have 3 green lines and it's hard to remember what is what. I think they should certainly be different colours and maybe then we could add a legend?

Or the other option would be try to annotate the line with some text. @mducle Any ideas for how you would like to see this?

mducle:
@martyngigg @MatthewBowles I think that it's ok as it is - the lines with heavier isotopes will extend further out in |Q| for a given energy transfer. However, if it's not a lot of effort, then you make them different colors and if a user plots two (or more) lines then to bring up the legends key box showing which is which.

However, I have spotted a problem with the calculation of the lines themselves - in line 113 of `mslice/models/slice/mantid_slice_algorithm.py` instead of

```
line = np.square(momentum_transfer * HBAR_MEV) / (2 * relative_mass * constants.neutron_mass)
```

it should be:

```
line = np.square(momentum_transfer * 1.e10 * constants.hbar) / (2 * relative_mass * constants.neutron_mass) / (constants.elementary_charge / 1000)
```

This is because `constants.neutron_mass` is in SI units (in kilograms) so we can't mix the units by using `hbar` in `meV` and momentum transfer in Angstrom^-1 -- we have to convert everything into SI and then convert back to `meV` (with the term in the last bracket).

martyngigg:
Thanks, @mducle. I hadn't gotten to checking the formulae yet.

@MatthewBowles Would the different line colours and legend be much work?

MatthewBowles:
@martyngigg Not at all, I can do that.

PR #179, opened 2017-11-17
Refactored cut parameter cache code

### LINKED ISSUE #172: Cut parameters cache bugs
There are several bugs associated with this issue, all to do with the behaviour of the cut parameters caching feature which is supposed to remember previously entered values of the start, end, step size and integration ranges of a cut.

1. If the workspace is renamed, the cached cut parameters are lost. If a workspace is renamed to a previously existing workspace, its cached parameters are used instead of the correct ones.
2. After setting the parameters for one type of cut, selecting a new workspace which is of the same type (e.g. Q-E) but has not had its cut parameters set, should copy the set parameters of the previously selected workspace onto the newly selected workspace. This does not work if a non-cut workspace is selected in the mean time - but it should do.
3. If the user has just selected a workspace but have not set the parameters, this "empty" set of parameters is remembered, which disables the parameter copying mechanism above - this behaviour should be modified so that "empty" parameters are not remembered.

Issue 2 and 3 can be addressed by changes to the current code. However, issue 1 would need a large refactoring - because at present the `CutPresenter` where the cached parameters are stored (in a `dict` indexed by the workspace name) does not know about the `Rename` operation (handled by the `WorkspaceManagerPresenter` and `WorkspaceProvider`). 

The simplest option is to modify `WorkspaceProvider` (a model) to know about `CutPresenter` (a presenter) and to `pop` the dict entry when a workspace is renamed. However, this is against MVP principles since the model would change a property of a presenter.

Perhaps the most elegant option is to use the new `*Workspace` wrapper classes designed for the CLI to store the cached information. This is access via `WorkspaceProvider` to which all the presenters currently have access. This option might have to wait until after further work on the CLI is completed.

mducle:
The code to store cut parameters previously defined by the user in the `Cut` tab has now been moved into the `MantidWorkspaceProvider` class from the `CutPresenter` class. This means that workspace operations (especially `RenameWorkspace`) will not break the caching system (because the parameters are stored as `dict` entries keyed by the workspace name).

**To test:**

Load several input files and calculate projections for them (making sure they are the same type, e.g. all Q-E or 2theta-E). Select one of the projected workspace and go to the cut axis and put some numbers into the fields there. Now click on another projected workspace and check the numbers in the fields remain unchanged. Click on a non-projected workspace and check the numbers are cleared and fields are disabled. Click again on a previously selected projected workspace and check the numbers are the same as what you previously inputed.

Now change the axis type (e.g. from `MomentumTransfer` to `DeltaE`) and put in some other numbers. Now click on another projected workspace and check that the numbers are copied over (remain unchanged). Click on a non-projected workspace and check the numbers are cleared and fields are disabled. Click on any of the previously selected workspaces and check that the cut parameters (including the axis type you last left it on) are correct.

Rename one of the projected workspaces and check that the cut parameters are still ok.

Fixes #172

### DIFF
--- a/mslice/models/cut/cut_algorithm.py
+++ b/mslice/models/cut/cut_algorithm.py
@@ -26,3 +26,12 @@ class CutAlgorithm(object):
 
     def set_workspace_provider(self, workspace_provider):
         pass
+
+    def set_saved_cut_parameters(self, workspace, axis, parameters):
+        pass
+
+    def get_saved_cut_parameters(self, workspace, axis=None):
+        pass
+
+    def is_axis_saved(self, workspace, axis):
+        pass
--- a/mslice/models/cut/mantid_cut_algorithm.py
+++ b/mslice/models/cut/mantid_cut_algorithm.py
@@ -124,6 +124,15 @@ class MantidCutAlgorithm(AlgWorkspaceOps, CutAlgorithm):
         cut_axis.step = (cut_axis.end - cut_axis.start)/float(cut_binning[-1])
         return cut_axis, integration_range, is_norm
 
+    def set_saved_cut_parameters(self, workspace, axis, parameters):
+        self._workspace_provider.setCutParameters(workspace, axis, parameters)
+
+    def get_saved_cut_parameters(self, workspace, axis=None):
+        return self._workspace_provider.getCutParameters(workspace, axis)
+
+    def is_axis_saved(self, workspace, axis):
+        return self._workspace_provider.isAxisSaved(workspace, axis)
+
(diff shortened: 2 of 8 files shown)


martyngigg:
@mducle This has developed a merge conflict.

mducle:
@martyngigg Actually the behaviour we agreed on was what was intended, but I forgot to add a `clear_input_fields` call which caused the bug because it failed to clear the fields when switching from the `_QE` to `_EQ` types. Should now be fixed.

martyngigg:
Thanks, @mducle. That looks good now.

PR #182, opened 2017-11-22
Informational overplot - powder lines

### LINKED ISSUE #163: Informational overplot - recoil and powder lines

MatthewBowles:
Plot powder lines over slices.
Populates Information -> Bragg peaks menu on slice window with Aluminium, Copper, Niobium and Tantalum options. A crystal structure can also be loaded from a CIF file.

Fixes #163

### DIFF
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -1,7 +1,7 @@
 from __future__ import (absolute_import, division, print_function)
 import numpy as np
 
-from mantid.simpleapi import BinMD
+from mantid.simpleapi import BinMD, ConvertUnits
 from mantid.api import IMDEventWorkspace
 from scipy import constants
 
@@ -18,7 +18,7 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
     def __init__(self):
         self._workspace_provider = MantidWorkspaceProvider()
 
-    def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing, norm_to_one, sample_temp):
+    def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing, norm_to_one):
         workspace = self._workspace_provider.get_workspace_handle(selected_workspace)
         assert isinstance(workspace, IMDEventWorkspace)
         self._fill_in_missing_input(x_axis, workspace)
@@ -31,13 +31,13 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
         y_dim = workspace.getDimension(y_dim_id)
         xbinning = x_dim.getName() + "," + str(x_axis.start) + "," + str(x_axis.end) + "," + str(n_x_bins)
         ybinning = y_dim.getName() + "," + str(y_axis.start) + "," + str(y_axis.end) + "," + str(n_y_bins)
-        thisslice = BinMD(InputWorkspace=workspace, AxisAligned="1", AlignedDim0=xbinning, AlignedDim1=ybinning)
-        # perform number of events normalization then mask cells where no data was found
+        thisslice = BinMD(InputWorkspace=workspace, AxisAligned="1", AlignedDim0=xbinning, AlignedDim1=ybinning,
+                          StoreInADS=False)
+        # perform number of events normalization
         with np.errstate(invalid='ignore'):
             plot_data = thisslice.getSignalArray() / thisslice.getNumEventsArray()
         # rot90 switches the x and y axis to to plot what user expected.
         plot_data = np.rot90(plot_data)
-        self._workspace_provider.delete_workspace(thisslice)
         boundaries = [x_axis.start, x_axis.end, y_axis.start, y_axis.end]
         if norm_to_one:
             plot_data = self._norm_to_one(plot_data)
@@ -125,6 +125,10 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
         line = np.square(momentum_transfer * 1.e10 * constants.hbar) / (2 * relative_mass * constants.neutron_mass) /\
             (constants.elementary_charge / 1000)
         return momentum_transfer, line
+
+    def compute_powder_line(self, ws_name, x_axis, y_axis, element):
+        # assume x_axis is 2theta for now
+        return # ...
 
     def _norm_to_one(self, data):
         data_range = data.max() - data.min()
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -160,7 +160,7 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
             return ws.sample().getCrystalStructure()
         else:
             return CrystalStructure(crystal_structure[element][0], crystal_structure[element][1],
-                                         crystal_structure[element][2])
+                                    crystal_structure[element][2])
 
     def _compute_powder_line_degrees(self, ws_name, theta_axis, element, efixed, cif_file):
         wavelength = np.sqrt(E2L / efixed)
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -18,7 +18,6 @@ crystal_structure = {'Copper': ['3.6149 3.6149 3.6149', 'F m -3 m', 'Cu 0 0 0 1.
                      'Aluminium': ['4.0495 4.0495 4.0495', 'F m -3 m', 'Al 0 0 0 1.0 0.05'],
                      'Niobium': ['3.3004 3.3004 3.3004', 'I m -3 m', 'Nb 0 0 0 1.0 0.05'],
                      'Tantalum': ['3.3013 3.3013 3.3013', 'I m -3 m', 'Ta 0 0 0 1.0 0.05']}
-MEV2J = 1 / (4000 * np.pi * constants.epsilon_0) # coulomb constant / 1000
 
 
 class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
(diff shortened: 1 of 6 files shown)


mducle:
@MatthewBowles I think instead of the hiding lines code (which seems a bit complex), you could change the `x` and `y` values in `compute_powder_line` from being a list of np.arrays to being just single lists with pairs of x and y values separated by NaNs. This will be plotted as a single curve but with breaks, e.g.

```
        efixed = self._workspace_provider.get_EFixed(self._workspace_provider.get_parent_by_name(ws_name))
        if x_axis.units == 'MomentumTransfer':
            x0 = self._compute_powder_line_momentum(ws_name, x_axis, element)
        elif x_axis.units == 'Degrees':
            x0 = self._compute_powder_line_degrees(ws_name, x_axis, element, efixed)
        else:
            raise RuntimeError("units of x_axis not recognised")
        x = sum([[xv, xv, np.nan] for xv in x0], [])
        y = sum([[efixed / 20,  -efixed / 20, np.nan] for xv in x0], [])
        return x, y
```

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -12,13 +13,19 @@
 KB_MEV = constants.value('Boltzmann constant in eV/K') * 1000
 HBAR_MEV = constants.value('Planck constant over 2 pi in eV s') * 1000
 E_TO_K = np.sqrt(2*constants.neutron_mass)/HBAR_MEV
+E2L = 1.e23 * constants.h**2 / (2 * constants.m_n * constants.e)  # energy to wavelength conversion E = h^2/(2*m_n*l^2)
+crystal_structure = {'Copper': ['3.6149 3.6149 3.6149', 'F m -3 m', 'Cu 0 0 0 1.0 0.05'],
+                     'Aluminium': ['4.0495 4.0495 4.0495', 'F m -3 m', 'Al 0 0 0 1.0 0.05'],
+                     'Niobium': ['3.3004 3.3004 3.3004', 'I m -3 m', 'Nb 0 0 0 1.0 0.05'],
+                     'Tantalum': ['3.3013 3.3013 3.3013', 'I m -3 m', 'Ta 0 0 0 1.0 0.05']}
+MEV2J = 1 / (4000 * np.pi * constants.epsilon_0) # coulomb constant / 1000
```
mducle: I don't think `MEV2J` is used anywhere - maybe it can be removed?

Also, this is not the right conversion factor - `1/(4* np.pi * constants.epsilon_0)` is indeed the "Coulomb constant" which comes from Coulomb's Law, but actually what we need is the "elementary electronic charge" which is 1.602e-19 Coulombs (the unit) - and denoted by `constants.e` or `constants.elementary_charge`. Sorry about the confusion but amongst physics neutron scatterers this is often refered to (incorrectly) as the "Coulomb constant" (because of the unit).

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -31,29 +38,29 @@ def compute_slice(self, selected_workspace, x_axis, y_axis, smoothing, norm_to_o
         y_dim = workspace.getDimension(y_dim_id)
         xbinning = x_dim.getName() + "," + str(x_axis.start) + "," + str(x_axis.end) + "," + str(n_x_bins)
         ybinning = y_dim.getName() + "," + str(y_axis.start) + "," + str(y_axis.end) + "," + str(n_y_bins)
-        thisslice = BinMD(InputWorkspace=workspace, AxisAligned="1", AlignedDim0=xbinning, AlignedDim1=ybinning)
-        # perform number of events normalization then mask cells where no data was found
+        thisslice = BinMD(InputWorkspace=workspace, AxisAligned="1", AlignedDim0=xbinning, AlignedDim1=ybinning,
+                          StoreInADS=False)
```
mducle: @martyngigg `StoreInADS` is only in nightly at the moment I think... I know that we're targetting MSlice at the nightly builds for now, and it uses the nightly build on all the instrument computers and on isiscompute, but I'm concerned that some people might be running it on the stable releases...

I guess we can leave it as it is and see if anyone complains...?

For the future, I guess we should target the current stable version from Mantid 3.12 onwards - since we've got a rough beta release deadline of Feb 2018, when 3.12 will come out?

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -71,16 +74,23 @@ def add_slice_plotter(self, slice_plotter):
...
+        self.actionHelium.triggered.connect(partial(self.toggle_overplot_line, self.actionHelium, 4, True))
         self.actionArbitrary_nuclei.triggered.connect(self.arbitrary_recoil_line)
-
-    def toggle_recoil_line(self, action, relative_mass):
+        self.actionAluminium.triggered.connect(partial(self.toggle_overplot_line, self.actionAluminium,
+                                                       'Aluminium', False))
+        self.actionCopper.triggered.connect(partial(self.toggle_overplot_line, self.actionCopper, 'Copper', False))
+        self.actionNiobium.triggered.connect(partial(self.toggle_overplot_line, self.actionNiobium, 'Niobium', False))
+        self.actionTantalum.triggered.connect(partial(self.toggle_overplot_line, self.actionTantalum,
+                                                      'Tantalum', False))
```
mducle: For me for some reason `partial` passes `True` for the last argument (`extra_info`) instead of leaving as the default (`None`) - so the `Aluminium` etc lines gives me an error (since it tries to find a CIF file called `True`. So I had to add another `False` to the end of these lines...

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -89,16 +99,21 @@ def arbitrary_recoil_line(self):
             self.arbitrary_nuclei, confirm = QInputDialog.getInt(self, 'Arbitrary Nuclei', 'Enter relative mass:')
             if not confirm:
                 return
-        self.toggle_recoil_line(self.actionArbitrary_nuclei, self.arbitrary_nuclei)
+        self.toggle_overplot_line(self.actionArbitrary_nuclei, self.arbitrary_nuclei, True)
+
+    def cif_file_powder_line(self):
+        cif_path = str(QFileDialog().getOpenFileName(self, 'Open CIF file', '/home', 'Files (*.cif)'))
+        key = path.basename(cif_path).rsplit('.')[0]
+        self.toggle_overplot_line(self.actionCIF_file, key, False, extra_info=cif_path)
```
mducle: Because the key is the CIF file name, and this is asked even on an "uncheck" action, if the user does not select the same CIF file, `hide_overplot_line` does not know which line to remove. 

On the plus side, this lets users add powder lines from more than one CIF at a time; on the minus side, it's impossible to remove the lines without replotting the slices.

For the moment, I think if you just add a check for `self.actionCIF_file.isChecked()` and only if it it is checked to ask the user for a file, this would be fine. (E.g. it restricts the user to plotting single CIF file at a time... but we can change to code later if the ability to plot more than one CIF file is requested).

mslice/models/slice/mantid_slice_algorithm.py
martyngigg: @mducle True. I think this is something that we'll want to sort properly when the packaging is sorted out. When we do that we can properly specify a minimum version. When your fixes for Rebin2D etc get in we will want those too I think?

If this is the only place we are using `StoreInADS` then maybe we should take it out for now and once we have 3.12 as a minimum we can tidy those places up?

MatthewBowles: I agree, it's better to be safe since it's a small change. It'll be easy to refactor once 3.12 is out.

mslice/plotting/plot_window/plot_figure.py
MatthewBowles: Apparently the action also passes in the checkbox status as True/False so this was getting confused for the default parameter. Should be ok now.

mducle:
LGTM!

PR #188, opened 2017-12-18
Plot interactivity - cuts

### LINKED ISSUE #180: Plot window interactivity
It would be nice to make plot windows more interactive - e.g. clicking on axes labels or legends allows them (text, fonts) to be changed rather than going through the plot options menu.

This could be implemented using Matplotlib's event handling. See here for details:
https://matplotlib.org/users/event_handling.html

It is possible to detect when the user clicks on a certain line or legend using the object picking. This could then display the various line options. 
Clicking on the axis/title can be detected by using MouseEvents. These can notify the plot figure when the mouse is clicked and give the figure coordinates. For example, it may be determined that if the x co-ordinates of the click are in the range 30-60, then the x-axis label was clicked. The figure size will have to be factored into this since resizing the window will affect the coordinates.

MatthewBowles:
**Not to be merged until after #187**

Features of a cut plot can now be changed by clicking on them (issue #180). This includes labels,  axis ranges, and the actual lines.

**To test:**
-  Plot a cut
- Click a feature (axis labels or ranges, lines)
- Change settings in the dialog
- Click OK
- Check that the plot is updated with new settings
- Repeat for each feature

Fixes #180


mslice/plotting/plot_window/cut_plot.py
```diff
@@ -0,0 +1,314 @@
...
+        self._canvas = canvas
+        self._cut_plotter = cut_plotter
+        self._lines_visible = {}
+        self._legends_shown = True
+        self._legends_visible = []
+        self._legend_dict = {}
+        self._lines = self.line_containers()
+        plot_figure.menuIntensity.setDisabled(True)
+        plot_figure.menuInformation.setDisabled(True)
+        np.seterr(invalid='ignore')
```
martyngigg: Can we avoid setting this globally?

MatthewBowles: This was added to avoid errors in the console when changing the axis scale to logarithmic. I tried only ignoring the errors within the `change_axis_scale` method, but it still prints the errors in the next subsequent call to `canvas.draw()`.
I agree it would be better not to set this globally, but I'm not sure how to prevent the errors being printed (short of disabling them every single time draw is called)

PR #208, opened 2018-01-16
New DataLoader widget

### LINKED ISSUE #206: New DataLoader widget
Part of the GUI redesign #204 

New widget for loading workspaces. Consists of a file browser and buttons to load one or more files or merge several files into one workspace.

MatthewBowles:
New widget for loading workspaces. Appears on the main 'Data Loading' tab and includes a file browser.

**To test:**
- Navigate to a workspace file and verify that the workspace appears in the workspace manager after hitting the 'Load Data' button. 
- Select multiple files and press 'Load Data'. Check each workspace appears in the workspace manager individually.
- Select multiple files and press 'Load and Merge Data'. Check a single `_merged` workspace appears in the workspace manager.
- Also check the Refresh/Back buttons with the path textbox and error handling 

Fixes #206


mslice/widgets/dataloader/dataloader.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import (absolute_import, division, print_function)
+
+import os
+
+from functools import partial
+
+from qtpy.QtWidgets import QWidget, QFileSystemModel, QAbstractItemView, QMessageBox
```
martyngigg: These should use the `mslice.util.qt` wrappers too.

mslice/widgets/dataloader/dataloader.py
```diff
@@ -0,0 +1,104 @@
...
+class DataLoaderWidget(QWidget): # and some view interface
+
+    error_occurred = Signal('QString')
+
+    def __init__(self, parent=None):
+        QWidget.__init__(self, parent)
+        load_ui(__file__, 'dataloader.ui', self)
+
+        self.file_system = QFileSystemModel()
+        self.directory = QDir("C:/")
```
martyngigg: This doesn't really make sense for Linux & OSX. How about we start in `os.path.expanduser('~')` (basically $HOME)?

I also noticed a disparity on my system that when I started `mslice` it displayed `/` in the edit box but then when I hit "refresh" it went to my current working directory. I assume it should refresh what is currently displayed?

MatthewBowles:
> typing manually into the filepath and pressing return doesn't update the listing as I would expect

Strange, this works fine for me. Could you be more specific?

martyngigg:
You're right, it does work. Sorry for the confusion. I was typing an invalid path and I didn't spot the error message at the bottom - should we make those red to stand out?

MatthewBowles:
@martyngigg ok, this should be better now.

martyngigg:
Thanks, @MatthewBowles. The changes look good but with one minor issue: when I change the sort option I get an error in my terminal:
```Traceback (most recent call last):
  File "/home/mgigg/Code/git/mantidproject/mslice/mslice/widgets/dataloader/dataloader.py", line 75, in sort_files
    self.table_view.sortByColumn(column, order=column % 2)  # descending for size/modified, ascending for name/type
TypeError: sortByColumn() takes no keyword arguments
```
and it doesn't seem to do anything.

A couple of other minor points:
  - we've got an empty `Options` menu - can we remove that? I suggest that we add a standard `File` menu with a `Quit` option in its place
  - can we add a QSplitter between the tab widget and the IPython widget? - http://doc.qt.io/qt-5/qsplitter.html#details. It will allow users to resize IPython vertically should they want more space.

PR #209, opened 2018-01-17
Split WorkspaceManager into tabs

### LINKED ISSUE #207: Split WorkspaceManager into tabs
Part of GUI redesign #204 

The `WorkspaceManagerWidget` should now be divided into tabs based on workspace type. The selected tab determines whether the `slice`/`cut` or `powder` widgets are displayed. There will also be some changes to remove data loading methods and instead allow it to interact with the new `DataLoaderWidget`.

MatthewBowles:
**Not to be merged until after #208**

Implements new `WorkspaceManager` design with tabs for each workspace type. The save / rename / combine controls remain but have moved. The `Load` command is now handled by the `DataLoaderWidget`, not the `WorkspaceManager`. 

**To test:**

- Using the `Data Loading` tab, load workspaces of each type (`Workspace2D`, `MDEventWorkspace`, `MDHistoWorkspace`) and check they are placed in the correct `WorkspaceManager` tab.
- Also check that the save / rename / combine controls function as previously.

Fixes #207

### DIFF
--- a/mslice/app/mainwindow.py
+++ b/mslice/app/mainwindow.py
@@ -7,6 +7,7 @@ from mslice.presenters.main_presenter import MainPresenter
 from mslice.util.qt import load_ui
 from mslice.views.mainview import MainView
 from mslice.widgets.ipythonconsole.ipython_widget import IPythonWidget
+from mslice.widgets.workspacemanager.command import Command as ws_command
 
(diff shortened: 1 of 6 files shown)


martyngigg:
@MatthewBowles There are a few conflicts here now.

PR #217, opened 2018-01-18
Fix quick_options error when editing recoil/powder lines

### LINKED ISSUE #210: Double-clicking recoil/powder line causes error
The quick options to allow editing of the recoil/powder lines on a slice has stopped working. Double-clicking these lines now gives the following error:
```
AttributeError: 'SlicePlot' object has no attribute 'get_line_data'
```

MatthewBowles:
Recent changes broke the option to change the style of recoil / powder lines by double-clicking. This fixes the issue.

**To test:**
- Plot a slice
- Plot a line using the information menu
- Double click the line (or legend) and change the properties
- Press ok
- Verify that the line changed correctly

Fixes #210

### DIFF
--- a/mslice/plotting/plot_window/plot_options.py
+++ b/mslice/plotting/plot_window/plot_options.py
@@ -307,7 +307,7 @@ class LegendAndLineOptionsSetter(QtWidgets.QWidget):
         row3.addWidget(self.line_marker)
 
         if line_options['shown'] is not None and line_options['legend'] is not None:
-            self.show_line_label = QtWidgets.QLabel(self)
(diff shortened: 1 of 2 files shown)


martyngigg:
How feasible is it to add more tests around this area?

MatthewBowles:
The trouble is the plot figure has to be mocked out in the tests. It just checks if the presenter calls the right methods, and doesn't catch when the method doesn't exist or has been moved like in this case. 
That said, now that we have the `SlicePlot` and `CutPlot` classes (which don't have to be mocked) it should be possible to add more. I'll make an issue for it.

martyngigg:
:+1:

PR #222, opened 2018-01-19
Changes to cuts for new interface

### LINKED ISSUE #51: MSlice will not recognise a cut when the parent is not in the ADS

### LINKED ISSUE #220: Slice / cut widget adjustments
Part of GUI redesign #204 

These widgets need some adjustments to work with the GUI changes:

- Automatically save cuts to workspace
- Possibly switch tabs automatically after cut/projection operations

MatthewBowles:
**Not to be merged until after #209**

- Add some workspace type specific controls to `main_window`
- Each cut is now saved as a workspace
- Some refactoring to `mslice.models.cut` to allow `MDHistoWorkspaces` to be loaded / plotted independently. Before, the cut parameters and parent workspace were cached and re-run every time the cut was plotted. Now, the data is saved and read from the `MDHistoWorkspace`.
- Changed name of cuts that are saved to a file to include integration limits -  before, they looked like `MAR21335_Ei60.00meV_QE_cut1` with an incrementing index on the end. Now they look like this: `MAR21335_Ei60_meV_QE_cut(-1.0,1.0) `
- Automatically change tabs after load/powder/plot cut operations

**To test:**
- Test all buttons that appear on each workspace tab function as expected
- save a cut to a file, then load it in a different session and plot (without a need for the parent workspace)

Fixes #51 
Fixes #220

### DIFF
--- a/mslice/app/mainwindow.ui
+++ b/mslice/app/mainwindow.ui
@@ -31,7 +31,7 @@
         <enum>QTabWidget::Rounded</enum>
        </property>
        <property name="currentIndex">
-        <number>0</number>
+        <number>1</number>
        </property>
        <property name="iconSize">
         <size>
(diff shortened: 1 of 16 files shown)


martyngigg:
@MatthewBowles There are a few conflicts in here now.

martyngigg:
I have a question as to whether we have chose the best names for the workspace manager tabs. `2D`, `MDEvent` & `MDHisto` mean something to us but my concern is that this is mantid implementation language. I was wondering whether something like `Loaded`, `Slices` & `Cuts` might be more informative for users?
@mducle What do you think?

mslice/app/mainwindow.ui
```diff
@@ -31,142 +31,228 @@
         <enum>QTabWidget::Rounded</enum>
        </property>
        <property name="currentIndex">
-        <number>0</number>
+        <number>1</number>
```
martyngigg: Did we mean to change the default tab to the workspace manager?

MatthewBowles: Nope, I wish QtDesigner wouldn't do this...

martyngigg: I agree - it is very annoying.

mducle:
@martyngigg @MatthewBowles I think for now the names are ok - the instrument scientists know what they mean. For the beta release we should change it though. I was thinking of either `Raw` or `Loaded` for `2D`; `Projected` for `MD Event` and `Rebinned` or `Saved` for `MD Histo`.

I'll add it to the things to discuss at the Mantid meeting.

martyngigg:
@mducle Thanks.

PR #225, opened 2018-01-22
Add sample temperature by text box

### LINKED ISSUE #212: Add sample temperature via text entry
When plotting a slice and changing the Intensity option, the sample temperature field from the workspace sample logs is required as an input. It has been raised that sometimes it would be more convenient to have the option to type in the correct value instead.

MatthewBowles:
A sample temperature can now be entered by value rather than using a sample log entry. 

**To test:**
- Plot a slice
- Change Intensity option
- Enter a number instead of choosing from the list

Fixes #212

### DIFF
--- a/mslice/plotting/plot_window/slice_plot.py
+++ b/mslice/plotting/plot_window/slice_plot.py
@@ -223,9 +223,11 @@ class SlicePlot(object):
             else:
                 try:
                     temp_value = float(temp_value)
+                    if temp_value < 0:
+                        raise ValueError
                 except ValueError:
-                    self.plot_figure.error_box("Invalid value entered for sample temperature. Enter a number or a \
-                                               sample log field.")
+                    self.plot_figure.error_box("Invalid value entered for sample temperature. Enter a value in Kelvin \
+                                               or a sample log field.")
                     self.intensity_selection(previous)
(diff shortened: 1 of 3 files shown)


mslice/plotting/plot_window/slice_plot.py
```diff
@@ -213,27 +213,37 @@ def _run_temp_dependent(self, slice_plotter_method, previous):
...
     def ask_sample_temperature_field(self, ws_name):
-        if ws_name[-3:] == '_QE':
-            ws_name = ws_name[:-3]
+        ws_name = ws_name[:ws_name.rfind("_")]
         ws = AnalysisDataService[ws_name]
         temp_field, confirm = QtWidgets.QInputDialog.getItem(self.plot_figure, 'Sample Temperature',
                                                              'Sample Temperature not found. ' +
-                                                             'Select the sample temperature field:',
-                                                             ws.run().keys(), False)
+                                                             'Select the sample temperature field or enter a value:',
```
martyngigg: My only minor comment here is that I think we should mention the units that we expect the temperature to be in, i.e.
"Select the sample temperature field or enter a value in Kelvin" (assuming it is in Kelvin)

PR #228, opened 2018-01-24
Add 2D Workspaces

### LINKED ISSUE #223: Adding workspaces
This is a description of the workflow for the `Add`, `MergeMD` or `Sum Norm` buttons.

Depending on the workspace type, the `MergeRuns` (for loaded data, e.g. the `Add` button), or `MergeMD` (for projected data, e.g. in the other workspace tabs) algorithm should be used.

* If the user selects multiple workspaces, the workspace list should be passed to `MergeRuns`/`MergeMD` and the output workspace named after the first workspace in the list with "_sum" appended.

* If the user selects _one_ workspace, the a popup should appear asking them to select another workspace of the same type to be added to the first workspace. The two workspace is then passed to `MergeRuns`/`MergeMD` as a list.

MatthewBowles:
**Not to be merged until after #220**

Loaded unprojected workspaces can now be combined using the 'Add' button. This uses the `MergeRuns` algorithm. If only one workspace is selected, a dialog appears asking for another. Otherwise, all selected workspaces are passed to `MergeRuns`.

**To test:**
- Load one or more workspaces and select in the workspace manager
- Click 'Add'
- Check new workspace with _sum appended appears.

Fixes #223


MatthewBowles:
Tests for what? In `MantidWorkspaceProvider` it's worth testing `subtract` since it uses multiple algorithms and could give errors, but `add_workspace_runs` just calls `MergeRuns` which is tested elsewhere.

martyngigg:
I still think simple things are worth testing. The tests are there to ensure that the behaviour expected of that method doesn't change in the future if it were refactored to do something different - `MergeRuns` is just the current implementation of how to add things.

PR #231, opened 2018-01-26
Interactive cuts

### LINKED ISSUE #230: Interactive Cuts
Allow the user to plot a dynamically updating cut, similar to the Mantid SliceViewer.

Cuts should be parallel to the x- or y- axes only (not diagonal). **Current** plan for how it should work is like this:

1. Select "Interactive cut" in a menu option in the 2D slice window.
2. The cursor turns into a cross-hair and you draw the line you want to cut along (like in SliceViewer), which will have a default thickness (maybe 5% of the range of the relevant axis).
3. It plots a box showing the extent of the cut just defined and opens a new window with the 1D cut itself. If latency is an issue, it could be improved by plotting just the actual pixel data in the 2D plot (e.g. it will not go back to the MDEventWorkspace). This means that the cut widths would have to implicitly align with the bin boundaries (e.g. the user can select an arbitrary width but what they get is the nearest bin edge). 
4a. If the user click on and drags the edges of the box, we interpret that as them wanting to change the size of the box (e.g. the integration area or limits of the cut).
4b. If the user click inside the box and drags that, we interpret it as them wanting to change the position of the box (e.g. the limits of the cuts).
5. As the user drags the box / edges, the cut in the 1D window dynamically updates.
6. There should also be a button in the 1D window called "Cut from data" which will behave like the normal cut routines (e.g. use MDSlice from the MDEventWorkspace) but with the currently defined cut limits from the box in the 2D window.

MatthewBowles:
Allow the user to specify a region of a slice and display a dynamically updating cut of that area. Includes the ability to resize the area (by dragging the sides) or move it (by clicking inside and dragging it). 

**To test:**
- Plot a slice
- Click 'Interactive cuts' in menu
- Click and drag to draw a line onto the plot
- Cut of area should appear and update as the area is moved or resized

Fixes #230


mslice/models/cut/matplotlib_cut_plotter.py
```diff
@@ -10,25 +10,46 @@
...
         plt.draw_all()
 
+    def create_quick_cut(self):
+        self.canvas = plt.gcf().canvas
+        self.axes = self.canvas.figure.gca()
+        plt.gcf().canvas.manager.add_cut_plot(self)
+        self.canvas.manager.set_as_icut()
+        self.canvas.draw()
+
+    def plot_quick_cut(self, workspace, cut_axis, integration_start, integration_end, norm_to_one, intensity_start,
```
martyngigg: Is there a reason we don't just reuse the same `plot_cut` method? A little bit of refactoring would make that possible I think and then we wouldn't miss things like the axis labels.

mslice/plotting/plot_window/plot_window.ui
```diff
@@ -84,6 +84,8 @@
    <addaction name="actionDataCursor"/>
    <addaction name="actionPlotOptions"/>
    <addaction name="actionToggleLegends"/>
+   <addaction name="actionInteractive_Cuts"/>
```
martyngigg: I think we're running into space issues now with the toolbar. In my version the `Keep`/`Make Current` buttons are now hidden away and I have to click the `>>` to make them visible. Also, on the cut preview the `Save Cut` button is again hidden but by quite a few buttons that are disabled and off no use. Is it possible to remove the disabled buttons at least?

@mducle Any thoughts? Icons are one option but there has been negative reaction to these in the past. We could at least shorten things like "Save Plot"/"Print Plot" to Save/Print. Also, we have the data cursor button but it seems a bit of a waste of the space - should we just have the coordinates activated all of the time?

mducle: @martyngigg I like the icons, but I'll ask around to see who really doesn't like them. 

I agree that we can remove the data cursor button and just have the coordinates activated all the time.

mducle: @martyngigg @MatthewBowles So I think we can replace the "standard" buttons (e.g. what's in the equivalent window in Mantid 4.0) with icons, but leave the non-standard buttons (e.g. `Keep`, `Make Current`) as text. The objection seems to be that in some cases the functionality of an iconed-button is unguessable from the icon picture... But I think we can assume that people will know what `Save`, `Print`, `Zoom`, etc. means.

If we have the same icon set as the Mantid 4.0 plot windows, it would also make the "look" of both programs more uniform - something which was an issue discussed at the meeting iirc...

MatthewBowles: @martyngigg Where can I get the icons used in Mantid?

mducle: @MatthewBowles I think Mantid 4 uses [qtawesome](https://github.com/spyder-ide/qtawesome) with [Font Awesome](https://fontawesome.com/icons?m=free) icons. The relevant code is here:

https://github.com/mantidproject/mantid/blob/master/qt/applications/workbench/workbench/plotting/toolbar.py

martyngigg: @mducle You're correct. Again though `qtawesome` is currently on in the Windows nightly build so for now we should only include that on the `new_gui` branch?

mducle: Yes. That makes sense (to have it only in `new_gui`)

MatthewBowles:
@martyngigg Could you check this again?

mducle:
This looks really nice. One usability issue I've got is that if I just click and drag inside the `RectangleSelector` but not on the centre button to drag it, it triggers the `motion_notify_event` call back but with `x1==x2` which makes `BinMD` give an error. (This is on Ubuntu, I haven't checked it on Windows).

Could you add a check in `InteractiveCut.plot_cut()` to check that `x2>x1` and `y2>y1` before calling `_cut_plotter.plot_cut()`?

MatthewBowles:
I've changed the default positions of the plot windows and added the check for the BinMD error.

mslice/plotting/plot_window/slice_plot.py
```diff
@@ -29,6 +29,8 @@ def __init__(self, plot_figure, canvas, slice_plotter):
         self._legend_dict = {}
         self.icut_event = [None, None]
         self.icut = None
+        pos = self.plot_figure.pos()
+        self.plot_figure.move(pos.x() - 375, pos.y())
```
martyngigg: Why do we need to move the slice window as well? Shouldn't the movement width be based on the window size?

MatthewBowles: Yes it probably should be based on the window size actually. We don't need to move the slice window, but I thought it would be better to keep them both somewhat central rather than having the slice in the centre and the cut on the right.

MatthewBowles:
@martyngigg That bug should be fixed now.

martyngigg:
@MatthewBowles There are some conflicts now that the new gui stuff has been merged

martyngigg:
This looks really nice now. There's just the CI failures to deal with and then I think it can be merged.

martyngigg:
Scratch that. My page hadn't updated properly so it didn't display the fixed version.

PR #233, opened 2018-02-08
Add icons to plot window

### LINKED ISSUE #232: Use icons in plot window toolbar
The plot window toolbar is getting overcrowded. Some of the more obvious options (save, print, zoom, etc)  should be replaced by icons that match those used in the new Mantid workbench, as discussed in #231.

MatthewBowles:
Replaces the text in the toolbar of the plot window with icons for the most obvious controls.

Fixes #232


MatthewBowles:
@martyngigg this should be ok now.

PR #234, opened 2018-02-08
Remove MDEventWorkspace dependency on parent

### LINKED ISSUE #226: Some operations rely on parent workspace
Due to a bug in SliceMD which does not copy the `ExperimentInfo` to the output workspace (mantidproject/mantid#21586) some operations within MSlice rely on getting this information from a parent workspace. For example, `ask_sample_temperature_field` in `SlicePlot` gets the list of sample log entries from a similarly named unprojected workspace. 
If the projected workspace is loaded independently, the parent will not be found. This leads to errors.
When SliceMD is fixed, these operations should be changed to be independent of any parent workspaces.

MatthewBowles:
Removes the search for a parent workspace before doing some operations. This was previously necessary to circumvent a bug in `SliceMD` that is now fixed. It should also be possible to load `_QE` / `_ThE` workspaces (skipping the calculate projections step) and perform all operations with them (e.g. change intensity). 

**To test:**
- Load a projected workspace directly
- Plot a slice, cut, change intensity options etc and check nothing breaks.

Fixes #226


MatthewBowles:
@martyngigg @mducle 
I want to be able to skip the calculate projections step, load a projected workspace and plot a slice immediately. At the moment, that looks like this:
image
This is because mslice doesn't know the default step size to use, so it just uses 100 bins for each axis (which is too few in this case). I am trying to find a way to get a good default step size using only the `MDEventWorkspace`.
To my understanding, this can be done in a couple of ways with a `MatrixWorkspace`:
1. Just use the original number of bins
2. Calculate with `detectorTwoTheta` - using the original scattering angles from the detectors (which is what mslice does now in  `MantidWorkspaceProvider`, even though 1 seems simpler...)

But neither of those will work with an `MDEventWorkspace` since it doesn't have access to the right data. 
Do either of you have any ideas on how to estimate a reasonable step size / number of bins using an `MDEventWorkspace`?

MatthewBowles:
Looks like this might be possible by adding python exports to `SpectrumInfo`.

martyngigg:
@MatthewBowles Do we know why [`get_limits`](https://github.com/mantidproject/mslice/blob/master/mslice/models/workspacemanager/mantid_workspace_provider.py#L40) doesn't do the calculation properly as in the case where we loaded then calculated?

MatthewBowles:
`get_limits` just reads the cache in `self._limits` if it exists. If a `MatrixWorkspace` is loaded, `ProcessLoadedWSLimits` is called and the limits are cached. `ProcessLoadedWSLimits` currently isn't compatible with any other workspace type, so nothing is cached, and `get_limits` just uses 100 steps.

The main obstacle seems to be [getting the twotheta values](https://github.com/mantidproject/mslice/blob/877543cb741f3b0500206d7e7b28362be9c97591/mslice/models/workspacemanager/mantid_workspace_provider.py#L118) used to calculate the step size but it looks like this could be done through `SpectrumInfo` if the relevant methods are exported.

martyngigg:
Oh I see. The problem here is an `MDEventWorkspace` doesn't know anything about spectra any more - it's just a collection of signal values in `|Q|/E`-space. `SpectrumInfo` won't really help as they don't exist. 

I think we've lost our link back to the original theta values so I'm not quite sure if we can pick back up what the original angular resolution was.

MatthewBowles:
>SpectrumInfo won't really help as they don't exist.

Could you explain this? `SpectrumInfo` is in `ExperimentInfo` which is copied across from the `MatrixWorkspace`. I don't see how that process would remove the `twotheta` data.

martyngigg:
@MatthewBowles My apologies, my error. I'm still thinking in instrument 1 days and I'd forgotten some bits have moved as part of the instrument 2.0 work.

@OwenArnold @SimonHeybrock Are there any plans to export the `SpectrumInfo` layers to Python so they can be accessed for this type of use case?

MatthewBowles:
I've already discussed with @SimonHeybrock about this, seems adding exports should be fine.

SimonHeybrock:
Correct. Adding exports was initially planned as part of Mantid-4.0, which was then cut in scope, so it never happened.

MatthewBowles:
Going to get this done for 1.0.0 since it'll make #255 much easier.

martyngigg:
@MatthewBowles This is still marked as being merged with `new_gui`. I assume that should be `master` now?

MatthewBowles:
That's right, thanks.

PR #236, opened 2018-02-18
Qt5 and Matplotlib2 compatibility

### LINKED ISSUE #218: MSlice not compatible with Qt5
When run on a system with `pyqt5`, MSlice fails to load a workspace.

mducle:
Changes the code to make it compatible with `Qt5` and `Matplotlib2`. In addition to changes to string handling (`Qt5` returns unicode strings), the major changes are in "views" which are not tested or in "presenter" code which call `Qt` or `Matplotlib` directly and are mocked out in tests, and hence have not been picked up by the testing scripts.

**To test:**

Run in a python/mantid environment which uses `Qt5` and `Matplotlib2`. The `mantid-framework` conda package provides this (and is the impetus for this PR). Unfortunately, the `mantid-framework` conda package is only available for Linux. On other operating systems you will need to recompile Mantid for Qt5. 

On Linux, install miniconda from the anaconda webpage and then do:

```
conda install pyqt ipython
conda install -c mantid nexus
conda install -c mantid/label/nightly mantid-framework
```

If you have mantid/python installed already (outside conda) you might have to either create a conda environment and change the `PYTHONPATH` environment variable.
then change to the mslice folder and do:

```
python scripts/start_mslice.py
```

Then check that all functionality (loading files, plotting slices / cuts, changing plot options, overplots, saving plots / ascii, etc.) works under `Qt5` / `Matplotlib2`

Fixes #218.


mducle:
@martyngigg It's an existing bug and there are also similar issues with other file dialogs... I'll fix them here anyway.

martyngigg:
That looks fine now. One more minor thing though - if I click cancel on the dialog then the "CIF file" option is still checked in the menu.

PR #240, opened 2018-02-21
Handle `displayNormalization` for event data

### LINKED ISSUE #216: Cannot load SNS reduced nxs file
Reported by an SNS user that MSlice cannot load SNS `nxs` reduced files (but loads `nxspe` files). Suspect that it is because the `nxs` file is a Mantid `MDEventWorkspace` rather than a `Workspace2D`.

mducle:
`mslice` by default plots all data using the `NumEventsNormalization` whatever the underlying data type. This is fine for histogrammed data which is converted to MD but is incorrect for event data because the normalisation has already been applied during reduction. For these data, `NoNormalization` should be used. This is why SNS event data (autoreduced nexus files, rather than autoreduced `nxspe` files which are histogrammed) is displayed incorrectly.

Edit: also removed ascii saving code in `cut_presenter.py` which is no longer used (there will be a separate PR to add saving to ascii as an option for the `Save` button in addition to the default `nxs`), and moved a function which calls a mantid algorithm from `cut_presenter.py` to `mantid_cut_algorithm.py`.

**To test:**

Load an SNS event `nxs` dataset (these are large files, so are not attached here. A test set can be found on \\Babylon5\Public\duc\mslice_test_data). Compute projection and plot the slice / make cuts and check that the data looks ok (e.g. similar to the equivalent `nxspe` file).

Fixes #216.

### DIFF
--- a/mslice/models/cut/cut_algorithm.py
+++ b/mslice/models/cut/cut_algorithm.py
@@ -6,6 +6,9 @@ class CutAlgorithm(object):
     def compute_cut_xye(self, selected_workspace, cut_axis, integration_start, integration_end, is_norm):
         pass
 
+    def get_arrays_from_workspace(self, workspace):
+        pass
+
     def is_cuttable(self, workspace):
         pass
 
--- a/mslice/models/cut/mantid_cut_algorithm.py
+++ b/mslice/models/cut/mantid_cut_algorithm.py
@@ -33,7 +33,11 @@ class MantidCutAlgorithm(AlgWorkspaceOps, CutAlgorithm):
         plot_data = self._num_events_normalized_array(cut)
         plot_data = plot_data.squeeze()
         with np.errstate(invalid='ignore'):
-            errors = np.sqrt(cut.getErrorSquaredArray())/cut.getNumEventsArray()
+            if cut.displayNormalization() == MDNormalization.NoNormalization:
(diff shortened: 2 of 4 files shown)


MatthewBowles:
@mducle This branch has merge conflicts in `cut_presenter` now.

MatthewBowles:
Actually, never mind that last one. Seems to be a problem with the axis being the wrong way around to what it expects and should probably be a separate issue since I also get it on master.

mducle:
Displaying it as `EQ` when the projection is `QE` gives the wrong axes because the `QE` values for `min`, `max` and `step` is used (e.g. the axes limits should be swapped around too). If you manually do this, it seems to be fine - was this what you meant by the unusual result? I think we need to modify the code so that when you swap the axes it swaps the default values too. But that's another issue...

In any case the data doesn't look the same for the two cases... but I think this is something in the data itself rather than in `mslice`'s handling of it - since `sliceviewer` shows the same difference...
mslice_nxs_nxspe_comparison

MatthewBowles:
Ah ok, the differences in the data must be how it's stored then. 
I think I must've mistyped one of the values because I can't reproduce the result I got before. Sorry about that... although I agree the `min/max/step` values should be swapped with the axis automatically, I'll make a separate issue for that.

PR #252, opened 2018-02-25
Interactive cut bug fix

### LINKED ISSUE #246: Cannot switch off interactive cut
When the interactive cut is selected, it pops up a cut window. If this window is closed and interactive cut mode is switched off, and a new cut is made using the `Cut` tab, this is plotted on an "interactive cut" window (e.g. without the `plot options`, `keep` and `make current` buttons). 

Also, if a cut is made using the `Cut` tab without first plotting a slice, the "interactive cut" type window (without `plot options`, `keep` and `make current` buttons) is created instead of the full cut window. However, if a slice is first plotted and then a cut the `keep` / `make current` buttons appear but the `make current` option does not work (e.g. plotting a new cut does not update the graph window - and also does not create a new plot window [e.g. clicking plot appears to do nothing].).

mducle:
Fixes two bugs in interactive (and normal) cuts.

1. Changing from interactive to normal mode did not restore the disabled `keep` / `make current` buttons.
2. If a cut is kept, the `CutPlotter` may still retain a reference to its `canvas` due to caching whereas the new plot data, due to the actions of the `FigureManager` has been sent to a new `canvas`, which is not updated. So the new plot is not shown.

**To test:**

Load a data file and make several cuts, keeping each one (two or three is enough). Then go back and `make current` one of the kept cuts and plot another cut - check that the plot updates as expected. Select another plot window to `make current`, then make a slice and click `Interactive Cut`. Check that the interactive cut is overplotting on the current plot window. Check that the `Keep` / `Make Current` buttons are hidden and a `Save` button is enabled. Move the interactive cut box around a check it works. Then click `Interactive Cut` again to disable it. Check that the `Keep` / `Make Current` buttons on the cut window are shown again.

Fixes #246.

### DIFF
--- a/mslice/models/cut/matplotlib_cut_plotter.py
+++ b/mslice/models/cut/matplotlib_cut_plotter.py
@@ -15,15 +15,10 @@ class MatplotlibCutPlotter(CutPlotter):
 
     def plot_cut(self, selected_workspace, cut_axis, integration_start, integration_end, norm_to_one, intensity_start,
                  intensity_end, plot_over):
-        if self.background is None:
-            self._create_cut()
         x, y, e = self._cut_algorithm.compute_cut_xye(selected_workspace, cut_axis, integration_start, integration_end,
                                                       norm_to_one)
         integrated_dim = self._cut_algorithm.get_other_axis(selected_workspace, cut_axis)
         legend = self._generate_legend(selected_workspace, integrated_dim, integration_start, integration_end)
-        # Note that we cannot cache the reference to canvas and axes here because under the keep/make current mechanism
-        # there is no one-to-one link between any particular CutPlotter and any CutPlot (or mpl canvas) object
-        plt.gcf().canvas.restore_region(self.background)
         self.plot_cut_from_xye(x, y, e, cut_axis.units, selected_workspace, plot_over, legend)
         plt.ylim(intensity_start, intensity_end)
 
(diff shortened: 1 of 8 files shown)


mslice/models/cut/matplotlib_cut_plotter.py
```diff
@@ -33,35 +35,36 @@ def plot_cut_from_xye(self, x, y, e, x_units, selected_workspace, plot_over, leg
...
         plt.ylabel(INTENSITY_LABEL, picker=picker)
         plt.autoscale()
+        plt.gcf().canvas.manager.add_cut_plot(self)
         if not plot_over:
             plt.gcf().canvas.manager.update_grid()
-        self.axes.draw_artist(self.canvas.figure.get_children()[1])
-        self.canvas.blit(self.canvas.figure.gca().clipbox)
+        plt.gcf().canvas.draw()
+        plt.gca().draw_artist(plt.gcf().canvas.figure.get_children()[1])
+        plt.gcf().canvas.blit(plt.gcf().canvas.figure.gca().clipbox)
```
MatthewBowles: Why do we need `draw()` here? `draw_artist` and `blit` were used to avoid calling `draw()` (which seems quite slow).

mducle:
@MatthewBowles Sorry the merge master seems to have messed things up a bit...

MatthewBowles:
If I had to guess, that's most likely due to [changes made here.](https://github.com/mantidproject/mslice/pull/247) I didn't really consider multiple plot windows when writing this PR... I was thinking about 1 slice and 1 cut at a time. My mistake there.

mducle:
No worries... hopefully it will be fixed now... I don't understand how MPL works sometimes...

mslice/models/cut/matplotlib_cut_plotter.py
```diff
@@ -33,35 +30,43 @@ def plot_cut_from_xye(self, x, y, e, x_units, selected_workspace, plot_over, leg
...
-        self.axes.draw_artist(self.canvas.figure.get_children()[1])
-        self.canvas.blit(self.canvas.figure.gca().clipbox)
+        if self.background is None:
+            self._create_cut()
+        plt.gcf().canvas.restore_region(self.background)
+        try:
+            plt.gca().draw_artist(plt.gcf().canvas.figure.get_children()[1])
+        except AttributeError:
+            plt.gcf().canvas.draw()
+        plt.gcf().canvas.blit(plt.gcf().canvas.figure.gca().clipbox)
```
MatthewBowles: This is better, but in the case where `draw()` is called, we shouldn't need to call `blit` afterwards.

PR #253, opened 2018-02-26
Save to ASCII or Matlab formats

### LINKED ISSUE #251: Option to save workspace to ASCII and Matlab
The option to save a cut to ascii was removed with the move to the GUI layout.

This is a feature that is in quite high demand, so there should be an option in the current `Save` dialog to select the file type (not just `nexus`) to allow saving to a text file. In addition, the ability to save a workspace to a Matlab `mat` file (using `scipy.io.savemat`) would be also desirable.

MatthewBowles:
Adds ability to save cuts in other file formats.
Adds dropdown to `WorkspaceManager` save button to select a file format. If saving a single file, the format can also be changed in the save dialog.
These new save options are also accessible using the save button in the plot window (although only cuts can currently be saved in the new formats, support for slices / `MDEventWorkspace` will be added in another PR).
Cuts saved as `ASCII` files can now be loaded.

**To test:**
- Save different workspace types as `.txt` or `.mat` files.
- Save a cut in the new formats using the save button in the plot window.
- Load a cut from a `.txt` file and check it plots correctly.

Fixes #251


MatthewBowles:
@mducle could you try saving a workspace in the new formats (using the `save` button in the `workspace manager`, not the plot window one yet) and check it works as expected?

mducle:
@MatthewBowles Yes, it works fine for cuts. For slices (`MDEventWorkspace`) it doesn't work (as expected) because there is no way to extract the `x`, `y`, `e`. However, the error message could be a bit more helpful than at present (`Something went wrong while trying to save`). 

I do want to be able to save slices though. The options would be:

1. Automatically save the `MDHistoWorkspace` created by `MantidSliceAlgorithm` (like for cuts) and then get the `x`, `y`, `e` from here (where `x` would actually be two arrays `x1` and `x2`) and save this. 
2. Create a slice using the default parameters and get the arrays from this. 

I'd actually like to have both options, but I think for this PR we can go with option 2, if that's ok with you?

MatthewBowles:
Could you clarify what you mean by 2? As stated, we can't get the arrays from a slice because we can't extract them.

mducle:
Sorry. I meant that option 2 would run `MantidSliceAlgorithm.compute_slice` using the default parameters which would give the signal array and boundaries. I guess it can then work out the step size from the size of the signal. We would want the error array too though. It's looking a bit complicated now, and also messes with the MVP... Maybe option 1 is better? Would you make a new PR to automatically save an `MDHistoWorkspace` of slices? (and change the behaviour of that tab to also plot slices)? If this is too much let me know and I can do this. (I also want to remove the behaviour where it switches tab to `MDHisto` after you click to plot a cut, and just leave a message saying that a cut has been saved to that tab, this is because I might want to change the cut parameter after seeing the cut and not want to have to click back on the `MDEvent` tab.).

MatthewBowles:
@mducle Yeah, I'll make a separate issue to automatically save the `MDHistoWorkspace` used in `compute_slice` as a hidden workspace and then we can use that to save the slice. I prefer that approach purely because it's a lot easier to get the errors out.
> (I also want to remove the behaviour where it switches tab to MDHisto after you click to plot a cut, and just leave a message saying that a cut has been saved to that tab, this is because I might want to change the cut parameter after seeing the cut and not want to have to click back on the MDEvent tab).

Agreed, the tab switching can be annoying sometimes. I think it would be best to somehow highlight the tabs that have new workspaces, perhaps by bolding the tab name?

mducle:
Yes. Bold names for tabs with new workspaces is a good idea.

mslice/plotting/plot_window/plot_figure.py
```diff
@@ -139,6 +140,24 @@ def print_plot(self):
             painter.drawPixmap(0,0,pixmap_image)
             painter.end()
 
+    def save_plot(self):
+        file_path = QtWidgets.QFileDialog.getSaveFileName(
+            self, filter="Image (*.png);; PDF (*.pdf);; Nexus (*.nxs);; Ascii (*.txt);; Matlab (*.mat)")
+        ext = file_path[file_path.rfind('.'):]
+        save_name = os.path.basename(file_path)
+        file_path = os.path.dirname(file_path)
+        title = self._plot_handler._ws_title
```
mducle: I think `_ws_title` should be just `title` - it gives an error otherwise for me.

Also, when I click on the save icon in a cut window, and then select a file type and just type the name of the file (without extension) it always saves a `png` whichever file type I select. This is because the `QFileDialog` is just returning the name I type without adding the file extension.

However, even when I type the filename with extension, it only saves a `png` or `pdf` - typing `nxs`, `txt`, or `mat` results in nothing being saved, a no error messages given either.

MatthewBowles:
Thanks @mducle, I think this should be fixed now.

PR #257, opened 2018-03-01
Save slices as ASCII or matlab

### LINKED ISSUE #255: Saving slices to alternative formats
As discussed in #253, to save slices to ASCII or matlab formats the x, y, e arrays are needed. Currently, the suggestion for how to do this is to save the `MDHistoWorkspace` from `compute_slice` as a hidden workspace and then follow the method used for cuts to get the arrays.

This is an extension of the saving methods implemented in #253 
Since this requires compute_slice to be called with default parameters, this depends on #234

MatthewBowles:
**Shouldn't merge until after #253**

Output a slice to ASCII or matlab. Works by caching the `MDHistoWorkspace` outputted by `BinMD` in `SliceAlgorithm`'s `compute_slice` as a hidden workspace. If no hidden workspace is found, runs `compute_slice` with default parameters.

Fixes #255

### DIFF
--- a/mslice/models/workspacemanager/file_io.py
+++ b/mslice/models/workspacemanager/file_io.py
@@ -1,6 +1,6 @@
 from __future__ import (absolute_import, division, print_function)
 import os.path
-from mantid.api import IMDEventWorkspace, IMDHistoWorkspace
+from mantid.api import IMDHistoWorkspace
 from mantid.simpleapi import CreateMDHistoWorkspace, SaveMD, SaveNexus, SaveAscii
 from mslice.util.qt.QtWidgets import QFileDialog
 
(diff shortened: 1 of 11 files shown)


mslice/models/workspacemanager/file_io.py
```diff
@@ -0,0 +1,143 @@
...
+        _save_slice_to_ascii(workspace, path)
+    elif isinstance(workspace, IMDHistoWorkspace):
+        _save_cut_to_ascii(workspace, workspace.name(), path)
+    else:
+        SaveAscii(InputWorkspace=workspace, Filename=path)
+
+
+def save_matlab(workspace, path, is_slice):
+    if is_slice:
+        x, y, e = _get_slice_mdhisto_xye(workspace)
```
mducle: Could we leave the Matlab output as a 2D matrix rather than a collapsed vector? E.g. break `_get_slice_mdhisto_xye` at line 134 (after `e = `) and return the `x`, `y`, `e` values at that point for the matlab write-out, whilst leaving the rest (the `reshape`) for the `ascii` write-out?

This is because if I want to plot a 2D colorplot in Matlab with the current loaded data, I need to `reshape` it again into 2D form, which is a bit of a pain.

MatthewBowles:
@mducle I think this should be OK now.

mslice/models/workspacemanager/file_io.py
```diff
@@ -0,0 +1,143 @@
...
+        filter = "Nexus (*.nxs);; Ascii (*.txt);; Matlab (*.mat)"
+        if save_as_image:
+            filter += ";; Image (*.png);; PDF (*.pdf)"
+        file_dialog.setNameFilter(filter)
+
+        if default_ext:
+            ext_to_qtfilter = {'.nxs': 'Nexus (*.nxs)', '.txt': 'Ascii (*.txt)', '.mat': 'Matlab (*.mat)', }
+            file_dialog.selectNameFilter(ext_to_qtfilter[default_ext])
+        if (file_dialog.exec_()):
+            path = file_dialog.selectedFiles()[0] + file_dialog.selectedFilter()[-5:-1]
```
mducle: Could you add a check here if the file name already has an extension? (If I type something like `test.nxs` in the dialog, I get a file named `test.nxs.nxs`)

MatthewBowles:
@mducle Thanks, I spotted this a while ago and forgot about it. Fixed now.

mducle:
Cool! Looks good to me now.

PR #260, opened 2018-03-09
Highlight when tab has new workspaces

### LINKED ISSUE #256: Bold WorkspaceManager tabs to show new workspaces
At the moment, the `WorkspaceManager` tabs switch automatically whenever a new workspace appears, but this can be annoying e.g. if you wanted to make multiple cuts. Instead, the tab names should be highlighted to show their contents have changed.

MatthewBowles:
Appends an asterisk to the `MD Histo` tab to show new workspaces instead of changing to it automatically.
(This was by far the easiest way to implement tab highlighting, changing font color/style is possible but requires more code.)

**To test:**
- Create a cut
- Instead of switching to `MD Histo` tab, it should appear as `MD Histo*`
- Upon clicking the `MD Histo` tab, asterisk disappears.

Fixes #256

### DIFF
--- a/mslice/widgets/workspacemanager/workspacemanager.py
+++ b/mslice/widgets/workspacemanager/workspacemanager.py
@@ -59,7 +59,9 @@ class WorkspaceManagerWidget(WorkspaceView, QWidget):
         self.tabWidget.setCurrentIndex(tab)
 
     def highlight_tab(self, tab):
-        self.tabWidget.setTabText(tab, self.tabWidget.tabText(tab) + '*')
+        tab_text = self.tabWidget.tabText(tab)
+        if not tab_text.endswith('*'):
+            self.tabWidget.setTabText(tab, tab_text + '*')
 
     def _btn_clicked(self):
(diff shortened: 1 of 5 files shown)


MatthewBowles:
@mducle Do you think this method of highlighting the tab is ok?
Is there anywhere else this should be done? it's very easy to add if so.

mslice/widgets/workspacemanager/workspacemanager.py
```diff
@@ -56,6 +58,9 @@ def current_list(self):
     def change_tab(self, tab):
         self.tabWidget.setCurrentIndex(tab)
 
+    def highlight_tab(self, tab):
+        self.tabWidget.setTabText(tab, self.tabWidget.tabText(tab) + '*')
```
mducle: Could we add a check here to see if the tab text already has an `*` in it. At the moment, if you create multiple cuts, it keeps adding a star each time. I think just one star is enough to show that a workspace has been added to that tab.

PR #262, opened 2018-03-13
Disable smoothing widget until smoothing implemented

mducle:
This just disables the `smoothing` text box until we implement smoothing - with the beta release imminent we don't want users putting numbers in the box and then seeing nothing happening (and no error messages coming up).

**To test:**

Check that the smoothing box is disabled for both slices and cuts.

No associated issue.


martyngigg:
Should we remove the box instead? Boxes that can't be activated look a like a bug to me from the outside.

mducle:
Yes. I'll `hide` the boxes; we can re-enable them later.

PR #263, opened 2018-03-13
Non-PSD Mode

### LINKED ISSUE #254: Non-PSD mode workflow
At present the MSlice workflow is to load a file as `Workspace2D` or `EventWorkspace` and then to convert it to an `MDEventWorkspace` for further manipulation (slice / cut). This is fine except that `ConvertToMD` which is used to do this can only perform a "center-point rebinning" which is not appropriate for non-PSD data. Instead the code in `MantidProjectionCalculator` first runs the `SofQWNormalisedPolygon` algorithm to do a polygon rebinning before running `ConvertToMD`.

`SofQWNormalisedPolygon` has recently been updated to be able to rebin in both `DeltaE` as well as `MomentumTransfer`. Furthermore the data volume for non-PSD data is small, so it can be used directly to construct slices and cut.

So, this issue proposes to check whether loaded workspace is of non-PSD type or not (the check will be whether the list of detector `theta` angles is monotonically increasing or not). If it is, then a new workflow using `SofQWNormalisedPolygon` rather than `BinMD` will be followed, and the "Calculate Projection" (`ConvertToMD`) step will be removed. If it is not, the current behaviour will be used.

mducle:
This PR adds a different workflow for non-PSD type data (from MARI or from other instruments using a `rings_map`). These data are defined by having detector lists in order sorted by `2theta` angles. The main characteristics is that there are few detectors in these data, and that centre-point rebinning (using  `ConvertToMD`) is not suitable. Until now, the work-around has been to use `SofQW3` to rebin the data into fine `|Q|` bins and then to convert to MD. However, this two step binning process overestimates the uncertainty (errors), because the first step involves partitioning bins. It would thus be better if the workflow includes only a single rebin.

This PR does this. Non-PSD data now do not need to go through a `Calculate Projection` step (i.e. conversion to `MDEventWorkspace`). `SofQW3` is used directly to rebin the data into a 2D slice in `|Q|` and `DeltaE` *and* also 1D cuts in these dimensions. For slices in `Degrees` (or `theta`), `ConvertSpectrumAxis` and `Rebin2D` is used (similar to the old workflow where `ConvertSpectrumAxis`, `ConvertToMD` and `BinMD` was used). 

**To test:**

Load a non-PSD input data file (e.g. `MAR21335_Ei60.00meV.nxs` or `MER30936_Ei10.00meV_rings.nxs`). Select the loaded workspace and check that the `Powder` tab is *disabled* and `Slice` and `Cut` *enabled*. Make slices and cuts as normal - check all supported features (recoil lines, powder lines, change intensity, interactive cuts, simultaneous cuts of multiple workspaces, cuts with integration widths) work as expected. 

Load a PSD input data file (e.g. `MER30936_Ei10.00meV_one2one.nxs`). Select the loaded workspace and check that `Powder` is *enabled* and `Slice` and `Cut` is *disabled*. Calculate projections and make cuts / slices / etc. and check everything works as before.

Fixes #254 .

### DIFF
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -3,7 +3,8 @@ from six import string_types
 import numpy as np
 
 from mantid.simpleapi import BinMD, LoadCIF, SofQW3, ConvertSpectrumAxis, Rebin2D
-from mantid.api import IMDEventWorkspace, MDNormalization
+from mantid.api import IMDEventWorkspace, MDNormalization, WorkspaceUnitValidator
+from mantid.dataobjects import Workspace2D
 from mantid.geometry import CrystalStructure, ReflectionGenerator, ReflectionConditionFilter
 from scipy import constants
 
@@ -224,3 +225,11 @@ class MantidSliceAlgorithm(AlgWorkspaceOps, SliceAlgorithm):
     def _norm_to_one(self, data):
         data_range = np.nanmax(data) - np.nanmin(data)
         return (data - np.nanmin(data))/data_range
+
+    def is_sliceable(self, workspace):
+        workspace = self._workspace_provider.get_workspace_handle(workspace)
+        if isinstance(workspace, IMDEventWorkspace):
+            return True
+        else:
+            validator = WorkspaceUnitValidator('DeltaE')
+            return isinstance(workspace, Workspace2D) and validator.isValid(workspace) == ''
--- a/mslice/models/slice/mantid_slice_algorithm.py
+++ b/mslice/models/slice/mantid_slice_algorithm.py
@@ -2,7 +2,7 @@ from __future__ import (absolute_import, division, print_function)
 from six import string_types
 import numpy as np
 
-from mantid.simpleapi import BinMD, LoadCIF
+from mantid.simpleapi import BinMD, LoadCIF, SofQW3, ConvertSpectrumAxis, Rebin2D
 from mantid.api import IMDEventWorkspace, MDNormalization
 from mantid.geometry import CrystalStructure, ReflectionGenerator, ReflectionConditionFilter
 from scipy import constants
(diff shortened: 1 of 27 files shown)


mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -134,7 +165,8 @@ def get_sample_temperature_from_string(self, string):
     def compute_recoil_line(self, ws_name, axis, relative_mass=1):
         efixed = self._workspace_provider.get_EFixed(self._workspace_provider.get_workspace_handle(ws_name))
         x_axis = np.arange(axis.start, axis.end, axis.step)
-        if axis.units == 'MomentumTransfer':
+        print(axis)
```
martyngigg: It looks like we have a stray `print` statement here.

mslice/presenters/cut_presenter.py
```diff
@@ -238,6 +240,11 @@ def _cut_axis_changed(self):
                 self._cut_view.populate_input_fields(saved_parameters)
         min_step = self._minimumStep[self._cut_view.get_cut_axis()]
         self._cut_view.set_minimum_step(min_step)
+        self._cut_view.update_integration_axis()
+
+
+    def _integration_axis_changed(self):
```
martyngigg: This seems to be an empty method. Is there some missing functionality?

martyngigg:
One other comment: should we be displaying the mode somewhere or is it confusing for users?

mslice/presenters/cut_presenter.py
mducle: Ah, no. It was a place-holder I forgot to delete - I was thinking of putting a callback to check the validity of the new integration axis, but then decided that the code should be in the callback for the cut axis (it repopulates the list of integration axes with allowed values).

mducle:
I guess we can put a message at the bottom of the gui. I'll ask others what they think

martyngigg:
@mducle Is this ready to go now?

mducle:
@martyngigg Yes. I've finished all the changes needed.

PR #269, opened 2018-03-13
Select and update loaded workspaces

### LINKED ISSUE #235: Workspaces can fail to update/slice correctly using the new GUI
Workspaces are only updated on changing the `2D/MD Event/MD Histo` tabs in `Workspace Manager`, not changing the `Data Loading / Workspace Manager` tabs. This means if a workspace is loaded and overwrites one with the same name, it is treated as the old workspace until the tab or selection changes.

**To reproduce:**
1. Load a workspace and calculate projections
2. Load a projected workspace with the same name and overwrite
3. Immediately plot a slice

MatthewBowles:
Avoids a bug by automatically selecting a workspace when it's loaded.

**To test:**
- Save a slice as 'slice1'
- Delete slice workspace
- Calculate projections from a different workspace and name this slice 'slice1' as well.
- Load original slice1 and overwrite.
- Slice1 should be automatically selected.
- Plot slice1 and check it displays correctly.

Fixes #235

### DIFF
--- a/mslice/presenters/data_loader_presenter.py
+++ b/mslice/presenters/data_loader_presenter.py
@@ -45,9 +45,9 @@ class DataLoaderPresenter(PresenterUtility, DataLoaderPresenterInterface):
                 else:
                     self.check_efixed(ws_name, multi)
                     self._main_presenter.show_workspace_manager_tab()
+                    self._main_presenter.update_displayed_workspaces()
                     self._main_presenter.show_tab_for_workspace(self._workspace_provider.get_workspace_handle(ws_name))
         self._report_load_errors(ws_names, not_opened, not_loaded)
-        self._main_presenter.update_displayed_workspaces()
         self._view.busy.emit(False)
 
     def check_efixed(self, ws_name, multi=False):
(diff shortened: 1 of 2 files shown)


MatthewBowles:
That was my intention, but putting it there would run even if no files were loaded and cause an error. Looking into it now.

mducle:
Looks good, works well for me.

MatthewBowles:
Moving `update_selected_workspaces` into the else block isn't ideal (since it'll run for every workspace loaded) but it should avoid any errors and bigger changes to the code so I think it's ok for now.

mducle:
@MatthewBowles What about:

```
self._main_presenter.set_selected_workspaces([ws for ws in ws_names if ws not in not_loaded + not_opened])
```

and leaving the `update_selected_workspaces` outside the `else`?

MatthewBowles:
That doesn't account for if `load` causes a `ValueError` since in that case the error is passed directly to the view and not via `not_opened` or `not_loaded`... 
I think the method needs restructuring really but I'm wary about doing that right before release.

mducle:
Ok. No problems. We can keep it as it is now, and then refactor it after the release. I don't really see any slow down in loading with the update inside the else clause anyway...

martyngigg:
I just want to check this is good to go before I hit merge?

mducle:
I think it's good now.

PR #270, opened 2018-03-13
Create unit tests for data loader

### LINKED ISSUE #238: Add unit tests for the new data loader widget/presenter
There are currently no tests for the new data loader widget after it was refactored out. Add tests similar to those that existed for the `workspace_manager_presenter.py`

MatthewBowles:
Reinstates old tests surrounding loading data for new `data_loader` class and adds some new ones.

Fixes #238

### DIFF
--- a/mslice/presenters/data_loader_presenter.py
+++ b/mslice/presenters/data_loader_presenter.py
@@ -17,13 +17,19 @@ class DataLoaderPresenter(PresenterUtility, DataLoaderPresenterInterface):
     def set_workspace_provider(self, workspace_provider):
         self._workspace_provider = workspace_provider
 
-    def load_workspace(self, file_paths, merge):
+    def load_workspace(self, file_paths, merge=False):
+        '''
+        Loads one or more workspaces.
+        :param file_paths: list of paths to files to load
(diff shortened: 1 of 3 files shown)


mslice/presenters/data_loader_presenter.py
```diff
@@ -17,13 +17,19 @@ def __init__(self, data_loader_view):
...
         self._workspace_provider = workspace_provider
 
-    def load_workspace(self, file_paths, merge):
+    def load_workspace(self, file_paths, merge=False):
+        '''
+        Loads one or more workspaces.
+        :param file_paths: list of paths to files to load
+        :param merge: boolean - whether to combine files into a single workspace
+        '''
         self._view.busy.emit(True)
```
martyngigg: This is better in a new PR I think but this type of switch something on then off at the end of a block would be a good use case for a context manager? For an example see https://github.com/mantidproject/mantid/blob/master/qt/python/mantidqt/utils/qt/__init__.py#L82 which can be used as:

with block_signals(w):
   ... wrapped code

This way you can guarantee that any exception in the middle won't cause the final statement to be ignored. It's Python's way of implementing the RAII style in C++.

PR #275, opened 2018-03-14
Dataloader usability fixes

### LINKED ISSUE #273: Dataloader usability umbrella issues
There are a few small fixes which could improve the usability of the `DataLoader`:

1. Show only files which MSlice can load (e.g. `*.nxs`, `*.nxspe`, `*.txt`).
2. Allow a file to be loaded if it is double-clicked or when `Enter` / `Return` on the keyboard is pressed in addition to the `Load` button being clicked.
3. Pressing `Backspace` should have the same effect as clicking `Back`.

mducle:
Adds various small features to the `DataLoader` widget:

* Only files loadable by MSlice (`nxs`, `nxspe`, `txt`, `xye`) are shown.
* Double clicking a file now loads it. (Double clicking a folder opens it as before).
* Pressing `Enter` on a file loads it, on a folder opens it.
* Pressing `Backspace` moves up a folder.

**To test:**

Code review.

Run MSlice and check the above functionality works as expected.

Fixes #273.

### DIFF
--- a/mslice/widgets/dataloader/dataloader.py
+++ b/mslice/widgets/dataloader/dataloader.py
@@ -5,7 +5,7 @@ import os
 from functools import partial
 
 from mslice.util.qt.QtWidgets import QWidget, QFileSystemModel, QAbstractItemView, QMessageBox
-from mslice.util.qt.QtCore import Signal, QDir
+from mslice.util.qt.QtCore import Signal, QDir, Qt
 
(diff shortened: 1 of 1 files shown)


MatthewBowles:
I think this is because double clicking also triggers the `activated` signal. Removing the `table_view.doubleClicked.connect` line should fix the problem.

PR #292, opened 2018-03-21
Move check_mpl call

### LINKED ISSUE #291: Check_mpl not called on launch
`check_mpl()` can be skipped when launched from `mantidplot`, which causes an error where `MPL_COMPAT` is mistakenly set to false.

Observed on Ubuntu 14.04

MatthewBowles:
Moves `check_mpl()` inside `show_gui` so it is always called to prevent errors on old matplotlib versions.

Fixes #291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ branches:
 matrix:
     include:
         - python: "2.7_with_system_site_packages"
-          env: PY=2.7 PYTHONPATH=$PYTHONPATH:/opt/mantidnightly/bin
+          env: PY=2.7 PYTHONPATH=$PYTHONPATH:/opt/mantidnightly/bin QT_API=pyqt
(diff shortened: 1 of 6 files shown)


MatthewBowles:
@martyngigg I'm struggling to figure out why python 2.7 is segfaulting here...

martyngigg:
@MatthewBowles It's most likely the fact that there is no X server running as it does run locally for me.

Looking at the code a little close it's probably best if we move the check all together so that we don't have to import the `app` module in `models`. How about we just move it to `util.__init__.py` and just set the flag there when the module is imported?

martyngigg:
There seems to be an issue with Python 2 and PyQt 5.10 on travis: travis-ci/travis-ci#9202.

I've forced PyQt4 for 2.7 tests.

PR #296, opened 2018-03-22
Move common axis labelling code

### LINKED ISSUE #293: Refactor axis labelling code
There is duplicate code in both [`matplotlib_slice_plotter`](https://github.com/mantidproject/mslice/blob/master/mslice/models/slice/matplotlib_slice_plotter.py#L53) and [`matplotlib_cut_plotter`](https://github.com/mantidproject/mslice/blob/master/mslice/models/cut/matplotlib_cut_plotter.py#L78) regarding axes labels and dealing with latex support.

We should refactor the axes label code into a separate class, something like LabelMaker, that can be composed with the plotter classes to consolidate the code. Consider having methods for generating axis labels and legend labels.

MatthewBowles:
Moves axis labels and latex code into new `labels` module.

**To test:**

- check axis labels are still displayed correctly (on slices and cuts)

Fixes #293

### DIFF
--- a/mslice/models/labels.py
+++ b/mslice/models/labels.py
@@ -30,4 +30,3 @@ def generate_legend(workspace_name, integrated_dim, integration_start, integrati
     integrated_dim = mappings[integrated_dim] if integrated_dim in mappings else integrated_dim
     return workspace_name + " " + "%.2f" % integration_start + "<" + integrated_dim + "<" + \
         "%.2f" % integration_end
-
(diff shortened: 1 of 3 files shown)


mslice/models/labels.py
```diff
@@ -0,0 +1,32 @@
+'''Module to create axis and legend labels'''
+
+from __future__ import (absolute_import, division, print_function)
+from mslice.util import MPL_COMPAT
+
+INTENSITY_LABEL = 'Signal/#Events'
```
martyngigg: If we're moving this out of the cut model then I'd suggest renaming this `CUT_INTENSITY_LABEL`

PR #298, opened 2018-03-27
Use context manager for busy/idle status

### LINKED ISSUE #272: Use context manager for busy/idle status
`mainwindow.py` has a `show_busy` method to show when mslice is busy. 
In the presenters this method is called twice (often in `notify`) with `show_busy(true)` before some operation and `show_busy(false)` afterwards. 
This would be better handled by a context manager (to ensure the `show_busy(false)` call always happens regardless of exceptions etc.)

MatthewBowles:
This means that it will always exit correctly regardless of exceptions etc.

**To test:**
Check that busy/idle in the bottom corner still toggles appropriately.

Fixes #272

### DIFF
--- a/mslice/app/mainwindow.py
+++ b/mslice/app/mainwindow.py
@@ -169,15 +169,3 @@ class MainWindow(MainView, QMainWindow):
         else:
             return
         QApplication.processEvents()
-
-
-class ShowBusy():
-
-    def __init__(self, view):
-        self.view = view
-
-    def __enter__(self):
-        self.view.busy.emit(True)
-
-    def __exit__(self, *args):
-        self.view.busy.emit(False)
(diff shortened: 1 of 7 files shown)


mslice/app/mainwindow.py
```diff
@@ -169,3 +169,15 @@ def show_busy(self, busy):
         else:
             return
         QApplication.processEvents()
+
+
+class ShowBusy():
```
martyngigg: I don't think this is the best place for this as it requires the presenter modules to have links back to the mainwindow module.

How about we just have a separate `busy.py` in presenters? We can also do this using [`contextlib`](https://docs.python.org/2/library/contextlib.html) without the need to define a class:

```python
from contextlib import contextmanager

@contextmanager
def show_busy(view):
    self.view.busy.emit(True)
    yield
    self.view.busy.emit(False)
```

then

```python
with show_busy(...):
    ...
```

MatthewBowles: Done, that's way better.

PR #302, opened 2018-04-20
Use workspace wrappers for caching parameters

### LINKED ISSUE #178: Refactor to move all workspace parameter caching into new `Workspace` classes
The `WorkspaceProvider` class in the GUI current caches various parameters related to the workspaces, such as whether Efixed has been defined (either read in from the workspace or user inputted for indirect geometry `nxspe` file types), the precomputed default axes limits and the cut limits and integration ranges if they have been previously defined by the user.

The current cache is stored as `dict` attributes of `MantidWorkspaceProvider`, so all presenters must look at the same instance of  `MantidWorkspaceProvider` - there is code in the `__init__` methods of each presenter to cascade down the same instance of `MantidWorkspaceProvider`. 

It would be much cleaner for these parameters to be stored in the workspace objects themselves. Since we will be using Python objects to wrap the underlying Mantid workspaces types for the CLI, it thus makes more sense for the cached parameters to be assigned to these workspace objects instead.

So, when the CLI implementation is mature enough, we should move the current cache code in `MantidWorkspaceProvider` over to the the workspace objects.

MatthewBowles:
The `mantid_workspace_provider` class is used to store cached information about workspaces. This causes a problem with having to pass around references to this object everywhere, which would become even more difficult when the CLI is expanded.

It makes more sense to keep any state and cached values as part of the workspace itself. This PR stores this information in the workspace wrapper classes, and makes sure all workspaces are wrapped on load. 
This means `mantid_workspace_provider` no longer has state and is a module of free functions instead of a class.

This will make upcoming CLI improvements much easier to implement.

Fixes #178

### DIFF
--- a/mslice/models/workspacemanager/mantid_workspace_provider.py
+++ b/mslice/models/workspacemanager/mantid_workspace_provider.py
@@ -8,12 +8,16 @@ It uses mantid to perform the workspace operations
 from __future__ import (absolute_import, division, print_function)
 import os.path
 from six import string_types
+from mantid.api import IMDEventWorkspace, IMDHistoWorkspace
 
 from mantid.simpleapi import (AnalysisDataService, DeleteWorkspace, Load, Scale,
                               RenameWorkspace, MergeMD, MergeRuns, Minus)
 
 from mslice.presenters.slice_plotter_presenter import Axis
-from mantid.api import IMDEventWorkspace, MatrixWorkspace, Workspace
+from mslice.workspace.base import WorkspaceBase as Workspace
+from mslice.workspace.workspace import Workspace as MatrixWorkspace
+from mslice.workspace.pixel_workspace import PixelWorkspace
+from mslice.workspace.histogram_workspace import HistogramWorkspace
 from .file_io import save_ascii, save_matlab, save_nexus
 import numpy as np
 from scipy import constants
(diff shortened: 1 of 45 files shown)


mslice/tests/workspace_provider_test.py
```diff
@@ -1,90 +1,93 @@
...
 class MantidWorkspaceProviderTest(unittest.TestCase):
 
     def setUp(self):
-        self.ws_provider = MantidWorkspaceProvider()
-
-        self.test_ws_2d = CreateSimulationWorkspace(Instrument='MAR', BinParams=[-10, 1, 10],
-                                                    UnitX='DeltaE', OutputWorkspace='test_ws_2d')
-        AddSampleLog(self.test_ws_2d, LogName='Ei', LogText='3.', LogType='Number')
-        self.test_ws_md = ConvertToMD(InputWorkspace=self.test_ws_2d, OutputWorkspace="test_ws_md", QDimensions='|Q|',
+        self.test_ws_2d = wrap_workspace(CreateSimulationWorkspace(Instrument='MAR', BinParams=[-10, 1, 10],
```
martyngigg: I guess it could get a little repetitive to keep having to call `wrap_workspace` every time we call an algorithm. Can we centralise this somehow? Maybe a single function to execute a mantid algorithm by string name and this could also centralise the `StoreInTheADS` call?

martyngigg:
I was also trying out the intensity calculations and I got an error with `Chi''`:
```
Load-[Notice] Load started
Load-[Notice] Load successful, Duration 0.15 seconds
SofQWNormalisedPolygon-[Notice] SofQWNormalisedPolygon started
SofQWNormalisedPolygon-[Notice] SofQWNormalisedPolygon successful, Duration 0.10 seconds
  File "C:\Users\gigm\Code\mantidproject\mslice\mslice\plotting\plot_window\slice_plot.py", line 221, in show_intensity_plot
    if not self._run_temp_dependent(slice_plotter_method, previous):
  File "C:\Users\gigm\Code\mantidproject\mslice\mslice\plotting\plot_window\slice_plot.py", line 246, in _run_temp_dependent
    self._slice_plotter.update_sample_temperature(self.ws_name)
  File "C:\Users\gigm\Code\mantidproject\mslice\mslice\models\slice\matplotlib_slice_plotter.py", line 150, in update_sample_temperature
    temp = self._slice_algorithm.sample_temperature(workspace, self._sample_temp_fields)
  File "C:\Users\gigm\Code\mantidproject\mslice\mslice\models\slice\mantid_slice_algorithm.py", line 202, in sample_temperature
    sample_temp = ws.getExperimentInfo(0).run().getLogData(field_name).value
AttributeError: 'Workspace' object has no attribute 'getExperimentInfo'```

To reproduce this:
* load `MAR21335_Ei60.00meV`
* display the default slice
* select Intensity -> `Chi''(Q,E)` -> select the `t_head` entry

File: mslice/models/workspacemanager/mantid_workspace_provider.py [hunk 2]
```diff
@@ -29,375 +28,393 @@
...
+def get_workspace_handle(workspace_name):
+    """"Return handle to workspace given workspace_name_as_string"""
+    # if passed a workspace handle return the handle
+    if isinstance(workspace_name, Workspace):
+        return workspace_name
+    return _loaded_workspaces[workspace_name]
+
+
+def run_alg(alg_name, output_name=None, store=True, **kwargs):
+    if 'InputWorkspace' in kwargs.keys() and isinstance(kwargs['InputWorkspace'], Workspace):
```
martyngigg: How about 
```python
if isinstance(kwargs.get('Inputworkspace', None), Workspace):
```
?

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -29,375 +28,393 @@
...
+
+    ws = getattr(mantid_algs, alg_name)(**kwargs)
+
+    if store:
+        ws = wrap_workspace(ws, output_name)
+    return ws
+
+
+def get_workspace_names():
+    return [key for key in _loaded_workspaces.keys() if key[:2] != '__']
```
martyngigg: I would suggest using [`six.iterkeys`](https://pythonhosted.org/six/#six.iterkeys) to avoid making a copy of the `keys()` in Python 2

mslice/models/workspacemanager/mantid_workspace_provider.py
```diff
@@ -29,375 +28,393 @@
...
+
+
+def delete_workspace(workspace):
+    workspace = get_workspace_handle(workspace)
+    del _loaded_workspaces[get_workspace_name(workspace)]
+    del workspace
+
+
+def workspace_exists(ws_name):
+    return True if ws_name in _loaded_workspaces.keys() else False
```
martyngigg: I think this can be shrunk to 
```python
return ws_name in _loaded_workspaces
```

PR #304, opened 2018-04-25
Add load command

### LINKED ISSUE #303: Load using CLI
Implement a `Load` command for the CLI:

`ws = load('path_to_file')`

MatthewBowles:
Adds load command to CLI.

```
import mslice.cli as mc
mc.load('path_to_file')
```

Fixes #303

### DIFF
--- a/mslice/cli/_mslice_commmands.py
+++ b/mslice/cli/_mslice_commmands.py
@@ -9,8 +9,9 @@ from __future__ import (absolute_import, division, print_function)
 from mantid.api import IMDWorkspace as _IMDWorkspace
 from mantid.api import Workspace as _Workspace
 from mantid.kernel.funcinspect import lhs_info as _lhs_info
-from mantid.simpleapi import mtd, Load, ConvertUnits, RenameWorkspace # noqa: F401
+from mantid.simpleapi import mtd, ConvertUnits, RenameWorkspace # noqa: F401
 
+from mslice.app import MAIN_WINDOW
 # Helper tools
 from mslice.models.workspacemanager.mantid_workspace_provider import MantidWorkspaceProvider as _MantidWorkspaceProvider
 from mslice.presenters.slice_plotter_presenter import Axis as _Axis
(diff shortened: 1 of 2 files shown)


MatthewBowles:
Now that I think about it probably need some validation / error handling on this. Not sure how thorough it should be though.

Also, I think it makes sense to do the CLI stuff on a different branch so it's only put into master when it all works... If so, should probably try and get the CI working here.

martyngigg:
We're also going to have to consider putting the IPython terminal back I presume. @mducle I'm assuming that's how we're going to do interact with this? Also if we're running inside MantidPlot should we then use the `MantidPlot` one and not use a separate one?

mducle:
@martyngigg Yes, I think we should put the IPython terminal back. I think if we can do this without conflicting with matplotlib in Mantidplot then it might give a more uniform ui to keep the IPython terminal even in MantidPlot. Otherwise, we can use the scripting engine in MantidPlot.

mducle:
@MatthewBowles I agree to have a separate CLI branch and then to merge it to master when it's ready - like with the new GUI layout.

martyngigg:
I can randomly get a segfault when using `cli.load`:
```
Load-[Notice] Load started
Load-[Notice] Load successful, Duration 0.14 seconds
QObject::connect: Cannot queue arguments of type 'QItemSelection'
(Make sure 'QItemSelection' is registered using qRegisterMetaType().)
QObject::connect: Cannot queue arguments of type 'QItemSelection'
(Make sure 'QItemSelection' is registered using qRegisterMetaType().)
QObject::setParent: Cannot set parent, new parent is in a different thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
X Error: BadImplementation (server does not implement operation) 17
  Major opcode: 20 (X_GetProperty)
  Resource id:  0x0
```

The exact form of the error is linux specific but indicates that a UI draw command has been attempted on a non-GUI thread. This is most likely due to our [custom run code](https://github.com/mantidproject/mslice/blob/master/mslice/widgets/ipythonconsole/ipython_widget.py#L76) that runs the operation on a separate thread. I propose that we remove the use of a separate thread for now as we're not using it anywhere else and we tackle that separately.

MatthewBowles:
@martyngigg Did this fix the problem?

martyngigg:
@MatthewBowles That looks good now.

PR #305, opened 2018-04-26
Run algorithms with StoreInADS=False

### LINKED ISSUE #183: Refactor Mantid algorithms to use StoreInADS
This should not be implemented until MSlice properly specifies a minimum required version of Mantid (and probably not until Mantid ver. 3.12)

Currently, many Mantid algorithms used in MSlice are either run normally and then the output is deleted from the ADS or they are run as a child, e.g. the `run_child_alg` method in `workspace_mixin`. These should be refactored to make use of the StoreInADS keyword.

MatthewBowles:
**Depends on mantidproject/mantid#22348**

Uses StoreInADS keyword to eliminate unnecessary overhead when running algorithms.

Fixes #183


martyngigg:
@MatthewBowles Could you rebase this with `master` so we can get a cleaner diff?

MatthewBowles:
@martyngigg Is it supposed to do that?

martyngigg:
@MatthewBowles No - I think the default config only let's admin's build without asking - I'll take a look

PR #307, opened 2018-04-30
MakeProjection as Algorithm

MatthewBowles:
**Do not merge until after #302**

Lays groundwork for converting some MSlice functions to algorithms and creates the `MakeProjection` algorithm.


mslice/app/__init__.py
```diff
@@ -20,6 +23,8 @@ def show_gui():
     If this is the first call then an instance of the Windows is cached to ensure it
     survives for the duration of the application
     """
+    AlgorithmFactory.subscribe(MakeProjection)
+    _translate()
```
MatthewBowles: @martyngigg I had to add the `_translate()` call here for `simpleapi` to pick up the algorithm... is there a better way of doing this?

martyngigg: @MatthewBowles You're right - this is a little messy. The first thing I would do is have a separate module and function to handle all of the registration. Maybe a `mslice.util.mantid` with an `initialize_mantid` or something? We could also then move the `run_alg` function into there.

For the `_translate` step, how about calling `_create_algorithm_function` directly? That way we avoid having to run through all of the algorithms again?

MatthewBowles: This is better. Apologies for the cluttered diff, I had to split `workspace_provider` to avoid dependency problems.

mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -0,0 +1,347 @@
+"""A concrete implementation of a WorkspaceProvider
```
martyngigg: I think this comment might be out of date?

mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -0,0 +1,347 @@
...
+        # If we cannot get the step size from the data, use the old 1/100 steps.
+        ws_h = workspace.raw_ws
+        dim = ws_h.getDimension(ws_h.getDimensionIndexByName(axis))
+        minimum = dim.getMinimum()
+        maximum = dim.getMaximum()
+        step = (maximum - minimum) / 100
+        return minimum, maximum, step
+
+
+def _processEfixed(workspace):
```
martyngigg: Should these have `_` prefixes if they are used in other modules?

mslice/util/mantid.py
```diff
@@ -0,0 +1,30 @@
...
+
+def initialize_mantid():
+    AlgorithmFactory.subscribe(MakeProjection)
+    s_api._create_algorithm_function('MakeProjection', 1, MakeProjection())
+
+
+def run_alg(alg_name, output_name=None, store=True, **kwargs):
+    return run_algorithm(getattr(s_api, alg_name), output_name, store, **kwargs)
+
+def run_algorithm(algorithm, output_name=None, store=True, **kwargs):
```
martyngigg: I think the `algorithm` parameter might be better named as `algorithm_function`.

mslice/util/mantid.py
```diff
@@ -0,0 +1,30 @@
...
+from mslice.workspace import wrap_workspace
+from mslice.workspace.base import WorkspaceBase as Workspace
+
+
+def initialize_mantid():
+    AlgorithmFactory.subscribe(MakeProjection)
+    s_api._create_algorithm_function('MakeProjection', 1, MakeProjection())
+
+
+def run_alg(alg_name, output_name=None, store=True, **kwargs):
```
martyngigg: I would say it might be confusing in the code to have a both a `run_alg` and `run_algorithm` function - it's not immediately obvious what the difference is.

How about a single `run_algorithm` that can accept either a string or algorithm function?

PR #310, opened 2018-05-01
Slice as Algorithm

MatthewBowles:
Converts `compute_slice` operations to an algorithm.


mslice/models/axis.py
```diff
@@ -5,6 +5,9 @@ def __init__(self, units, start, end, step):
         self.end = end
         self.step = step
 
+    def to_dict(self):
```
martyngigg: Nice. :+1:

MatthewBowles:
Not as such, it's probably worth adding some tests here.

MatthewBowles:
I've added a test that checks the format of the output... It seems that the reason there weren't many tests here before is because it's hard to check without using real data. No matter what kind of test workspace I try to pass through `compute_slice` the `plot_data` always comes back as zeroes which isn't really very useful for testing.

mslice/models/slice/mantid_slice_algorithm.py
```diff
@@ -32,6 +32,7 @@ def compute_slice(self, selected_workspace, x_axis, y_axis, norm_to_one):
         slice =  run_algorithm('Slice', output_name = '__' + workspace.name, InputWorkspace=workspace,
                                XAxis=x_axis.to_dict(), YAxis=y_axis.to_dict(), PSD=workspace.is_PSD,
                                EMode=workspace.e_mode, NormToOne=norm_to_one)
+        print(slice.get_signal())
```
martyngigg: Extra debugging statement left by accident?

PR #313, opened 2018-05-03
Cut as Algorithm

### LINKED ISSUE #306: Convert some operations to Mantid algorithms
To make certain MSlice operations easier to use from the CLI and make use of the workspace history, they should be converted to Mantid algorithms. These should only be registered with `AlgorithmFactory` when MSlice is loaded. 

Algorithms:
- MakeProjection
- Slice
- Cut

MatthewBowles:
Converts cut operations to an algorithm.

Fixes #306


martyngigg:
#319 discovered during testing but it is already on `master` and so unrelated to these changes.

PR #324, opened 2018-05-17
Improve algorithm wrapper

MatthewBowles:
Improves algorithm wrapper and automatically applies it to all Mantid algorithms. This will allow most Mantid algorithms to be used from the `ipython` console seamlessly.
Also imports `mslice.cli` and `mantid` algorithms by default.

### DIFF
--- a/mslice/app/__init__.py
+++ b/mslice/app/__init__.py
@@ -3,7 +3,7 @@ and entry points.
 """
 
 from mslice.util.qt.QtWidgets import QApplication
-from mslice.util.mantid import initialize_mantid
+from mslice.util.mantid.init_mantid import initialize_mantid
 
 # Module-level reference to keep main window alive after show_gui has returned
 MAIN_WINDOW = None
@@ -25,4 +25,5 @@ def show_gui():
     if MAIN_WINDOW is None:
         from mslice.app.mainwindow import MainWindow
         MAIN_WINDOW = MainWindow()
+        MAIN_WINDOW.setup_ipython()
     MAIN_WINDOW.show()
--- a/mslice/cli/_mslice_commmands.py
+++ b/mslice/cli/_mslice_commmands.py
@@ -12,6 +12,7 @@ from mantid.api import IMDWorkspace as _IMDWorkspace
 
 from mslice.app import MAIN_WINDOW
 from mslice.workspace.base import WorkspaceBase as Workspace
+from mslice.util.mantid import run_algorithm, Scale
 # Helper tools
 from mslice.models.workspacemanager.workspace_provider import get_workspace_handle, workspace_exists
 from mslice.presenters.slice_plotter_presenter import Axis as _Axis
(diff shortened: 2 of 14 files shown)


MatthewBowles:
@martyngigg Seem to be getting a weird test failure on Linux. Seems like `mantid_algorithms` might be being imported before `initialize_mantid` is run which should not be the case. Passes fine on Windows.

mslice/app/__init__.py
```diff
@@ -25,4 +25,5 @@ def show_gui():
     if MAIN_WINDOW is None:
         from mslice.app.mainwindow import MainWindow
         MAIN_WINDOW = MainWindow()
+        MAIN_WINDOW.setup_ipython()
```
martyngigg: Can you help me understand why we don't do this as part of the constructor anymore?

mslice/cli/_mslice_commmands.py
```diff
@@ -8,39 +8,24 @@
...
-_CUT_PLOTTER = MatplotlibCutPlotter(_CUT_ALGORITHM)
 
 # -----------------------------------------------------------------------------
 # Convenience functions
 # -----------------------------------------------------------------------------
 
 def _process_axis(axis, fallback_index, input_workspace):
-    if axis is None:
-        axis = _SLICE_ALGORITHM.get_available_axis(input_workspace)[fallback_index]
+    # if axis is None:
```
martyngigg: Can we remove the dead code?

mslice/app/__init__.py
MatthewBowles: I can't remember. I changed it back and everything seems fine so I guess I don't understand either.

martyngigg:
@MatthewBowles There are a few merge conflicts here now after a few other PRs were merged.

mslice/app/__init__.py
MatthewBowles: I remember why I did this now  - if `setup_ipython` is done as part of the `MainWindow` constructor, the CLI commands can't find the `MainWindow` instance e.g.

```
AttributeError                            Traceback (most recent call last)
<ipython-input-3-8acdd31d31dc> in <module>()
----> 1 ws = Load('C:\Users\ygh95688\Documents\MSlice_Test_Data\MAR21335_ei60.00meV.nxs')

C:\Users\ygh95688\PycharmProjects\mslice_2\mslice\cli\_mslice_commmands.pyc in Load(path)
     84     if not ospath.exists(path):
     85         raise RuntimeError('could not find the path %s' % path)
---> 86     MAIN_WINDOW.dataloader_presenter.load_workspace([path])
     87     return get_workspace_handle(ospath.splitext(ospath.basename(path))[0])
     88 

AttributeError: 'NoneType' object has no attribute 'dataloader_presenter'
```

Is there another way around this other than moving the `setup_ipython` call?

martyngigg: It's not clear to me why `MAIN_WINDOW` has not been set. The only way I can see that `MAIN_WINDOW` is `None` is if the `MainWindow` constructor raises an exception. Is that the case?

MatthewBowles: It's not clear to me either. I stepped through the constructor in the debugger and everything seems to run fine.

MatthewBowles: This was a problem with the way `MAIN_WINDOW` was imported meaning it wasn't updated, should be ok now.

martyngigg: Ah yes, I understand now. :+1:

MatthewBowles:
That is supposed to work, I'll take a look...

MatthewBowles:
@martyngigg Could you check this again please?

martyngigg:
:+1:

PR #327, opened 2018-05-22
Update pyplot to be compatible with matplotlib 2.2.3

martyngigg:
**Description of work**

While testing mantidproject/mantid#22322 it was observed that matplotlib 2.2.3 breaks mslice. This fixes the problems caused. The major change is that `mslice.plotting.pyplot` is now generated by a script much like `matplotlib.pyplot`. There has also been some minor refactoring to clean up code around the current figure management and decorators.

Note that the current api is still backwards compatible with older versions of `matplotlib`.

**To test:**

* Ideally all aspects of the plotting should be verified.

*No associated issue*


martyngigg:
Thanks for spotting that, Matt. I'd tested it with slice then cut which seemed fine!
I think this could have implications for the cli and how users could find it confusing in some situations when functions like `pyplot.cla` are used. We'll have to see how this goes as the cli progresses I think.

martyngigg:
I've just rebased this after the changes from #331.

mslice/models/cut/matplotlib_cut_plotter.py
```diff
@@ -19,27 +21,37 @@ def plot_cut(self, selected_workspace, cut_axis, integration_axis, norm_to_one,
...
             self._create_cut(cut_ws_name if cut_ws_name is not None else selected_workspace)
-        plt.gcf().canvas.restore_region(plt.gcf().canvas.manager.get_cut_background())
+            cur_canvas.restore_region(cur_canvas.manager.get_cut_background())
         try:
-            plt.gca().draw_artist(plt.gcf().canvas.figure.get_children()[1])
-            plt.gcf().canvas.blit(plt.gcf().canvas.figure.gca().clipbox)
+            children = cur_fig.get_children()
+            for artist in children:
+                cur_axes.draw_artist(artist)
+                cur_canvas.blit(cur_axes.clipbox)
```
MatthewBowles: You can move the `blit` call outside the loop, it only needs to be done once.

martyngigg: Good catch. I hate Python's indentation sometimes...

Issue #322, opened 2018-05-15
Tidy up slice presenter


MatthewBowles:
The `display_slice` method in `slice_plotter_presenter` is a little messy, especially after the validation of the axis parameters was moved to the `Axis` class in #320.

PR #336, opened 2018-07-11
Improve default Q limits for large data sets

### LINKED ISSUE #335: Wrong limits for HYSPEC NXSPE
I've generated an NXSPE file from systemtest data:

```
Load(Filename='HYS_13657_event.nxs', OutputWorkspace='HYS_13657_event')
DgsReduction(SampleInputWorkspace='HYS_13657_event', OutputWorkspace='HYS_ws2D')
SaveNXSPE(InputWorkspace='HYS_ws2D',Filename='test_mslice_HYS.nxspe',Efixed=50,Psi=0)
```

I've loaded it in MSlice. I clicked calculate projection. The energy range is OK, but the `|Q|` step is so small that when I click the display button it's complaining that the step is too small. I've changed the step to 0.05. The minimum limit is wrong, it cuts into the data.  I get the default to be 4.4, when it should be around 3.3 instead.

I've put in the step by hand to be 0.05, but I've deleted the minimum value. When I click display, I get the following error message:

```
  File "/home/andrei/Mantid/Build/scripts/ExternalInterfaces/mslice/src/mslice/mslice/models/axis.py", line 27, in start
    self._start = float(value)
ValueError: could not convert string to float: 

During handling of the above exception, another exception occurred:

  File "/home/andrei/Mantid/Build/scripts/ExternalInterfaces/mslice/src/mslice/mslice/widgets/slice/slice.py", line 49, in _btn_clicked
    self._presenter.notify(Command.DisplaySlice)
  File "/home/andrei/Mantid/Build/scripts/ExternalInterfaces/mslice/src/mslice/mslice/presenters/slice_plotter_presenter.py", line 40, in notify
    self._display_slice()
  File "/home/andrei/Mantid/Build/scripts/ExternalInterfaces/mslice/src/mslice/mslice/presenters/slice_plotter_presenter.py", line 52, in _display_slice
    self._slice_view.get_slice_x_end(), self._slice_view.get_slice_x_step())
  File "/home/andrei/Mantid/Build/scripts/ExternalInterfaces/mslice/src/mslice/mslice/models/axis.py", line 4, in __init__
    self.start = start
  File "/home/andrei/Mantid/Build/scripts/ExternalInterfaces/mslice/src/mslice/mslice/models/axis.py", line 29, in start
    raise ValueError("Invalid axis parameter - start '%s' is not a valid float" % value)
ValueError: Invalid axis parameter - start '' is not a valid float

```

MatthewBowles:
Previously, if a workspace had more than a thousand histograms, reading the `theta` values would be cut short to save time. This resulted in incorrect default Q limits. This PR removes this 'shortcut'.

Also tidied up some of the validation when entering slice values, so that the user gets meaningful error messages if a value is invalid. This is just a quick fix for now as there is already an issue to refactor this particular part of the code (#322).

**To test:**
See issue

Fixes #335

### DIFF
--- a/mslice/models/axis.py
+++ b/mslice/models/axis.py
@@ -34,12 +34,13 @@ class Axis(object):
 
     @end.setter
     def end(self, value):
-        if value < self.start:
-            raise ValueError("Invalid axis parameter - end is smaller than start")
         try:
-            self._end = float(value)
+            end = float(value)
         except ValueError:
             raise ValueError("Invalid axis parameter - end '%s' is not a valid float" % value)
+        if end < self.start:
+            raise ValueError("Invalid axis parameter - end is smaller than start")
+        self._end = end
 
     @property
     def step(self):
--- a/mslice/models/axis.py
+++ b/mslice/models/axis.py
@@ -34,6 +34,8 @@ class Axis(object):
 
     @end.setter
     def end(self, value):
+        if value < self.start:
+            raise ValueError("Invalid axis parameter - end is smaller than start")
         try:
             self._end = float(value)
         except ValueError:
--- a/mslice/models/axis.py
+++ b/mslice/models/axis.py
@@ -35,12 +35,12 @@ class Axis(object):
     @end.setter
     def end(self, value):
         try:
-            end = float(value)
+            end_float = float(value)
         except ValueError:
             raise ValueError("Invalid axis parameter - end '%s' is not a valid float" % value)
-        if end < self.start:
+        if end_float < self.start:
             raise ValueError("Invalid axis parameter - end is smaller than start")
-        self._end = end
+        self._end = end_float
 
     @property
     def step(self):
--- a/mslice/models/workspacemanager/workspace_algorithms.py
+++ b/mslice/models/workspacemanager/workspace_algorithms.py
@@ -107,6 +107,7 @@ def _get_property_from_history(name, history):
 
 
 def get_q_limits(theta, emax, efix):
+    # calculate |Q| limits in Angstrom^-1 from theta (scattering angle) in radians
     qmin, qmax, qstep = tuple(np.sqrt(E2q * 2 * efix * (1 - np.cos(theta)) * meV2J) / m2A)
     qmax = np.sqrt(E2q * (2 * efix + emax - 2 * np.sqrt(efix * (efix + emax)) * np.cos(theta[1])) * meV2J) / m2A
     return qmin, qmax, qstep
@@ -123,18 +124,8 @@ def set_limits(ws, qmin, qmax, qstep, theta, emin, emax, estep):
 def _get_theta_for_limits(ws):
     # Don't parse all spectra in cases where there are a lot to save time.
     num_hist = ws.raw_ws.getNumberHistograms()
-    if num_hist > 1000:
-        n_segments = 5
-        interval = int(num_hist / n_segments)
-        theta = []
-        for segment in range(n_segments):
-            i0 = segment * interval
-            theta.append([ws.raw_ws.detectorTwoTheta(ws.raw_ws.getDetector(i))
-                          for i in range(i0, i0+200)])
-        round_fac = 573
-    else:
-        theta = [ws.raw_ws.detectorTwoTheta(ws.raw_ws.getDetector(i)) for i in range(num_hist)]
-        round_fac = 100
+    theta = [ws.raw_ws.detectorTwoTheta(ws.raw_ws.getDetector(i)) for i in range(num_hist)]
+    round_fac = 100
     ws.is_PSD = not all(x < y for x, y in zip(theta, theta[1:]))
     # Rounds the differences to avoid pixels with same 2theta. Implies min limit of ~0.5 degrees
     thdiff = np.diff(np.round(np.sort(theta)*round_fac)/round_fac)
--- a/mslice/models/workspacemanager/workspace_algorithms.py
+++ b/mslice/models/workspacemanager/workspace_algorithms.py
@@ -79,7 +79,7 @@ def process_limits_event(ws):
     theta = _get_theta_for_limits_event(ws)
     estep = _original_step_size(ws)
     emax_1 = -emin if (str(ws.e_mode == 'Direct')) else emax
-    qmin, qmax, qstep = get_q_limits(theta, emax_1, ws.e_fixed)
+    qmin, qmax, qstep = get_q_limits(theta, emax_1, ws.e_fixed, ws.e_mode)
     set_limits(ws, qmin, qmax, qstep, theta, emin, emax, estep)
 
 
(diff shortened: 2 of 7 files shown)


mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -107,6 +107,7 @@ def _get_property_from_history(name, history):
 
 
 def get_q_limits(theta, emax, efix):
+    # calculate |Q| limits in Angstrom^-1 from theta (scattering angle) in radians
     qmin, qmax, qstep = tuple(np.sqrt(E2q * 2 * efix * (1 - np.cos(theta)) * meV2J) / m2A)
     qmax = np.sqrt(E2q * (2 * efix + emax - 2 * np.sqrt(efix * (efix + emax)) * np.cos(theta[1])) * meV2J) / m2A
     return qmin, qmax, qstep
```
mducle: @MatthewBowles Thanks for looking into this. The Q-limits bug can be fixed by changing the code here:

```
--- a/mslice/models/workspacemanager/workspace_algorithms.py
+++ b/mslice/models/workspacemanager/workspace_algorithms.py
@@ -68,9 +68,8 @@ def _processLoadedWSLimits(workspace):
 def process_limits(ws):
     en = ws.raw_ws.getAxis(0).extractValues()
     theta = _get_theta_for_limits(ws)
-    # Use minimum energy (Direct geometry) or maximum energy (Indirect) to get qmax
-    emax = -np.min(en) if (str(ws.e_mode == 'Direct')) else np.max(en)
-    qmin, qmax, qstep = get_q_limits(theta, emax, ws.e_fixed)
+    # If Direct geometry use +e_fixed, if indirect geometry, use -e_fixed
+    qmin, qmax, qstep = get_q_limits(theta, en, ws.e_fixed if (str(ws.e_mode == 'Direct')) else -ws.e_fixed)
     set_limits(ws, qmin, qmax, qstep, theta, np.min(en), np.max(en), np.mean(np.diff(en)))
 
 
@@ -106,9 +105,12 @@ def _get_property_from_history(name, history):
     return None
 
 
-def get_q_limits(theta, emax, efix):
-    qmin, qmax, qstep = tuple(np.sqrt(E2q * 2 * efix * (1 - np.cos(theta)) * meV2J) / m2A)
-    qmax = np.sqrt(E2q * (2 * efix + emax - 2 * np.sqrt(efix * (efix + emax)) * np.cos(theta[1])) * meV2J) / m2A
+def get_q_limits(theta, en, efix):
+    Ei = np.abs(efix)
+    qlines = [np.sqrt(E2q * (2 * Ei - en - 2 * np.sqrt(Ei * (Ei - en)) * np.cos(tth)) * meV2J) / 1e10 for tth in theta[:2]]
+    qmin = np.min(qlines[0 if (efix > 0) else 1])
+    qmax = np.max(qlines[1 if (efix > 0) else 0])
+    qstep = np.sqrt(E2q * 2 * efix * (1 - np.cos(theta[2])) * meV2J) / m2A
     return qmin, qmax, qstep
```

MatthewBowles: Thanks!

mslice/models/axis.py
```diff
@@ -35,9 +35,12 @@ def end(self):
     @end.setter
     def end(self, value):
         try:
-            self._end = float(value)
+            end = float(value)
```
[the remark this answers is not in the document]
MatthewBowles: I think it's okay because they're in different scopes but I'll change it to be safe.

mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -106,9 +104,12 @@ def _get_property_from_history(name, history):
...
 
-def get_q_limits(theta, emax, efix):
-    qmin, qmax, qstep = tuple(np.sqrt(E2q * 2 * efix * (1 - np.cos(theta)) * meV2J) / m2A)
-    qmax = np.sqrt(E2q * (2 * efix + emax - 2 * np.sqrt(efix * (efix + emax)) * np.cos(theta[1])) * meV2J) / m2A
+def get_q_limits(theta, en, efix):
+    #calculates the Q(E) line for the given two theta and then finds the min and max values
+    qlines = [np.sqrt(E2q * (2 * efix - en - 2 * np.sqrt(efix * (efix - en)) * np.cos(tth)) * meV2J) / 1e10 for tth in theta[:2]]
+    qmin = np.nanmin(qlines[0])
+    qmax = np.nanmax(qlines[1])
+    qstep = np.sqrt(E2q * 2 * efix * (1 - np.cos(theta[2])) * meV2J) / m2A
```
AndreiSavici: The default q step is too small if you have many detectors. I would choose something like (qmax-qmin)/100, or something independent of detector spacing

mducle: This was what we had before but it was found to be too coarse, hence we decided to go with the detector spacings, to be more physical. The idea is that the default value is the minimum and if people want to have coarser binning they can change the values themselves.

I would rather not have a step which depends on just a difference between `qmax` and `qmin` exactly because these values don't have anything to do with the detector spacings - e.g. it would give different step sizes for e.g. ARCS/MERLIN vs SEQUOIA/MAPS.

MatthewBowles: @mducle I added an if check for when there are more than 1000 detectors. Does that make sense or should I revert it back to just the spacing?

mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -106,35 +104,34 @@ def _get_property_from_history(name, history):
...
-    qmin, qmax, qstep = tuple(np.sqrt(E2q * 2 * efix * (1 - np.cos(theta)) * meV2J) / m2A)
-    qmax = np.sqrt(E2q * (2 * efix + emax - 2 * np.sqrt(efix * (efix + emax)) * np.cos(theta[1])) * meV2J) / m2A
+def get_q_limits(theta, en, efix, ndetectors):
+    #calculates the Q(E) line for the given two theta and then finds the min and max values
+    qlines = [np.sqrt(E2q * (2 * efix - en - 2 * np.sqrt(efix * (efix - en)) * np.cos(tth)) * meV2J) / 1e10 for tth in theta[:2]]
+    qmin = np.nanmin(qlines[0])
+    qmax = np.nanmax(qlines[1])
+
+    if(ndetectors > 1000):
+        qstep = (qmax - qmin) / 100
```
mducle: @MatthewBowles @AndreiSavici I would rather instead of having a constant ratio of `qmax-qmin` to use:

```
const_tth = 0.5 * np.pi / 180
qstep = np.sqrt(E2q * 2 * efix * (1 - np.cos(const_tth)) * meV2J) / m2A
```

which uses a constant spacing of 0.5 degrees between the detectors. This is about the same as the detector spacing for MERLIN/ARCS, but bigger than for MAPS/SEQUOIA.

mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -106,35 +104,34 @@ def _get_property_from_history(name, history):
...
+    else:
+        qstep = np.sqrt(E2q * 2 * efix * (1 - np.cos(theta[2])) * meV2J) / m2A
+        qstep = qstep / 3  # Use a step size a bit smaller than angular spacing ( / 3) so user can rebin if they want
```
mducle: Also, we can remove this line I think. I don't know what I was thinking when I put this in... (I think it might have been when we were using this as the minimum and users could not put in smaller steps, in case we run out of memory for the slice).

PR #344, opened 2018-07-13
Improve error when merging different file types on load

### LINKED ISSUE #342: Load & merge does not seem to work
This was found during unscripted testing.

If it try and load and merge two files with the same instrument i get the following error:
```
MSlice was not able to load the selected file
````

screen shot 2018-07-13 at 15 17 40

However if I try and load them separately it works fine.

### PR DESCRIPTION
**Pull request by MatthewBowles**
Attempting to `Load and merge` files of different formats would cause an error. This now displays a more helpful error message.

**To test:**
- Select multiple files
- If they have the same extension (e.g.`.nxs`) the `Load and Merge` button should work as normal.
- If they have different extensions, an error message should say files of different formats cannot be merged.

Fixes #342

### DIFF
--- a/mslice/presenters/workspace_manager_presenter.py
+++ b/mslice/presenters/workspace_manager_presenter.py
@@ -154,7 +154,7 @@ class WorkspaceManagerPresenter(WorkspaceManagerPresenterInterface):
         try:
             add_workspace_runs(selected_ws)
         except ValueError as e:
-            self._workspace_manager_view.display_error(str(e))
+            self._workspace_manager_view._display_error(str(e))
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: mslice/presenters/workspace_manager_presenter.py [hunk 1]
```diff
@@ -154,7 +154,7 @@ def _add_workspaces(self):
         try:
             add_workspace_runs(selected_ws)
         except ValueError as e:
-            self._workspace_manager_view.display_error(str(e))
+            self._workspace_manager_view._display_error(str(e))
```
mducle: I think that line 172 should also be changed to have the underscore.

PR #346, opened 2018-07-16
Export cut to MantidPlot ADS

### LINKED ISSUE #341: Use `MatrixWorkspace` for saved cuts instead of `MDHistoWorkspace`
The Mantid fitting interface only works with `MatrixWorkspace` so the 1D cuts outputted by MSlice (which are `MDHistoWorkspaces` cannot be used directly in the fit - the use has to right-click and do `Convert to MatrixWorkspace` first.

It would be good to either output directly as a `MatrixWorkspace` (which might confuse the 1D cuts with reduced [loaded] data), or to add an option in the 1D tab to convert the workspace to `MatrixWorkspace`.

mducle:
Adds the ability to export a saved cut to the MantidPlot ADS, since now we do not store MSlice workspaces in the ADS, so they are not visible from MantidPlot.

**To test:**

Run MSlice within MantidPlot, and check that in the `MD Histo` tab there is a `Save to ADS` button (when running MSlice outside MantidPlot, this button should be hidden).

Load a dataset and make a cut. Then in the `MD Histo` tab click `Save to ADS`. Check that a `MatrixWorkspace` appears in the list of workspaces in MantidPlot. Plot this workspace and check that the values agree with that plotted in MSlice.

Check that the fitting interface works on this workspace.

Fixes #341

### DIFF
--- a/mslice/app/__init__.py
+++ b/mslice/app/__init__.py
@@ -20,9 +20,9 @@ def show_gui():
     If this is the first call then an instance of the Windows is cached to ensure it
     survives for the duration of the application
     """
-    initialize_mantid()
+    in_mantidplot = initialize_mantid()
     global MAIN_WINDOW
     if MAIN_WINDOW is None:
         from mslice.app.mainwindow import MainWindow
-        MAIN_WINDOW = MainWindow()
+        MAIN_WINDOW = MainWindow(in_mantidplot)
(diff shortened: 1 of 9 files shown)


MatthewBowles:
Everything seems to work, just change the name of the button as discussed and it's good to go.

PR #352, opened 2018-07-24
Give plot objects consistent interface

### LINKED ISSUE #294: Common slice/cut plot methods should be consistent
Some methods common to `SlicePlot` and `CutPlot` have different interfaces - for example, `get_line_data` takes the `Line2D` object as a parameter in `SlicePlot`, but takes the line index in `CutPlot`. It would be less confusing if these were made consistent and may help reduce duplicated code.

MatthewBowles:
Includes common interface for `cut_plot` and `slice_plot` and some minor refactoring.

Fixes #294

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -2,7 +2,6 @@ from itertools import chain
 
 from matplotlib.container import ErrorbarContainer
 from matplotlib.legend import Legend
-from matplotlib.lines import Line2D
 from matplotlib import warnings
 import numpy as np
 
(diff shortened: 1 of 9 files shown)


MatthewBowles:
Is that using MARI data (PSD)? I don't think that error is related to changes made in this PR - I think this is a separate issue...

martyngigg:
@MatthewBowles I can't remember but I'll try again with both kinds and see what happens.

martyngigg:
@MatthewBowles I can't seem to replicate the bug now but I think you are correct that it is probably another case of #360 as I seem to remember I was fiddling with the limits for the plotting. The changes here certainly don't seem to relate to any of this I agree.

PR #355, opened 2018-07-25
Fix deletion of renamed workspaces

### LINKED ISSUE #343: Renaming then deleting a workspace does not work as expected.
Renaming a workspace in MSlice and then trying to delete the workspace seems to delete the workspace but leave the name in the workspace list. 

This is not high priority as users can still do their work so should not be for release 1.1

MatthewBowles:
Workspaces could not be deleted after they were renamed because
the workspace.name property was not updated correctly.

Also moved method that handles renaming to _workspace_provider
(from workspace_algorithms) to simplify calls.

**To test:**
- Load workspace
- Rename workspace
- Delete workspace

Fixes #343


martyngigg:
@MatthewBowles There seem to be some unit tests failing.

martyngigg:
@MatthewBowles Just one more left that is failing

MatthewBowles:
Good point, I forgot about the tests...

Looking at them they are now in the wrong place and don't check if `workspace.name` was updated, so they could do with some work. I'll make an issue, but I think it's low priority.

PR #363, opened 2018-08-03
Refactor plotting to use mantid projection

### LINKED ISSUE #358: Use mantid projection for plotting
The mantid projection defined [here](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/mantid/plots/__init__.py) can be used for plotting directly from workspaces, simplifying a lot of the existing plotting code. It will also make implementing the plotting commands for the CLI significantly easier.

MatthewBowles:
- All plotting is now done directly from workspace objects.
- Caching of slice-specific workspace data e.g. intensities is saved in `SliceCache` class.
- Plot window is deleted when closed, so state of checkboxes etc is not saved.

**To test:**
To test MSlice PRs copy the latest commit SHA1 hash (here it is `f75fe63d567a0c10bfd4609390c05fb2487ae3f4` at the moment) and paste it into [scripts/ExternalInterfaces/CMakeLists.txt](https://github.com/mantidproject/mantid/blob/1bb2c9824772b1d1fb1efaea190c03389f7b5700/scripts/ExternalInterfaces/CMakeLists.txt#L9). CMake should pull this commit when you rebuild mantid.

- Load a workspace  [MAR21335_Ei60.00meV.nxs.zip](https://github.com/mantidproject/mslice/files/2269886/MAR21335_Ei60.00meV.nxs.zip)

Test Slice
- Under 'Slice', enter some values in the intensity fields
- Display the slice. The entered values should be the limits of the colorbar.
- Check all of the options in the Information and Intensity menus at the top work without errors.
- Check that an overplotted line's appearance can be changed by double-clicking it.
- Close and reopen slice window, then check that any options selected in the Information/Intensity menus have been cleared.
- Check that clicking interactive cuts and drawing a rectangle produces a cut window correctly.

Test Cut
- Click the cut tab, with a workspace selected enter some values for the first and second axis (not width).
- Check that a line appearance can be changed by double-clicking
- With the plot still open, modify the values in the cut tab and click 'plot over.' Line should appear on the same plot and legend should be updated.
- Check that the axis ranges and labels can be edited by double-clicking on them.

Fixes #358


MatthewBowles:
@mducle I noticed that with the changes in this PR the intensity options rescale the color axis, whereas before they changed the values but kept the same axis min/max.

Which do you think is more intuitive?

TWJubb:
# Test Slice
----------

Apologies for the length of this, I have considered the following as bugs, but some may be intended behavior. None of them actually prevent the GUI working as I can always just plot from scratch at the click of a button; overall functionality works well and feels smooth, especially the cut selection which is a really nice feature.

I have ordered them similarly to the testing instructions (and just for the first test in this comment)

The Intensity Menu
------------------

- If I display data, then change intensity (to e.g. `Chi''(Q,E)`) and give `t_head` as the option in the dialog that appears. Then I select "keep" on the plot and change the intensity again. A new plot appears (as expected), but now I can select several intensity options at once in the new plot, and none of them actually execute and change the plot data.

- Additionally to the last point, the old plot is not altered except in the intensity menu where now the selection I just made is selected (but it should not have changed). Additionally, in the new plot, the selection from the old plot is selected. This is a bit difficult to word nicely, but basically the intensity options are the wrong way around.

Minor Style Points : 

- Although I can change to different intensity types through the menu; nothing on the plot itself actually indicates to me which selection I have made. They all manifestly look the same, except for "Chi''(Q,E) magnetic" for which the color-bar label is updated. Maybe you could update the color-bar to different text for each option.


The Information Menu
--------------------

- If I select Interactive Cuts, but don't actually create a cut yet, then when trying to add the information lines, they actually appear in their own plot window. This may be intentional. But if I now close the new plot window, the information line **remains selected** in the menu. The same thing happens for the "CIF file" option in the "Bragg Peaks" sub-menu

- Further to the above point, if I then unselect the item from the menu, and new window appears with a blank plot.

- If I have an interactive plot open (i.e. I select *Interactive Cuts* and actually define a cut), I again can't add any information lines (if I try all that happens is the cut rectangle disappears). If I try, the information does **remain selected** in the information menu, but no line appears on the plot or anywhere else.

- If I select "arbitrary nuclei" from the menu a dialog appears. If I cancel the dialog, the item becomes selected in the menu anyway.

- In the "arbitrary nuclei" dialog you allow negative numbers (i.e. negative relative mass). This could be ok depending on the definition of relative mass, but I suspect it should never be negative.

- Trying to add an arbitrary nucleus with relative mass zero does nothing to the GUI, but I see a RuntimeWarning divide by zero error in the console output. 

- Additionally to the above, a legend item actually appears for the zero mass nuclei (which it shouldn't since it isn't plotting anything).

- Deuterium has relative mass 2, if I add an 'arbitrary nuclei' of mass 2, it displays in the legend as deuterium (nice!) but it leaves Deuterium unselected in the information menu, so I can add a duplicate by selecting Deuterium. You may want to prevent this edge case.

- I like the Regex (?) on the arbitrary nuclei dialog; I can't enter anything stupid! Just a very minor note; you have an upper limit on what can be entered in the box, but the "^" button still allows me to exceed that number.

Minor style points : 

- Your legend does not distinguish between "Bragg Peaks" and "Recoil Lines". However, the distinction can easily be seen from the plot unless you gave them the same colour and style for some reason.


Interactive Cuts Option
-----------------------

- If I change the intensity option with the interactive cuts on, the plot appears in a new window (leaving the old window and interactive cut window both open). However I can't use the interactive cuts feature on it, even after closing the previous plot and interactive cut plot. If I then hit display in the main window to refresh the plot, this time a blank window appear when switching interactive cuts on.

- If I create an interactive cut plot, then close the window, it reappears when I hover over the main plot (and doesn't reappear otherwise). This may be intentional but its not what I expected to happen, I expected that closing the interactive cut plot just closed the window for good and also cleared the selected cut rectangle on the main plot and deselects "Interactive Cuts".

- I can double click the interactive cut plot and change the color/markers for the data. But when I hover back over the cut plot (this is only occasionally, or more often if I switch the focus to the main plot) the formatting returns to its default and clears all my changes.

- In the interactive cut plot, if I choose to untick the "show" option in the plot style dialog (from double clicking the line), it disappears (obviously), but then I can't get it back :(.

- If I open the interactive cuts, and then close the cut plot window, and **then** deselect interactive cuts from the main plot; a blank interactive cuts window pops-up. I would expect the deselection to simply do nothing in this case.

- When editing the y-axis in the interactive plot, I am checking "logarithm", and I can't seem to manually set the lower end of the range.

- Let's say I choose a cut which includes negative values, and then switch the axis to logarithmic, the behavior of the plot window is very erratic. Sometimes it cuts the axis at zero and plots the rest logarithmically (this is the behavior I would expect). Other times it just does nothing, other times it changes the plot up until I hover over it and then reverts it back. If I hover over the main plot again it always reverts back.

- Similar to the previous point; if I change the axis to logarithmic, as soon as I hover over the main plot, the axis of the interactive cut plot go back to linear.

MatthewBowles:
@TWJubb Thank you for the comments. However, there's no way I can address all of these in this PR and I think they're all present in `master`. 
One of your points is addressed in #365. I'll be adding issues for the remaining ones and dealing with them separately.

mslice/models/slice/slice_functions.py
```diff
@@ -28,139 +30,83 @@ def compute_slice(selected_workspace, x_axis, y_axis, norm_to_one):
...
-    :param reverse: If true then the axis should have values in decreasing order
-    :return: A new numpy array containing the axis values
-    """
-    if reverse:
-        values = np.linspace(axis.end, axis.start, get_number_of_steps(axis))
-    else:
-        values = np.linspace(axis.start, axis.end, get_number_of_steps(axis))
-    return values
+def axis_values(axis):
+    '''Compute a numpy array of bins for the given axis values'''
```
TWJubb: Should these be `"""`?

mslice/models/cut/matplotlib_cut_plotter.py
```diff
@@ -15,49 +16,55 @@ def __init__(self):
...
                                  integration_axis.end)
-        self.plot_cut_from_xye(x, y, e, cut_axis.units, selected_workspace, (intensity_start, intensity_end),
-                               plot_over, output_ws_name, legend)
+        self._plot_cut_impl(cut, cut_axis.units, (intensity_start, intensity_end), plot_over, legend)
+
+    def plot_interactive_cut(self, selected_workspace, cut_axis, integration_axis, norm_to_one, intensity_start,
+                             intensity_end, plot_over, store=True):
+        self.plot_cut(selected_workspace, cut_axis, integration_axis, norm_to_one, intensity_start, intensity_end,
+                      plot_over, store)
+        cur_fig = plt.gcf()
```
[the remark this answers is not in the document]
MatthewBowles: it's short for `current` - `gcf()` in `pyplot` stands for `get current figure`. Is this unclear?

mslice/util/numpy_helper.py
```diff
@@ -0,0 +1,25 @@
...
+    '''
+    if array.ndim == 1: # make 2D
+        array = array[:, np.newaxis]
+    if array.shape[1] == 1:
+        array = np.transpose(array)
+    if shape_array.shape[0] in array.shape and shape_array.shape[0] != 1: # array matches the wrong (leftmost) axis
+        return apply_with_swapped_axes(method, shape_array, array)
+    if shape_array.shape[1] == array.shape[1]:
+        return method(shape_array, array)
+    else:
```
TWJubb: is the `else` necessary here?

mslice/models/slice/matplotlib_slice_plotter.py
```diff
@@ -20,200 +21,118 @@ def __init__(self):
...
-            self.slice_cache[ws]['momentum_axis'] = x_axis
-            self.slice_cache[ws]['energy_axis'] = y_axis
-            self.slice_cache[ws]['rotated'] = False
-        else:
-            self.slice_cache[ws]['momentum_axis'] = y_axis
-            self.slice_cache[ws]['energy_axis'] = x_axis
-            self.slice_cache[ws]['rotated'] = True
+    def _cache_slice(self, slice, colourmap, norm, sample_temp, x_axis, y_axis):
+        rotated = x_axis.units not in ['MomentumTransfer', 'Degrees', '|Q|']
+        q_axis = y_axis if rotated else x_axis
```
TWJubb: Could this be done with a single `if`?

```
q_axis, e_axis = (x_axis, y_axis) if not rotated else (y_axis, x_axis)
```

mslice/models/slice/matplotlib_slice_plotter.py
```diff
@@ -20,200 +21,118 @@ def __init__(self):
...
+        self._show_plot(slice_cache, slice_cache.scattering_function)
+
+    def show_dynamical_susceptibility(self, workspace_name):
+        slice_cache = self.slice_cache[workspace_name]
+        self._show_plot(slice_cache, slice_cache.chi)
+
+    def show_dynamical_susceptibility_magnetic(self, workspace_name):
+        slice_cache = self.slice_cache[workspace_name]
+        self._show_plot(slice_cache, slice_cache.chi_magnetic)
         plt.gcf().get_axes()[1].set_ylabel('chi\'\'(Q,E) |F(Q)|$^2$ ($mu_B$ $meV^{-1} sr^{-1} f.u.^{-1}$)',
```
TWJubb: The following is formatted more in line with convention, and may look a bit neater as its all in latex

`$\\chi '' \\ \\left( Q , E \\right) \\left| F ( Q ) \\right|^2 \\  \\left[ {\\rm mu}_B {\\rm meV}^{-1} {\\rm sr}^{-1} {\\rm f.u.}^{-1} \\right]$`

mslice/models/slice/matplotlib_slice_plotter.py
```diff
@@ -20,200 +21,118 @@ def __init__(self):
...
+        cur_fig.canvas.draw_idle()
+        cur_fig.show()
+
+    def show_scattering_function(self, workspace_name):
+        slice_cache = self.slice_cache[workspace_name]
+        self._show_plot(slice_cache, slice_cache.scattering_function)
+
+    def show_dynamical_susceptibility(self, workspace_name):
+        slice_cache = self.slice_cache[workspace_name]
+        self._show_plot(slice_cache, slice_cache.chi)
```
TWJubb: As per one of my testing comments, you could add a slightly different axis label for each of these cases as otherwise there is no visual difference between them on the plot

mslice/models/slice/slice_cache.py
```diff
@@ -0,0 +1,63 @@
+from mslice.models.slice.slice_functions import (compute_chi, compute_chi_magnetic, compute_d2sigma, compute_gdos,
+                                                 compute_symmetrised)
+
+
+class SliceCache():
+    '''class that caches intensities and metadata for a single workspace'''
```
TWJubb: This class seems to contain all the important data related to a slice, is it worth increasing the documentation to make clear what each of the parameters in here are? This is probably documented elsewhere and if so ignore this comment.

mslice/plotting/plot_window/slice_plot.py
```diff
@@ -150,24 +147,26 @@ def update_legend(self):
...
+        colormesh = current_axis.collections[0]
         vmin, vmax = colorbar_range
-        if logarithmic and type(mappable.norm) != colors.LogNorm:
-            mappable.colorbar.remove()
-            if vmin == float(0):
+        if logarithmic:
+            label = self.colorbar_label
+            colormesh.colorbar.remove()
+            if vmin <= float(0):
                 vmin = 0.001
```
TWJubb: Why `0.001`? Could the colorbar range ever drop below that value?

mslice/models/cut/matplotlib_cut_plotter.py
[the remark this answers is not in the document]
TWJubb: Ah... its just me being stupid, all this "cut" nomenclature and the fact that "r" and "t" are next to each other on the keyboard threw me.

TWJubb:
Yes there are probably too many points to address here, I hope its still helpful to have the list recorded somewhere. Here are some points for the second test.

These are all basically stylistic points so feel free to ignore them.

# Test Cut
----------

- If I only enter values for the "along" row and hit plot, the red error message at the base of the window doesn't really make sense and I can't infer from it what I did wrong.

- As with the slice plot, if I modify the scale of the axes to logarithmic, I can't change the lower end of the range. 

- Un-checking the show legend option leaves a small square legend border in the top right corner of the plot.

- The Intensity boxes allow the user to change the range on the y-axis of the plot, but the y-axis itself is not labelled as intensity. This was confusing for me, but might it also be to actual users?

- If I select None as the marker type, the error bars still appear, should there also be an option to switch off the error bars for the sake of visual clarity when the data is extremely noisy?

- When I switch to logarithmic axis scaling, especially if only over a small range (say 0.1 to 10) it becomes very difficult to understand the axis without minor tick marks. In the linear case this is less important but still useful. Similarly you could enable a major/minor grid separately, or enable the minor grid in a lighter shade of grey.

mslice/plotting/plot_window/slice_plot.py
MatthewBowles: If I remember correctly, ideally this would be zero but that causes a bug in matplotlib so we just use something 'close enough'.

mslice/models/slice/matplotlib_slice_plotter.py
MatthewBowles: I'll put this in issue #366 so all the labels are updated at the same time and it stays consistent.

mslice/plotting/pyplot.py
```diff
@@ -411,49 +411,11 @@ def disconnect(cid):
...
-                num = get_fignums()[allLabels.index(arg)]
-                GlobalFigureManager.destroy(num)
-        elif isinstance(arg, Figure):
-            GlobalFigureManager.destroy_fig(arg)
-        else:
-            raise TypeError('Unrecognized argument type %s to close'
-                            % type(arg))
-    else:
-        raise TypeError('close takes 0 or 1 arguments')
+    GlobalFigureManager.figure_closed(num)
```
martyngigg: Out of curiosity is there a reason we need to deviate from the standard `matplotlib` behaviour?

MatthewBowles: I tried calling the method that was here before and I got this:
```
File "C:\Users\ygh95688\PycharmProjects\mslice\mslice\plotting\pyplot.py", line 429, in close
    GlobalFigureManager.destroy(arg)
AttributeError: type object 'GlobalFigureManager' has no attribute 'destroy'
```
I suppose it might be better to just add another method that does what I want? But I didn't see a reason to keep a close method that doesn't work...

martyngigg: Ah I see. I think the issue is that we have reimplemented the Gcf class from [`matplotlib`](https://github.com/matplotlib/matplotlib/blob/3ca813940b0b7a7640948640fe2c9ed1ebdf27e5/lib/matplotlib/_pylab_helpers.py) as https://github.com/mantidproject/mslice/blob/358_mantid_projection_plotting/mslice/plotting/globalfiguremanager.py and not provided all of the methods. Do you think we could bring our implementation into line and keep `pyplot.close` as it stands?

mslice/plotting/plot_window/slice_plot.py
```diff
@@ -199,14 +198,8 @@ def calc_figure_boundaries(self):
...
 
-    def reset_info_checkboxes(self):
-        for key, line in six.iteritems(self._slice_plotter.overplot_lines[self.ws_name]):
-            if str(line.get_linestyle()) == 'None':
-                if isinstance(key, int):
-                    key = self._slice_plotter.get_recoil_label(key)
-                self.plot_window.disable_action(key)
-
     def toggle_overplot_line(self, key, recoil, checked, cif_file=None):
+        self.manager.report_as_current()
```
martyngigg: I'm not sure if this is the best way to achieve this. I can see that something is required to ensure the overplot goes to the correct window but this will mean a change in behaviour with the keep/make current buttons.

@mducle Would this change to the current status make sense to users when switching intensities do you think? I can see it from both sides...

MatthewBowles: The alternative is to pass the `Axes` object into the `slice_plotter` methods which I'm starting to think might be better.

mducle: @MatthewBowles @martyngigg I think switching the current status back to the window where the user has changed the intensity is ok. I guess that it would be intuitive since if they clicked on that window to change the intensity, they should expect it to become current.

MatthewBowles: I don't know what's causing that but it's fixed in #375.

PR #376, opened 2018-08-14
Split cut_plotter into view and presenter

MatthewBowles:
- Use some methods from `cut_presenter` and `matplotlib_cut_plotter` to create `cut_plotter_presenter`.
- Rename remaining `cut_presenter` to `cut_widget_presenter`.

**To test:**
 Ideally check all aspects of cuts and plotting cuts still function correctly.

_No associated issue_


mslice/models/cut/cut_cache.py
```diff
@@ -0,0 +1,78 @@
+
+class CutCache(object):
```
martyngigg: Good call to create a type for this. How about just `Cut` for the type name? I would naively expect a cache to generally be a list of old cuts - this essentially encapsulates a cut as far as I can see?

MatthewBowles: I agree. I'll change the slice equivalent as well.

MatthewBowles:
Thanks Martyn, should be fixed now. 

I'm aware of the remaining problem with the legends control as @aybamidele noticed it earlier on master  - that will be solved in another PR.

martyngigg:
Thanks, Matthew. The legend thing is minor anyway as the toolbar button still works.

PR #377, opened 2018-08-14
CLI - Plot Slice Command

MatthewBowles:
Adds PlotSlice command to CLI that creates the standard figure as if plotted from the GUI.

Pyplot commands are also available by default.

**To test:**
```
ws = Load(...)
slice = Slice(ws, ...)
PlotSlice(slice, IntensityStart=10, IntensityEnd=200, Colourmap='jet')
```


martyngigg:
One question:
If I do
```
ws = Load(...)
slice = Slice(ws, ...)
PlotSlice(slice, IntensityStart=10, IntensityEnd=200, Colourmap='jet')
```

then run

```
plotSlice(slice)
```

should I expect the plot the revert to the default colour map and scales ? It currently doesn't seem to do that.

MatthewBowles:
Yes, I think that makes sense... should be fixed now.

mslice/cli/_mslice_commmands.py
```diff
@@ -125,42 +136,53 @@ def Slice(InputWorkspace, Axis1=None, Axis2=None, NormToOne=False):
...
     cut_axis = _process_axis(CutAxis, 0, workspace)
     integration_axis = _process_axis(IntegrationAxis, 1 if workspace.is_PSD else 2,
                                      workspace, string_function=_string_to_integration_axis)
 
     return mantid_algorithms.Cut(InputWorkspace=workspace, CutAxis=cut_axis.to_dict(),
                                  IntegrationAxis=integration_axis.to_dict(), EMode=workspace.e_mode,
                                  PSD=workspace.is_PSD, NormToOne=NormToOne)
+
+
+def PlotSlice(InputWorkspace, IntensityStart="", IntensityEnd="", Colourmap='viridis'):
```
martyngigg: Can we centralize the default colourmap selection? Maybe in `models/cmap`? If we wanted to change the default then I currently have to change it in a few different places.

mslice/models/cmap.py
```diff
@@ -3,6 +3,7 @@
 """
 from mslice.util import MPL_COMPAT
 
+DEFAULT_CMAP = 'viridis'
```
martyngigg: :+1:

The `allowed_cmaps` function below removes viridis if it is not available (some old Macs still don't have it!). Maybe we rewrite this as:

```python
ALLOWED_CMAPS = ['jet', 'summer', 'winter', 'coolwarm']
if not MPL_COMPAT:
    ALLOWED_CMAPS.insert(0, 'viridis')

DEFAULT_CMAP = ALLOWED_CMAPS[0]
```

and get rid of the function?

PR #381, opened 2018-08-20
#285 rectangle selector now persistent

### LINKED ISSUE #285: Interactive cut marker is sometimes hidden
Clicking plot options while an interactive cut is active hides the rectangle until the plot is clicked again.

aybamidele:
Description of work.
Added a `callback` method to redraw the rectangle selector whenever the canvas is redrawn.
**To test:**

Open mslice, perform a slice cut and an interactive cut, click the settings gear, click "Ok" and observe that the rectangle persists.

Fixes #285

### DIFF
--- a/mslice/plotting/plot_window/interactive_cut.py
+++ b/mslice/plotting/plot_window/interactive_cut.py
@@ -92,4 +92,3 @@ class InteractiveCut(object):
     def window_closing(self):
         self.slice_plot.plot_window.action_interactive_cuts.setChecked(False)
         self.slice_plot.toggle_icut()
-        self._canvas.mpl_disconnect(self.connect_event[3])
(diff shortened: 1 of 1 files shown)


mslice/plotting/plot_window/interactive_cut.py
```diff
@@ -62,6 +64,10 @@ def end_drag(self, event):
         self._canvas.mpl_disconnect(self.connect_event[0])
         self._canvas.mpl_disconnect(self.connect_event[1])
 
+    def callback(self, event):
```
MatthewBowles: I think this name could be better - it's arguably not a callback since it doesn't get passed into a different class, and it doesn't tell us what the method does. Maybe something like `redraw_rectangle`?

aybamidele: Alright

PR #382, opened 2018-08-21
Interactive cut: remove plot options

### LINKED ISSUE #380: Remove plot options functionality when double clicking in interactive cuts
Remove plot options functionality upon double-clicking in interactive cuts. It is irrelevant to have it there.

aybamidele:
Description of work.
Removed the ability to modify lines in an interactive cut.
**To test:**

Double-clicking on the line in interactive cut no longer does anything.

Fixes #380


mslice/plotting/plot_window/cut_plot.py
```diff
@@ -193,14 +194,28 @@ def toggle_errorbar(self, line_index, line_options):
                 element.set_alpha(1)
 
     def is_icut(self, is_icut):
+
+        if is_icut:
+            self.plot_window.action_keep.triggered.connect(self.keep_triggered)
+            self.plot_window.action_make_current.triggered.connect(self.make_current_triggered)
+
```
MatthewBowles: This works but I think there's a simpler way. `is_icut` is always called when interactive cuts is turned on or off, which is the important behaviour here. So I think we can just do:
```
 self._canvas.manager.button_pressed_connected(not is_icut)
 self._canvas.manager.picking_connected(not is_icut)
```
Also, I agree that this method is poorly named. I think it can be changed to something like 'set_interactive_cut_mode`. 
```

Comment (aybamidele) [reply mslice/plotting/plot_window/cut_plot.py]: Alright

File: mslice/plotting/plot_window/cut_plot.py [hunk 2]
```diff
@@ -193,12 +194,16 @@ def toggle_errorbar(self, line_index, line_options):
                 element.set_alpha(1)
 
     def is_icut(self, is_icut):
+        self._canvas.manager.button_pressed_connected(not is_icut)
+        self._canvas.manager.picking_connected(not is_icut)
```
MatthewBowles: Sorry I missed this before - this object already has access to the `plot_figure_manager`, so you can just do `self.manager.button_pressed_connected`. The same is true in `window_closing`.

PR #389, opened 2018-09-04
#371 error messages improved

### LINKED ISSUE #371: Improve cut error message for no integration values
In the cut widget, If I only enter values for the "along" row and hit plot, the red error message at the base of the window doesn't really make sense and I can't infer from it what I did wrong:
```
Invalid axis parameter - start " is not a valid float
```
This is because the `_cut` method in  `cut_widget_presenter` just displays the exception message from the Axis object. Two possible fixes:
- Improve error message in axis class, perhaps to include the units

OR

- Change how `cut_widget_presenter` handles ValueErrors thrown by `_parse_input`.

aybamidele:
Description of work.
Modified the error messages displayed in the status bar. They are more understandable.
**To test:**

Start mslice, load data, try to perform a cut without the axis parameters. The error shown should be clearer.

Fixes #371

### DIFF
--- a/mslice/widgets/cut/cut.ui
+++ b/mslice/widgets/cut/cut.ui
@@ -16,10 +16,47 @@
   <layout class="QGridLayout" name="gridLayout">
    <item row="0" column="0">
     <layout class="QGridLayout" name="gridLayout_12">
+     <item row="0" column="0">
+      <widget class="QLabel" name="label_251">
+       <property name="text">
+        <string>along</string>
+       </property>
+      </widget>
+     </item>
+     <item row="0" column="1">
+      <widget class="QComboBox" name="cmbCutAxis"/>
+     </item>
+     <item row="0" column="2">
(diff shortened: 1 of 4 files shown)


mslice/widgets/cut/cut.ui
```diff
@@ -34,7 +34,11 @@
       </widget>
      </item>
      <item row="0" column="3">
-      <widget class="QLineEdit" name="lneCutStart"/>
+      <widget class="QLineEdit" name="lneCutStart">
```
martyngigg: I'm not sure about the place holder text. In the rest of mantid we generally use placeholder text for showing default values. I'm not sure this makes it much clearer as the `from`/`to` labels are quite good indicators of what you need to provide.

aybamidele: Okay, I'll keep to that standard and remove them. Thank you.

PR #395, opened 2018-09-13
Fix issue with cancelling arbitrary nuclei user input

### LINKED ISSUE #369: Bugs in 'Arbitrary nuclei' option
- If I select "arbitrary nuclei" from the information -> recoil lines menu, a dialog appears. If I cancel the dialog, the item becomes selected in the menu anyway.

- Values zero or below should not be allowed. Currently, zero can be entered but doesn't plot anything.

aybamidele:
Description of work.
Cancelling the arbitrary_nuclei dialog leaves the option deselected. One can now only select integer values greater than zero. 
**To test:**

Perform a slice in mslice. Select "arbitrary nuclei" from the information -> recoil lines menu. Selecting cancel leaves the "arbitrary nuclei" options deselected. It is also only possible to choose integer values greater than 0.

Fixes #369 .

### DIFF
--- a/mslice/app/__init__.py
+++ b/mslice/app/__init__.py
@@ -8,12 +8,14 @@ from mslice.util.mantid import initialize_mantid
 # Module-level reference to keep main window alive after show_gui has returned
 MAIN_WINDOW = None
 
+
 def main():
     """Start the application.
     """
     qapp_ref = QApplication([])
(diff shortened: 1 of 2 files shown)


aybamidele:
@martyngigg I deleted the other branch and made a new one branching off master instead.

PR #397, opened 2018-09-14
129 option change resolution

### LINKED ISSUE #129: Option to change resolution (DPI) of save figure

aybamidele:
Description of work.
Added functionality for users to choose dpi when saving an image
**To test:**

Open mslice, perform a cut or slice. Click the save image icon. In the save dialog, opt to save as an image by choosing a .pdf or .png extension. Click save. Another dialog should come up asking what dpi the user wants. 30 is the minimum and 300 is the default.

Fixes #129 .


aybamidele:
Okay, I don't know why there are conflicts in the `cut_plot.py` file

mslice/plotting/plot_window/plot_figure_manager.py
```diff
@@ -159,7 +159,9 @@ def save_plot(self):
             save_workspaces([workspace], file_path, save_name, ext, slice_nonpsd=True)
         except RuntimeError as e:
             if str(e) == "unrecognised file extension":
-                self.save_image(os.path.join(file_path, save_name))
+                resolution, confirm = QtWidgets.QInputDialog.getDouble(
+                    self.window, 'Resolution', 'Enter image resolution (dpi):', min=30, value=300)
```
mducle: I think there should also be a `max=3000` or something similar here... `matplotlib` fails if you give it too high a resolution (runs out of memory I guess) - I put in `9999` for a laugh and it gave a `RuntimeError`... I don't know what the max really is and a quick google doesn't give anything...

PR #399, opened 2018-09-24
Information intensity options in interactive cut mode

### LINKED ISSUE #368: Information/Intensity options in interactive cut mode
Adding lines to the slice plot through the 'Information' menu while interactive cuts is selected doesn't work and can lead to strange behaviour. Trying to change the intensity has similar results.

> If I have an interactive plot open (i.e. I select Interactive Cuts and actually define a cut), I again can't add any information lines (if I try all that happens is the cut rectangle disappears). If I try, the information does remain selected in the information menu, but no line appears on the plot or anywhere else.

> If I change the intensity option with the interactive cuts on, the plot appears in a new window (leaving the old window and interactive cut window both open). However I can't use the interactive cuts feature on it, even after closing the previous plot and interactive cut plot. If I then hit display in the main window to refresh the plot, this time a blank window appear when switching interactive cuts on.

aybamidele:
Description of work.
Removed bug that prevents interactive cuts (in particular, the `RectangleSelector` widget) from interfacing well with overplot lines. When intensity plot lines are plotted, it clears and changes the axes which the rectangle selector needs to work. The problem was solved by setting the axes in the `RectangleSelector` each time intensity plot lines are plotted.

**To test:**

Open MSlice, perform an interactive cut and then an overplot. It should work.

Fixes #368 .


aybamidele:
@martyngigg I'm finding this one a bit tricky. The information/intensity lines show when in interactive cut mode. Sometimes two rectangle selectors are on the canvas with one of them inactive. To get this, show an information line while in interactive cut mode. Disable interactive cuts and enable it again. Show another information line and then try to move the rectangle selector. 

When I try to show the intensity lines, it all works fine, but I am unable to interact with the rectangle selector. I've been working on this for over a week and thought it's best if I ask for help now.

aybamidele:
@martyngigg 
Currently, it works, but there is a glitch:
Select `interactive_cuts` and make a cut with the `RectangleSelector`.
Plot one of each overplot line in the information and intensity menus.
Deselect `interactive_cuts`.
Select `interactive_cuts` again and make another cut.
Try to plot any other overplot line and move the rectangle selector, it is duplicated.

aybamidele:
@martyngigg 
While this feature of the program is very stable as of now as the glitch is difficult to reproduce at least with the steps I gave, it happens sometimes, especially when there are lots of overplot lines.

sh.exe.stackdump
```diff
@@ -0,0 +1,28 @@
+Exception: STATUS_STACK_OVERFLOW at rip=7FF891154D87
```
martyngigg: Has this file been committed by accident?

mslice/views/slice_plotter.py
```diff
@@ -45,6 +44,12 @@ def _show_plot(slice_cache, workspace):
     ax.set_ylim(y_axis.start)
     cb = plt.colorbar(image, ax=ax)
     cb.set_label('Intensity (arb. units)', labelpad=20, rotation=270, picker=PICKER_TOL_PTS)
+
+    # Because the axis is cleared, RectangleSelector needs to use the new axis
+    # otherwise it can't be used after doing an intensity plot (as it clears the axes)
+    if cur_fig.canvas.manager._plot_handler.icut is not None:
```
martyngigg: Am I right in saying that this is the core of the fix?

sh.exe.stackdump
aybamidele: I think so. I don't know why it keeps coming up. It's not the first time.

aybamidele: Do I delete it?

martyngigg: Yes please.

If you do a `git rm` on it and then commit that it should get rid of it. It's a good idea to check what you are going to commit with `git status` before you do the final commit. Hopefully you'll spot things like this that shouldn't be included.

martyngigg: :+1:

mslice/plotting/plot_window/slice_plot.py
```diff
@@ -216,7 +220,7 @@ def toggle_overplot_line(self, key, recoil, checked, cif_file=None):
         self._canvas.draw()
 
         # Since some methods in the plot figure manager do not check if the plot handler is a slice plot or cut plot
```
martyngigg: I'm not sure this comment fully explains what is happening here. As far as I can tell we call `self.manager.report_as_current()` at the start of the function and this just resets the state - is that correct? If so can we update the comment to reflect this?

aybamidele: My understanding is that `self.manager.report_as_current()` tells the global figure manager that the plot figure contained in `self.manager` is where "plot instructions" should be directed. Is this right?
I'll change the comment to `# Reset current active figure`?

mslice/plotting/plot_window/interactive_cut.py
```diff
@@ -90,5 +91,5 @@ def flip_axis(self):
         self.plot_cut(*self.rect.extents)
 
     def window_closing(self):
+        self.slice_plot.interactive_cuts()
```
martyngigg: Is this a toggle? It's not entirely clear from the name.

aybamidele: Yes, it is. It toggles the interactive cut buttons and creates the icut object. `toggle_icut` is what creates the icut object. `toggle_icut_button` toggles the buttons. The two functions are within `interactive_cuts`. I could change it to `toggle_interactive_cuts`?

martyngigg: Yes actually I think it would be worth tidying up to help with readability of the code.

aybamidele:
@martyngigg I found another issue with interactive cuts that has to do with cut marked as kept and then marked as current:
plot a slice
plot a cut using interactive cut
turn interactive cuts mode off
mark the cut as kept
plot a cut using interactive cut mode again
mark the first cut as current
Then try to flip the axis in the old one, you get this:
`AttributeError: 'NoneType' object has no attribute 'flip_axis'`
I will raise a new issue for this as I suspect that the solution might not be as trivial as the one from yesterday. Is this okay?

martyngigg:
@aybamidele That sounds sensible. This fixes the bulk of problems with the interactive cuts and overlays so let's merge this.

Issue #396, opened 2018-09-13
Keep figure functionality in CLI


mducle:
There should be a way to programatically "keep" or "make current" a figure using the command line interface.

Example syntax:

```
import mslice.cli as mc

ws = mc.Load('data.nxs')
mc.PlotSlice(mc.Slice(ws, '|Q|', 'DeltaE'))
mc.KeepFigure()   # Without arguments means it should apply to the last figure plotted.
mc.PlotSlice(mc.Slice(ws, 'DeltaE', '|Q|'))
```

This should produce two slice plot windows, with the second one showing the data transposed with respect to the first one.

In order to enable book-keeping, the `PlotSlice` and `PlotCut` commands should also return the integer number of the plot window, e.g.:

```
hcs = mc.PlotSlice(mc.Slice(ws, '|Q|', 'DeltaE'))
hcc = mc.PlotCut(mc.Cut(ws, 'DeltaE'))

mc.KeepFigure()   # Sets hcc to "kept" (not hcs)
mc.KeepFigure(hcs)
hcs2 = mc.PlotSlice(mc.Slice(ws, 'DeltaE', '|Q|'))

mc.MakeCurrent(hcc)
mc.PlotCut(mc.Cut(ws, 'DeltaE', '|Q|,0,2'), PlotOver=True)
```

This would also allow a `MakeCurrent` command too.

aybamidele:
@martyngigg what do you think of this issue in relation to the mslice projection we're making

aybamidele:
@mducle I have included this feature request as part of #404. While it is still in progress, I believe you can test the functionality in the pull request.

mducle:
@aybamidele Thanks Andre! It works really well!

aybamidele:
@mducle Is it okay to close this issue?

mducle:
@aybamidele Sure!

PR #406, opened 2018-10-17
Define MSlice Projection for CLI

### LINKED ISSUE #404: Define an mslice matplotlib projection
We want to replace the cli `PlotSlice`/`PlotCut` commands with the standard matplotlib `pcolormesh`/`plot` commands but have them understand mslice inputs.

This is done by creating a new [projection type](https://matplotlib.org/devel/add_new_projection.html#creating-a-new-projection) and defining our custom behaviour within it. there is an example of this in the [mantid](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/mantid/plots/__init__.py) code.

The projection will need to be registered when mslice is first imported.

aybamidele:
Description of work.
Defined an mslice projection for use with the command line interface. One can now use the matplotlib command plot function using the usual matplotlib arguments and mantid arguments as well.
**To test:**

Open MSlice and try to perform a plot using the `plot` command as usual, also try to load a workspace using the `Load` function then perform a cut using `Cut` or a slice using `Slice`. You can also plot the returned workspaces using the same `plot` command in the CLI.

Fixes #404 .


aybamidele:
@martyngigg I have pushed my early work on the projection here. I wanted you to look at it so you know what I'm doing and to see if it's in the right direction. There's currently a bug with an extra window appearing, I'm looking into it. Is there a reason why the builds are failing?

mslice/cli/_mslice_commands.py
```diff
@@ -172,42 +173,7 @@ def Cut(InputWorkspace, CutAxis=None, IntegrationAxis=None, NormToOne=False):
...
-    workspace = get_workspace_handle(InputWorkspace)
-    if not isinstance(workspace, HistogramWorkspace):
-        raise RuntimeError("Incorrect workspace type.")
-    if IntensityStart == 0 and IntensityEnd == 0:
-        intensity_range = None
-    else:
-        intensity_range = (IntensityStart, IntensityEnd)
-    app.MAIN_WINDOW.cut_plotter_presenter.plot_cut_from_workspace(workspace, intensity_range=intensity_range,
-                                                                  plot_over=PlotOver)
+def plot(*args, **kwargs):
```
martyngigg: We were okay with having the users specify the projection type in their scripts so we shouldn't need this function at all I don't think. I also think the `matplotlib.pyplot` import is what's causing the build failures.

mslice/cli/__init__.py
```diff
@@ -1,2 +1,93 @@
-from ._mslice_commands import * # noqa: F401
-from mslice.plotting.pyplot import * # noqa: F401
+from __future__ import (absolute_import, division, print_function)
+
+from ._mslice_commands import *  # noqa: F401
+
+import mantid.kernel
+import mantid.plots.plotfunctions
```
martyngigg: I don't think we need the `mantid.plots` imports?

mslice/cli/__init__.py
```diff
@@ -1,2 +1,93 @@
...
+
+
+class MSliceAxes(Axes):
+    name = 'mslice'
+
+    def plot(self, *args, **kwargs):
+
+        if validate_args(*args):
+            mantid.kernel.logger.debug('using mantid.plots.plotfunctions')
+            if is_cut(*args):
```
martyngigg: This nearly there except I think we want to keep the plot types separate. `plot` should just create a `cut` and a separate command `pcolormesh` (and its variants) will create the slice.

mslice/cli/__init__.py
```diff
@@ -1,2 +1,93 @@
...
+    """
+    _check_workspace_name(InputWorkspace)
+    workspace = get_workspace_handle(InputWorkspace)
+    if not isinstance(workspace, HistogramWorkspace):
+        raise RuntimeError("Incorrect workspace type.")
+    if IntensityStart == 0 and IntensityEnd == 0:
+        intensity_range = None
+    else:
+        intensity_range = (IntensityStart, IntensityEnd)
+    app.MAIN_WINDOW.cut_plotter_presenter.plot_cut_from_workspace(workspace, intensity_range=intensity_range,
```
martyngigg: I know this is how the cutting currently works, as in it keeps the GUI in step too, I'm not sure this is what we want.

I think it would be better if we don't have it update the GUI at all and we'll figure something else out for that later. I think what we probably need here is to use a separate `CutPlotterPresenter` instance for the cli. I'd create one at module scope here.

mslice/cli/_mslice_commands.py
aybamidele: The builds fail for the same reason in spite of removing `matplotlib.pyplot`

mslice/cli/__init__.py
```diff
@@ -1,2 +1,209 @@
...
+    else:
+        intensity_range = (IntensityStart, IntensityEnd)
+    manager = CLI_CUT_PLOTTER_PRESENTER.plot_cut_from_workspace(workspace, intensity_range=intensity_range,
+                                                                plot_over=PlotOver)
+    figure_manager = manager._current_figs
+    current_figure_number = manager.number
+    return manager
+
+
+def KeepFigure(figure_number=None):
```
martyngigg: Can I suggest using [globalfiguremanager](https://github.com/mantidproject/mslice/blob/master/mslice/plotting/globalfiguremanager.py#L229) directly and update `set_figure_as_kept` to accept a `None` argument?

mslice/cli/__init__.py
```diff
@@ -1,2 +1,209 @@
...
+    global figure_manager
+    global current_figure_number
+
+    if figure_number is not None:
+        figure_manager.set_figure_as_kept(figure_number)
+    else:
+        figure_manager.set_figure_as_kept(current_figure_number)
+
+
+def MakeCurrent(figure_number=None):
```
martyngigg: Can I suggest using [globalfiguremanager](https://github.com/mantidproject/mslice/blob/master/mslice/plotting/globalfiguremanager.py#L245) directly and update set_figure_as_current to accept a None argument?

mslice/cli/__init__.py
```diff
@@ -1,2 +1,209 @@
...
+class MSliceAxes(Axes):
+    name = 'mslice'
+
+    def plot(self, *args, **kwargs):
+
+        if is_cut(*args):
+            try:
+                mantid.kernel.logger.debug('using mantid.plots.plotfunctions')
+                PlotCut(*args, **kwargs)
+                return current_figure_number
```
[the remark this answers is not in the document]
aybamidele: This would not return the figure number as requested by mcducle in #396 ?

mslice/cli/__init__.py
```diff
@@ -1,2 +1,209 @@
...
+        else:
+            return Axes.plot(self, *args, **kwargs)
+
+    def pcolormesh(self, *args, **kwargs):
+
+        if is_slice(*args):
+            try:
+                mantid.kernel.logger.debug('using mantid.plots.plotfunctions')
+                PlotSlice(*args, **kwargs)
+                return current_figure_number
```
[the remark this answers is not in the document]
aybamidele: Same as above

mslice/cli/__init__.py
```diff
@@ -1,2 +1,209 @@
...
+    global current_figure_number
+    global figure_manager
+
+    _check_workspace_name(InputWorkspace)
+    workspace = get_workspace_handle(InputWorkspace)
+    _check_workspace_type(workspace, HistogramWorkspace)
+    slice_presenter = CLI_SLICE_PLOTTER_PRESENTER
+
+    # slice cache needed from main slice plotter presenter
+    slice_presenter._slice_cache = app.MAIN_WINDOW.slice_plotter_presenter._slice_cache
```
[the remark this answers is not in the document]
martyngigg: I'm going to suggest we meet with Duc soon as I think we need to decide whether the CLI should control the GUI or not...

mslice/cli/__init__.py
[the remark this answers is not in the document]
martyngigg: As discussed in the meeting we will keep out commands `PlotCut` and `PlotSlice` returning the figure number but the new projection `Axes` should return the `Line2D` list.

mslice/cli/__init__.py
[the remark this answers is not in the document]
martyngigg: As discussed in the meeting we will keep out commands `PlotCut` and `PlotSlice` returning the figure number but the new projection `Axes` should return the same as `pcolormesh`.

mslice/cli/__init__.py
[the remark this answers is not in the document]
aybamidele: This will be done in #410

PR #407, opened 2018-10-22
Re #393 - fixed problem with propagating limits for Workspace

### LINKED ISSUE #393: CLI binary operator bug
Workspace properties such as `limits` are not correctly propagated when using the command line interface.

To reproduce use:

```
import mslice.cli as mc
ws1 = mc.Load('MAR21335_Ei60.00meV.nxs')
wsc1 = mc.Cut(ws1, 'DeltaE')  # Runs ok
ws2 = ws1 * 2
wsc2 = mc.Cut(ws2, 'DeltaE, -5, 55, 0.5', '|Q|, 0, 10')  # Runs ok.
wsc2 = mc.Cut(ws2, 'DeltaE')  # Gives a KeyError
```

The third `Cut` command (after the scaling operation) gives a `KeyError` because it has to infer the limits of the data, which should have been calculated in the `Load` stage and should be propagated to all child workspaces, but in the scaling operator (`ws2 = ws1 * 2`) is not.

Same errors occur for other binary operations (`+, -, /`).

mducle:
Fix for `rewrap` method of `Workspace` to so that limits are also propagated. Note that no checks are performed. For all current uses of `rewrap` (to propagate MSlice specific properties to child workspaces) such as binary and unary operators this is correct but if `rewrap` is later used for other algorithms, this might be cause errors...

**To test:**

Test that the script in the issue work without errors.

Fixes #393 .

### DIFF
--- a/mslice/workspace/workspace.py
+++ b/mslice/workspace/workspace.py
@@ -27,4 +27,5 @@ class Workspace(WorkspaceMixin, WorkspaceBase):
         new_ws.e_mode = self.e_mode
         new_ws.e_fixed = self.e_fixed
         new_ws.ef_defined = self.ef_defined
+        new_ws.limits = self.limits
         return new_ws


aybamidele:
@mducle Have you any idea why the builds keep failing? I says something about pyqt5 not installable.

PR #411, opened 2018-11-07
Fix issues when running with matplotlib > 2.1

martyngigg:
Description of work.

Mantid is upgrading matplotlib on Windows and this causes a few issues in mslice. This fixes
them in a backward compatible manner. For more information see the commit messages.

A few things to test:

* double-click overplot line on slice plot and check you can change the options
* double click on cut and do the same as above
* use the settings button on cut window to change line styles

### DIFF
--- a/mslice/models/colors.py
+++ b/mslice/models/colors.py
@@ -1,0 +1,106 @@
+"""
+Handles mapping between color representations and their english
+names. For example we want to populate dialogs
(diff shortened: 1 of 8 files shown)


aybamidele:
I've tested it and it looks alright. One thing I noticed (which might have been there already) is that when you perform a slice and change the default overplot line properties and then do an intensity plot, the overplot line properties are not preserved.

PR #412, opened 2018-11-10
Allow for users to use CLI commands in MSlice without MAINWINDOW running

### LINKED ISSUE #410: CLI commands should be able to work without the GUI
Currently, if you run a script from plain python with CLI commands in it then `mslice` complains that the `MAIN_WINDOW` has not been created. The commands should be able to work without having started the GUI but if the GUI has been created then it should proceed as it currently does.

aybamidele:
Description of work.
I created a CLI presenter with all the relevant objects used in the MAINWINDOW to allow users run the CLI commands without the GUI
**To test:**

Fixes #410


mslice/cli/__init__.py
```diff
@@ -1,2 +1,38 @@
...
+class MSliceAxes(Axes):
+    name = 'mslice'
+
+    def plot(self, *args, **kwargs):
+        from mslice.cli.cli_mslice_projection_functions import PlotCutMsliceProjection
+        if is_cut(*args):
+            try:
+                lines = PlotCutMsliceProjection(*args, **kwargs)
+                return lines
+            except Exception as e:
```
martyngigg: Do we need to catch exceptions here? I think it would be better to let them out.

mslice/cli/__init__.py
```diff
@@ -1,2 +1,38 @@
...
+                print('Mslice Projection Error: ' + repr(e))
+        else:
+            return Axes.plot(self, *args, **kwargs)
+
+    def pcolormesh(self, *args, **kwargs):
+        from mslice.cli.cli_mslice_projection_functions import PlotSliceMsliceProjection
+        if is_slice(*args):
+            try:
+                mantid.kernel.logger.debug('using mantid.plots.plotfunctions')
```
martyngigg: I think this comment has been copied from `mantid`?

mslice/cli/_mslice_commands.py
```diff
@@ -8,199 +7,184 @@
...
-        if isinstance(workspace, MatrixWorkspace):
-            raise RuntimeError("Incorrect workspace type - run MakeProjection first.")
-        if not isinstance(workspace, correct_type):
-            raise RuntimeError("Incorrect workspace type.")
-    else:
-        if not isinstance(workspace, MatrixWorkspace):
-            raise RuntimeError("Incorrect workspace type.")
+# Show function to keep window running and not disappearing
+def show():
+    qpp.show()
```
martyngigg: Same typo as in the module name mentioned above.

mslice/app/qpp.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def create_qapp():
+    """
+    Create a qapplication for non gui plots
+    """
+    global qApp
+
+    if qApp is None:
+        qApp = QApplication([])
```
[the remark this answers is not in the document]
martyngigg: This may help with the segfault on the tests

mslice/plotting/globalfiguremanager.py
```diff
@@ -226,8 +238,15 @@ def get_category(cls, num):
         return figure_category
 
     @classmethod
-    def set_figure_as_kept(cls, num):
+    def set_figure_as_kept(cls, num=None):
```
martyngigg: Could you help me understand why `num` can now be `None`?

mslice/plotting/plot_window/plot_figure_manager.py
```diff
@@ -12,6 +11,8 @@
 from mslice.plotting.plot_window.slice_plot import SlicePlot
 from mslice.plotting.plot_window.cut_plot import CutPlot
 import mslice.plotting.pyplot as plt
+import mslice.app.qapp as qpp
```
martyngigg: Typo in `qpp`. I think this should be `qapp`?

mslice/plotting/plot_window/plot_figure_manager.py
```diff
@@ -28,7 +29,11 @@ def __init__(self, number, current_figs):
         self._current_figs = current_figs
 
         # window instance
-        self.window = PlotWindow(manager=weakref.proxy(self))
+        if not is_gui():  # if in cli only mode
```
martyngigg: How about we change `create_qapp` -> `create_qapp_if_required()` and then we can just do 
```
qapp.create_qapp_if_required()
self.window = PlotWindow(manager=weakref.proxy(self))
```
It makes the code a little easier to follow.

mslice/app/qapp.py
```diff
@@ -0,0 +1,20 @@
...
+qApp = None
+
+
+def create_qapp():
+    """
+    Create a qapplication for non gui plots
+    """
+    global qApp
+
+    if qApp is None:
```
martyngigg: It would be safer if we have
```python
if qApp is None:
    instance = QApplication.instance() 
    if instance is None:
        instance = QApplication['mslice']
        instance.lastWindowClosed.connect(instance.quit)
    qApp = instance

return qApp
```
as it could be possible that `qApp` is `None` but the `QApplication` object has been created by something else.

mslice/presenters/cut_plotter_presenter.py
```diff
@@ -86,9 +87,10 @@ def get_icut(self, workspace_name):
         except KeyError:
             return None
 
-    def update_main_window(self):
-        self._main_presenter.highlight_ws_tab(2)
-        self._main_presenter.update_displayed_workspaces()
+    def update_main_window(self, is_gui=True):
+        if is_gui:
```
martyngigg: Is it possible to make the `_main_presenter` optional so we just have `None` and check if `self._main_presenter` is `None`?

mslice/presenters/powder_projection_presenter.py
```diff
@@ -57,10 +57,11 @@ def _calculate_powder_projection(self):
     def calc_projection(self, workspace, axis1, axis2, units):
         return self._projection_calculator.calculate_projection(workspace, axis1, axis2, units)
 
-    def after_projection(self, projection_workspaces):
-        self._get_main_presenter().update_displayed_workspaces()
-        self._get_main_presenter().change_ws_tab(1)
-        self._get_main_presenter().set_selected_workspaces(projection_workspaces)
+    def after_projection(self, projection_workspaces, is_gui=True):
+        if is_gui:
```
martyngigg: Again here. If the main_presenter was allowed to be `None` we could just test for that here and do nothing if  `self._get_main_presenter()` is `None` meaning we wouldn't need the `is_gui` flag. We could then move the `after_projection` call to `calc_projection` and remove it from `_cli_commands` so there is a consistent behavioue between the GUI/CLI.

mslice/plotting/globalfiguremanager.py
aybamidele: I did this to simplify the code for the `KeepFigure` and `MakeCurrent` commands in #404.
When `figure_num` is `None` (This is only done using the `KeepFigure` and `MakeCurrent` commands), I set num to be the figure number of the last plot that was made and update the _last active_figure variable with num. This always denotes the number of the last figure plotted using mc.PlotCut(...) style commands etc. 
There is a need for a "_last_active_figure" for the cli as the globalfiguremanager sets the active figure to None when you mark a figure as kept, but we want to keep the reference to the last plot figure number as requested in #396.

mslice/app/qapp.py
```diff
@@ -0,0 +1,27 @@
...
+
+qApp = None
+
+
+def create_qapp_if_required():
+    """
+    Create a qapplication for non gui plots
+    """
+    global qApp
+    from mslice.app.presenters import is_gui
```
martyngigg: I'm not sure I see the need for the `is_gui` check?

aybamidele: I thought that was what we were trying to do here or is that taken care of by the .instance check? I brought it here from plot_figure_manager.

PR #415, opened 2018-12-07
Allow users to generate Workspace History in Cut and Slice windows

### LINKED ISSUE #385: Command History
There is a requirement for generating a script from a workspace so that a process can easily be repeated using the command line. 

This can be done using the mantid workspace history - a script can be generated using the `GeneratePythonScript` algorithm. This creates a script from all the algorithms executed on a workspace, including the mslice-specific algorithms `Cut`, `Slice`, and `MakeProjection`. 

Problems to solve:
- There are naming conflicts between the commands in `_mslice_commands` and the algorithm - when reading the history it thinks `Slice` refers to the command when it refers to the algorithm.
- There is currently an issue within mantid where `PropertyManagerProperty` cannot be read back from its string representation (which is used in the generated python script). Since `PropertyManagerProperty` is used in the mslice algorithms to represent an axis, these currently do not work. Attempting to run a generated script gives an error `Invalid value for property... Error setting value from string`.
- `GeneratePythonScript` doesn't like how mslice uses workspace pointers for everything, so the generated script contains stuff like `InputWorkspace=\'__TMP00000000201A48C0\'`
- There _might_ be things in the mslice commands which need to be done that aren't in the algorithm (so won't be repeated using a generated script). 

In addition, it would be nice to have a feature for saving the appearance of plots. Investigate available matplotlib features to achieve this. This could be wrapped alongside `GeneratePythonScript`
in a new command, or they could just run seperately.

aybamidele:
Description of work.
I have included menu buttons to allow users generate workspace history.
The script allows for intensity and intensity lines added to a plot after loading to be generated without the gui.

Things not yet accounted for: Zoom, CIF files for slices. The axes grid checkboxes are not ticked in the plot options if they were selected in the gui.
It is also important to note that `ax = ax.intensity_plot()` must be used this way since the original axis is cleared to perform an intensity plot. This makes the order of the statements important. For example, using the mesh object with this new axis does not work well to change the colorbar properties and using the old axis will not update labels (for example) if an intensity plot has been made.
Example of a generated Slice:

```python
# Python Script Generated by Mslice on 2019-01-28 10:18:18.514000
import matplotlib.colors as colors
import mslice.cli as mc

ws = mc.Load(Filename=r'C:/Users/qzp95127\MAR21335_Ei60.00meV.nxs', OutputWorkspace='MAR21335_Ei60.00meV')
slice_ws = mc.Slice(ws)

fig = plt.gcf()
ax = fig.add_subplot(111, projection='mslice')
mesh = ax.pcolormesh(slice_ws, cmap='viridis', intensity='gdos', temperature=88.0)
mesh.set_clim(0.088, 376.866578324)
mesh.set_norm(colors.LogNorm(0.088, 376.866578324))
cb = plt.colorbar(mesh, ax=ax)
cb.set_label('Intensity (arb. units)8', labelpad=20, rotation=270, picker=5)
ax.set_title('MAR21335_Ei60.00meV8')
ax.set_ylabel('Energy Transfer (meV)8')
ax.set_xlabel('$|Q|$ ($\mathrm{\AA}^{-1}$)8')
ax.grid(True, axis='y')
ax.grid(True, axis='x')
ax.set_ylim(bottom=-30.08, top=58.28)
ax.set_xlim(left=0.273568, right=11.03518)
ax.recoil(workspace='MAR21335_Ei60.00meV', element='Hydrogen', rmm=None)
ax.bragg(workspace='MAR21335_Ei60.00meV', element='Aluminium', cif=None)
ax.recoil(workspace='MAR21335_Ei60.00meV', element='Deuterium', rmm=None)
ax.recoil(workspace='MAR21335_Ei60.00meV', element='Helium', rmm=None)
ax.recoil(workspace='MAR21335_Ei60.00meV', element=None, rmm=8)
ax.bragg(workspace='MAR21335_Ei60.00meV', element='Copper', cif=None)
ax.bragg(workspace='MAR21335_Ei60.00meV', element='Niobium', cif=None)
ax.bragg(workspace='MAR21335_Ei60.00meV', element='Tantalum', cif=None)
mc.Show()
```

Example of a generated Cut
```Python
# Python Script Generated by Mslice on 2019-01-30 12:08:09.443000
import mslice.cli as mc

ws = mc.Load(Filename=r'C:/Users/qzp95127\MAR21335_Ei60.00meV.nxs', OutputWorkspace='MAR21335_Ei60.00meV')
fig = plt.gcf()
ax = fig.add_subplot(111, projection='mslice')

cut_ws_0 = mc.Cut(ws, CutAxis='|Q|, 0.0, 11.0, 1.0', IntegrationAxis='DeltaE, -1.0, 0.0, 0.0', NormToOne=True)
ax.errorbar(cut_ws_0, x_units='|Q|', label='MAR21335_Ei60.00meV -1.00<E<0.00', color='blue', marker='o', ls='-', lw='1.0', intensity_range=(None, None))

cut_ws_1 = mc.Cut(ws, CutAxis='|Q|, 0.0, 11.0, 1.0', IntegrationAxis='DeltaE, 0.0, 1.0, 0.0', NormToOne=True)
ax.errorbar(cut_ws_1, x_units='|Q|', label='MAR21335_Ei60.00meV 0.00<E<1.00', color='green', marker='o', ls='-', lw='1.0', intensity_range=(None, None))

ax.set_title('1')
ax.set_ylabel('Signal/#Events1')
ax.set_xlabel('$|Q|$ ($\mathrm{\AA}^{-1}$)1')
ax.grid(True, axis='y')
ax.grid(True, axis='x')
ax.set_ylim(bottom=0.01, top=1.21)
ax.set_xlim(left=0.01, right=12.01)
mc.Show()
```

**To test:**
Open MSlice and perform various data manipulations. In file, select generate script. Run the script.

Fixes #385 .


mslice/scripting/__init__.py
```diff
@@ -0,0 +1,13 @@
+from mantid.simpleapi import GeneratePythonScript
+from mslice.models.workspacemanager.workspace_provider import get_workspace_handle
+from mslice.util.qt import QtWidgets
+
+
+def generate_script(plot_window, ws_name, filename):
+    if filename is None:
+        path = QtWidgets.QFileDialog.getSaveFileName(plot_window, 'Save File', '', 'Python (.py)')
+        if path is u'':
+            raise RuntimeError("save_file_dialog cancelled")
```
martyngigg: I don't think we should raise an exception in this case. If the user cancelled the dialog then we should just return and let them carry on.

mslice/scripting/__init__.py
```diff
@@ -0,0 +1,13 @@
+from mantid.simpleapi import GeneratePythonScript
+from mslice.models.workspacemanager.workspace_provider import get_workspace_handle
+from mslice.util.qt import QtWidgets
+
+
+def generate_script(plot_window, ws_name, filename):
```
martyngigg: Maybe make `filename=None` a default to make clear it's optional?

mslice/scripting/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+    if filename is None:
+        path = QtWidgets.QFileDialog.getSaveFileName(plot_window, 'Save File', '', 'Python (.py)')
+        if path is u'':
```
martyngigg: It is slightly more pythonic to have an empty string check like:

```
if not path:
```

https://stackoverflow.com/questions/9573244/most-elegant-way-to-check-if-the-string-is-empty-in-python

mslice/scripting/__init__.py
martyngigg: Also the more standard matplotlib terminology would be `figure` so maybe use that as an argument name?

martyngigg:
Following our brief offline discussion we settled on:
```
# Python Script Generated by Mslice on 2019-01-24 09:04:09.460483
import mslice.cli as mc

ws = mc.Load(Filename='/home/dmn58364/mslice/MAR21335_Ei60.00meV.nxs', OutputWorkspace='MAR21335_Ei60.00meV')
slice_ws = mc.Slice(ws)

fig = plt.gcf()
ax = fig.add_subplot(111, projection='mslice')

mesh = ax.pcolormesh(slice_ws, cmap='viridis', intensity='s(q,e)', temperature=33)
cb = plt.colorbar(mesh, ax=ax)
cb.set_label('Intensity (arb. units)', labelpad=20, rotation=270, picker=5)

# def recoil(workspace, element=None, rmm=None)
# def bragg(workspace, element=None, cif=None)

ax.recoil(workspace=slice_ws, element="Hydrogen")
ax.recoil(workspace=slice_ws, rmm=20)

ax.bragg(workspace=slice_ws, element="Hydrogen")

mc.Show()
```
as an idea of all of the pieces.

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,138 @@
...
+
+    script_lines.append('ax.set_ylim(bottom={}, top={})\n'.format(*plot_handler.y_range)
+                        if plot_handler.is_changed('y_range') else '')
+    script_lines.append('ax.set_xlim(left={}, right={})\n'.format(*plot_handler.x_range)
+                        if plot_handler.is_changed('x_range') else '')
+
+
+def add_overplot_statements(script_lines, plot_handler):
+    ax = plot_handler._canvas.figure.gca()
+    line_artists = [artist for artist in ax.get_children() if isinstance(artist, Line2D)]
```
martyngigg: The lines are all stored in the `ax.lines` attribute so we can avoid having to do this loop here.

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,138 @@
...
+
+    for line in line_artists:
+        label = line._label
+        key = 1 if label == 'Hydrogen' else 2 if label == 'Deuterium' else 4 if label == 'Helium' else label
```
martyngigg: I'm a little concerned at the duplication of information having this hard-coded here. The [labels.py](https://github.com/mantidproject/mslice/blob/96e7167a0510fcfef70c60e59e9aa8f7151ca0c6/mslice/models/labels.py#L7) module currently has some information about this. Perhaps we should centralize this stuff by adding methods to the label module such as `get_recoil_label_key`?

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,138 @@
...
+        rmm = int(label.split()[-1]) if "Relative" in label else None
+        element = label if rmm is None else None
+        recoil = True if rmm is not None or key in [1, 2, 4] else False
+
+        if recoil:
+            if element is None:
+                script_lines.append(
+                    'ax.recoil(workspace=\'{}\', element={}, rmm={})\n'.format(
```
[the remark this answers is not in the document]
aybamidele: The difference was the quotation marks around the parentheses we want the element names to be strings. I felt like leaving it the other way would make the two ways of adding recoil lines obvious to the users. I have changed it already.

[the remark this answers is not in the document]
martyngigg: I think we are better going with showing the minimal required in this case as I think users will think you would always have to pass `None`.

martyngigg:
I was just testing the slicing again and it looks like the script doesn't take into account if I change anything on the original workspace manager screen. Anything like `from`, `to` & `step` don't seem to be taken into account in the genrated script.

aybamidele:
@martyngigg I've pushed the latest work I've got so far. A lot of tests have failed. I'll be looking at them later.

aybamidele:
@martyngigg It looks as though the fix we had in mind for script regeneration might be more convoluted than expected. I'm going to branch off this as it's in a reviewable state and continue the work in the new branch.

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,217 @@
...
+    return script_lines
+
+
+def add_slice_plot_statements(script_lines, plot_handler):
+    slice = plot_handler._slice_plotter_presenter._slice_cache[plot_handler.ws_name]
+    momentum_axis = str(slice.momentum_axis)
+    energy_axis = str(slice.energy_axis)
+    norm = slice.norm.vmin == 0.0 and slice.norm.vmax == 1.0
+
+    script_lines.append('slice_ws = mc.Slice(ws, Axis1=\'{}\', Axis2=\'{}\', NormToOne={})\n\n'.format(
```
martyngigg: Just a note that in Python you can mix `"` and `'` for strings so here you could do
```script_lines.append("slice_ws = mc.Slice(ws, Axis1='{}', Axis2='{}', NormToOne={})\n\n".format(```
to avoid having to escape the quotes `'`, which I think would improve readability.

This is true in multiple places in this file I think.

File: mslice/scripting/helperfunctions.py [hunk 8]
```diff
@@ -0,0 +1,217 @@
...
+        momentum_axis, energy_axis, norm))
+    script_lines.append('fig = plt.gcf()\n')
+    script_lines.append('ax = fig.add_subplot(111, projection=\'mslice\')\n')
+
+    if plot_handler.default_options['intensity'] is True:
```
martyngigg: It is good practice to extract frequently accessed objects into their own variables as it reduces duplication, improves readability by giving the object a name and increase performance by reducing redundant work. Here for example we access `plot_handler.default_options` multiple times so it would make sense to extract this into a local variable `default_options` and use this.

A similar argument `plot_handler._slice_plotter_presenter` and the `slice_cache` can be made within this function I think.

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,217 @@
...
+            script_lines.append(
+                'mesh = ax.pcolormesh(slice_ws, cmap=\'{}\', intensity=\'{}\')\n'.format(
+                    plot_handler._slice_plotter_presenter._slice_cache[plot_handler.ws_name].colourmap, intensity))
+    else:
+        script_lines.append(
+            'mesh = ax.pcolormesh(slice_ws, cmap=\'{}\')\n'.format(
+                plot_handler._slice_plotter_presenter._slice_cache[plot_handler.ws_name].colourmap))
+
+    script_lines.append('mesh.set_clim({}, {})\n'.format(*plot_handler.colorbar_range))
+    script_lines.append(
```
martyngigg: It is not always best to opt for the most succinct way expressing things in code. This line is a little difficult to parse due to the multiple uses of the ternary operator. I think something like
```python
if plot_handler.colorbar_log:
    min, max = plot_handler.colorbar_range[0], plot_handler.colorbar_range[1]
    min = max(min, LOG_SCALE_MIN):
```
would be more readable.

mslice/cli/projection_functions.py
```diff
@@ -1,46 +1,108 @@
...
+        plot_handler = cur_canvas.manager.add_cut_plot(presenter, workspace.name.rsplit('_', 1)[0])
+        plot_handler.save_default_options()
+    cur_fig.canvas.draw()
 
-def PlotSliceMsliceProjection(InputWorkspace, IntensityStart="", IntensityEnd="", Colormap=DEFAULT_CMAP):
+    return axes.lines
+
+
+@plt.set_category(plt.CATEGORY_SLICE)
+def PlotSliceMsliceProjection(axes, workspace, *args, **kwargs):
```
martyngigg: These function names don't really follow are naming conventions. They're not commands themselves so I think something like `errorbar` and `pcolormesh` would be fine names for these in a similar way that they are named in `mantid.plots.plotfunctions`. PyCharm rename should be able to help with this.

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,217 @@
+from datetime import datetime
```
martyngigg: Can we add some doccumentation to these functions?

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,217 @@
...
+            index = script_lines.index(line)
+            return script_lines[index:]
+
+
+def header(plot_handler):
+    from mslice.plotting.plot_window.cut_plot import CutPlot
+    from mslice.plotting.plot_window.slice_plot import SlicePlot
+    statements = ['# Python Script Generated by Mslice on {}\n'.format(datetime.now().replace(microsecond=0))]
+
+    for package in PACKAGES:
```
martyngigg: I find this loop a little difficult to read as we have lots of `nots` and `continues`. Should we consider keeping two lists of packages: a base list that is always used and a second list where we can first check the handler and then pull in from the list without needing a loop?

mslice/scripting/helperfunctions.py
aybamidele: @martyngigg I made these changes in the branch I made off this one. Do I just do it again here again?

mslice/scripting/helperfunctions.py
aybamidele: @martyngigg Wouldn't we need a loop for the base list?

mslice/scripting/helperfunctions.py
martyngigg: Ah I see. No that's okay. I hand't looked at the other one yet.

mslice/scripting/helperfunctions.py
martyngigg: I was thinking of something like
```python
statements.append('\n'.join(COMMON_PACKAGES))
if ...test for whether we have log color bar...:
   statements.append(MPL_COLORS_IMPORT)
if we need numpy:
   statements.append(NUMPY_IMPORT)
```
where `COMMON_PACKAGES` is just a list of import line strings.

mslice/presenters/cut_widget_presenter.py
```diff
@@ -96,7 +96,7 @@ def _set_minimum_step(self, workspace, axis):
         self._cut_view.set_minimum_step(self._minimumStep[axis[0]])
 
     def workspace_selection_changed(self):
-        if self._previous_cut is not None and self._previous_axis is not None:
+        if self. _previous_cut is not None and self._previous_axis is not None:
```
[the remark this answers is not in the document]
aybamidele: @mducle Yes, it shouldn't. I thought I removed it last week actually. I'll sort it out.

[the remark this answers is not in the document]
martyngigg: I'm surprised this doesn't produce a syntax error in Python. Do we not have a test that would cover this?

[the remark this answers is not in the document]
aybamidele: We do.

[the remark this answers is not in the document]
martyngigg: If we have a test how come it didn't fail?

mslice/cli/helperfunctions.py
```diff
@@ -4,6 +4,68 @@
...
+    'chi_mag': 'chi_magnetic',
+    'xsec': 'd2sigma',
+    'symm': 'symmetrised',
+    'gdos': 'gdos',
+}
+
+
+def _update_legend():
+    plot_handler = GlobalFigureManager.get_active_figure()._plot_handler
+    plot_handler.update_legend()
```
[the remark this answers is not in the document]
aybamidele: I am unable to reproduce this error, @mducle 
Could I have access to the same file? Or perhaps you could try again when I push to github again?

mslice/presenters/cut_widget_presenter.py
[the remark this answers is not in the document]
aybamidele: I am unable to reproduce the error and it seems to work as expected. Could you send me the steps you took?

mslice/cli/helperfunctions.py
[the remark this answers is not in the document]
mducle: I'll try it again when you've pushed to github... I can't seem to upload the file to github but it's on my folder on: \\Olympic\Babylon5\Public\duc\

mslice/presenters/cut_widget_presenter.py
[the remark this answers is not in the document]
mducle: @aybamidele you're right... it's still all working as it should (on the latest version of this branch)... sorry I think I was confusing this with something else (not sure what now)...

mslice/scripting/helperfunctions.py
```diff
@@ -0,0 +1,217 @@
...
+def add_cut_lines(script_lines, plot_handler):
+    cut = plot_handler._cut_plotter_presenter._cut_cache[plot_handler.ws_name]
+    errorbars = plot_handler._canvas.figure.gca().containers
+    add_cut_lines_with_width(errorbars, script_lines, cut)
+
+
+def add_cut_lines_with_width(errorbars, script_lines, cut):
+    integration_start = cut.integration_axis.start
+    integration_end = cut.integration_axis.end
+    cut_start, cut_end = integration_start, min(integration_start + cut.width, integration_end)
```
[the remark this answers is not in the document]
aybamidele: I have done this by modifying the width setter for the Cut object

mslice/cli/helperfunctions.py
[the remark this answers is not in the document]
mducle: Ok... This is very weird - so when I run the script externally, e.g. with: 

```
mantidpython test1.py
```

it works fine (PS - you can just use an input file you have rather than the one I used - it doesn't matter which file it is for this error).

However, when I try to run the file within Mantidplot or within the IPython console, using e.g.

```
import test1
```

it fails with the `plot_handler` = `None` problem. The same error occurs if I type all the commands into the IPython console, or if I run the file from the Python script window in MantidPlot (accessible using `F3`) in Mantidplot *after* I start the MSlice interface. If I run the file from the script window in MantidPlot *before* starting the MSlice interface it works fine...

So I think there is something in the MSlice GUI initialisation which is causing the problem...

mslice/cli/projection_functions.py
```diff
@@ -1,46 +1,108 @@
...
 
     # slice cache needed from main slice plotter presenter
     if is_gui():
         cli_slice_plotter_presenter._slice_cache = app.MAIN_WINDOW.slice_plotter_presenter._slice_cache
-    cli_slice_plotter_presenter.change_intensity(workspace.name, IntensityStart, IntensityEnd)
-    cli_slice_plotter_presenter.change_colourmap(workspace.name, Colormap)
-    quadmesh = cli_slice_plotter_presenter.plot_from_cache(workspace)
+    else:
+        # Needed so the figure manager knows about the slice plot handler
+        create_slice_figure(workspace.name[2:], get_slice_plotter_presenter())
```
mducle: So the issue with the plot having no `plot_handler` for the generated CLI commands when executed in the GUI comes from this part here. The `plot_handler` reference is only created in this call to `create_slice_figure` which is not run if `is_gui` is `True`. I suggest changing line 72 to:

```
    if is_gui() and GlobalFigureManager.get_active_figure()._plot_handler is not None:
```

aybamidele: Ah I see. I'm looking into it now

aybamidele: Yes I was able to reproduce the error. And this fixes it. Thank you.

PR #417, opened 2018-12-08
Allow mslice scripts to run in MantidPlot and MantidWorkbench

### LINKED ISSUE #400: MSlice CLI does not work in MantidPlot python script editor
MantidPlot crashes when using the MSlice CLI to plot from the Python script window (activated using `F3`) or the script interpreter - e.g.

```
import mslice.cli as mc
mc.PlotSlice(mc.Slice(mc.Load('MAR21335_Ei60.00meV.nxs')))
```

This is because each MantidPlot script window / the script interpreter runs in its own thread which is not the main or GUI thread. When MSlice tries to open up a new Qt window, it runs into memory errors - sometimes MantidPlot just hangs (very occasionally just after painting the new plot window) and sometimes it segfaults. (Note that plotting from the interface is fine because interfaces are launched in the main GUI thread).

NB. We need to decide whether to support MSlice running in the MantidPlot script window... if we do decide to, then need to add code to check for MantidPlot and to switch to the main GUI thread when drawing Qt windows.

martyngigg:
Description of work.

Ensures that all calls to Qt drawing functions are pushed to the main thread.

**To test:**

The following script should run from the command line `mantidpython` as well as in `MantidPlot`

```
# The following line helps with future compatibility with Python 3
# print must now be used as a function, e.g print('Hello','World')
from __future__ import (absolute_import, division, print_function, unicode_literals)

import sys
# use development copy of mslice
sys.path.insert(0, '/media/data1/source/github/mantidproject/mslice')

import mslice.cli as mc

ws = mc.Load("/home/dmn58364/mslice/MAR21335_Ei60.00meV.nxs")
cut_ws = mc.Cut(ws)
slice_ws = mc.Slice(ws)
f1 = mc.PlotCut(cut_ws)  # returns plot number
f2 = mc.PlotSlice(slice_ws)  # returns plot number
mc.Show()
```

You will most likely need to update the paths to development `mslice` and the data file.

Fixes #400.


mslice/cli/_mslice_commands.py
```diff
@@ -1,33 +1,64 @@
...
+        ipython_pylab = not pyplot.show._needmain
+        # IPython versions >= 0.10 tack the _needmain
+        # attribute onto pyplot.show, and always set
+        # it to False, when in %pylab mode.
+        ipython_pylab = ipython_pylab and mpl.get_backend() != 'WebAgg'
+        # TODO: The above is a hack to get the WebAgg backend
+        # working with ipython's `%pylab` mode until proper
+        # integration is implemented.
+    except AttributeError:
+        ipython_pylab = False
```
mducle: Is the above section so that the plotting would work in an IPython Notebook, or is it just for the IPython console which we include in the GUI?

mslice/plotting/globalfiguremanager.py
```diff
@@ -129,7 +138,7 @@ def _new_figure(cls, num=None):
             num = 1
             while any([num == existing_fig_num for existing_fig_num in cls._figures.keys()]):
                 num += 1
-        new_fig = PlotFigureManagerQT(num, GlobalFigureManager)
+        new_fig = QAppThreadCall(PlotFigureManagerQT)(num, GlobalFigureManager)
```
mducle: So it was just a matter of wrapping in a Mantid `QAppThreadCall` in the end?

mslice/cli/_mslice_commands.py
martyngigg: It should cover the notebook too. It's similar too the standard Qt5 backend version of `show` but we need to call `GlobalFigureManager` rather than the standard `matplotlib` one as the former contains our code to track keep/make current so is the class that holds all of our figures.

mslice/plotting/globalfiguremanager.py
martyngigg: This and a similar `QAppThreadCall(mainloop)()` in the mslice `Show` method yes.

PR #419, opened 2019-02-06
Cuts and slices made to be repositioned with respect to the window center

### LINKED ISSUE #418: Plot Window moves off screen when plots are created
When a slice or cut is created, the plot window opens in the top right-hand corner of the screen with half the window visible. One would need to drag the window to the centre of the screen to see all its contents.

Steps to recreate:
Open Mslice
Plot a Slice or a Cut

aybamidele:
Description of work.
The move window command now uses the screen centre to fix the position of cuts and slices
**To test:**
Open Mslice and plot a cut or slice. The windows open near the centre of the screen.

Fixes #418 .

### DIFF
--- a/mslice/plotting/plot_window/plot_figure_manager.py
+++ b/mslice/plotting/plot_window/plot_figure_manager.py
@@ -4,15 +4,13 @@ import weakref
 import six
 from mslice.util.qt.QtCore import Qt
 from mslice.util.qt import QtCore, QtGui, QtWidgets
-from mslice.util.qt.qapp import create_qapp_if_required
-
+from mslice.util.qt.qapp import create_qapp_if_required, QAppThreadCall
 from mslice.models.workspacemanager.file_io import get_save_directory
 from mslice.models.workspacemanager.workspace_algorithms import save_workspaces
 from mslice.plotting.plot_window.plot_window import PlotWindow
 from mslice.plotting.plot_window.slice_plot import SlicePlot
 from mslice.plotting.plot_window.cut_plot import CutPlot
 import mslice.plotting.pyplot as plt
-from mslice.util.qt.qapp import QAppThreadCall
 
 
 class PlotFigureManagerQT(QtCore.QObject):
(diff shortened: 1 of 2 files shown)


mslice/plotting/plot_window/plot_figure_manager.py
```diff
@@ -87,14 +85,14 @@ def flag_as_current(self):
 
     def add_slice_plot(self, slice_plotter_presenter, workspace):
         if self._plot_handler is None:
-            self.move_window(-self.window.width() / 2, 0)
+            self.move_window(self.window.width() * 1.05, self.window.height() / 2)
```
martyngigg: Agreed that this is a fine thing to do. Could we add a code comment about why we are doing it? At first glance it appears a little strange.

mslice/plotting/plot_window/plot_figure_manager.py
```diff
@@ -87,14 +85,14 @@ def flag_as_current(self):
...
         else:
             self._plot_handler.disconnect(self.window)
         self._plot_handler = SlicePlot(self, slice_plotter_presenter, workspace)
 
     def add_cut_plot(self, cut_plotter_presenter, workspace):
         if self._plot_handler is None:
-            self.move_window(self.window.width() / 2, 0)
```
martyngigg: Similar thing with a comment here too.

mslice/plotting/plot_window/plot_figure_manager.py
```diff
@@ -87,14 +85,18 @@ def flag_as_current(self):
 
     def add_slice_plot(self, slice_plotter_presenter, workspace):
         if self._plot_handler is None:
-            self.move_window(-self.window.width() / 2, 0)
+            # Move the top right corner of all slice plot windows to the left of the screen centre by 1.05
```
martyngigg: This is a good description of what is being done but could we add a description of why - that's actually more important for the future.

PR #421, opened 2019-02-07
Allow users generate scripts from generated scripts

### LINKED ISSUE #420: Allow users generate python scripts from a generated scripts plot window
Currently, it is not possible to generate scripts from a generated script plot window.
Allowing users modify their generated script from the gui would be a useful feature.

aybamidele:
Description of work.
Default plot options for cuts and slices and now saved when their respective windows are shown. This allows the generated scripts to know what changes have been made to the plot since it was loaded.
**To test:**
Open Mslice. Plot a slice or cut and generate a script.
Run the generated script, make changes and generate another script.
Observe the difference between the first and second generated script.

Fixes #420 .


martyngigg:
Can we rebase with master now after merging #415?

PR #425, opened 2019-02-13
Script generations now accounts for multiple workspaces used in cuts

### LINKED ISSUE #423: Multiple loaded workspaces not accounted for when python scripts are generated
MSlice currently does not take into account that there might be more than one workspace loaded when a python script is generated.

aybamidele:
Description of work.
Modified python script generation to account for multiple cuts from different workspaces.

**To test:**
Open Mslice. Load multiple workspaces. Using plot over, plot multiple cuts onto the same plot window. In the file options, click generate script.

Fixes #423 .

### DIFF
--- a/mslice/cli/_mslice_commands.py
+++ b/mslice/cli/_mslice_commands.py
@@ -89,7 +89,7 @@ def GenerateScript(InputWorkspace, filename):
     _check_workspace_name(InputWorkspace)
     workspace_name = get_workspace_handle(InputWorkspace).name[2:]
     plot_handler = GlobalFigureManager.get_active_figure().plot_handler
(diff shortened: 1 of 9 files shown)


martyngigg:
@mducle I would be tempted to say we should disable the history from interactive cuts and do it properly. The keep/current stuff for the interactive cuts is already a little fragile and it is probably best that we not mess with that too much? Do you think it would be okay to release it to the wild with the script generation for interactive cuts disabled?

mducle:
Yes. I think it would be ok to just disable the script generation for interactive cuts until we can do it properly.

martyngigg:
@aybamidele Can you open an issue for adding tests here for the multiple workspaces in maintenance?

PR #426, opened 2019-02-14
Projected workspaces can now be plotted and used with the python script generation

### LINKED ISSUE #424: Unable to plot slice from projection
Steps to reproduce:

* load the `MER30936_Ei10.00meV_one2one.nxs` from the drop box test data
* click "Make Projection"
* click "Display Slice"

and an error occurs:
```
Load-[Notice] Load started
Load-[Notice] Load successful, Duration 5.05 seconds
MakeProjection-[Notice] MakeProjection started
MakeProjection-[Notice] MakeProjection successful, Duration 3.62 seconds
Slice-[Notice] Slice started
Slice-[Notice] Slice successful, Duration 0.17 seconds
  File "/media/data1/source/github/mantidproject/mslice/mslice/widgets/slice/slice.py", line 49, in _btn_clicked
    self._presenter.notify(Command.DisplaySlice)
  File "/media/data1/source/github/mantidproject/mslice/mslice/presenters/slice_widget_presenter.py", line 39, in notify
    self._display_slice()
  File "/media/data1/source/github/mantidproject/mslice/mslice/presenters/slice_widget_presenter.py", line 61, in _display_slice
    norm_to_one, colourmap)
  File "/media/data1/source/github/mantidproject/mslice/mslice/presenters/slice_widget_presenter.py", line 74, in _plot_slice
    raise e
ValueError: Warning: To plot a cut use the plot function instead!
```

aybamidele:
Description of work.
I modified the is_slice function to return True for projected workspaces.
I modified the python script generation algorithm to account for projected workspaces.
**To test:**

Open Mslice
Load a workspace
Click Make Projection
Click Display Slice

Fixes #424 .


mslice/cli/_mslice_commands.py
```diff
@@ -92,7 +92,7 @@ def GenerateScript(InputWorkspace, filename):
     generate_script(ws_name=workspace_name, plot_handler=plot_handler, filename=filename)
 
 
-def MakeProjection(InputWorkspace, Axis1, Axis2, Units='meV'):
+def MakeProjection(InputWorkspace, Axis1, Axis2, Units='meV', OutputWorkspace=None, Limits=None):
```
martyngigg: We don't appear to be using these two arguments?

mslice/cli/helperfunctions.py
```diff
@@ -154,6 +154,8 @@ def is_slice(*args):
     Checks if args[0] is a WorkspaceBase or HistogramWorkspace
     """
     if len(args) > 0:
+        if isinstance(args[0], HistogramWorkspace) and args[0].is_PSD:
```
martyngigg: I'm not sure `is_PSD` is the correct check to use here. `is_PSD` just indicates whether the instrument is a PSD-tube-based instrument or not and is not really anything to do with the data type. The projection should return a `PixelWorkspace` I think so a slice should be able to be created if it is a `PixelWorkspace` or a `Workspace`

mslice/cli/helperfunctions.py
```diff
@@ -154,6 +154,8 @@ def is_slice(*args):
     Checks if args[0] is a WorkspaceBase or HistogramWorkspace
     """
     if len(args) > 0:
```
martyngigg: Can we have this function `return` either True or False in all cases rather than returning `None` or raising an exception? `is_`-type functions should return True/False as that's what a caller would expect.

martyngigg: I think something similar for `is_cut` would be good.

mslice/cli/_mslice_commands.py
aybamidele: That is tentative as I still am waiting for the solution to the algorithm being run without setting the workspace name

mslice/cli/helperfunctions.py
aybamidele: I checked and it was a histogram workspace that is being passed as the argument

martyngigg: That's odd as a projection should still be 2D and histogram is 1D.

mducle: Actually I think `is_PSD` is correct here - we are dealing with the slice and not the project workspace itself. The slice is either an `MDHistoWorkspace` generated from `BinMD` if the parent is from a PSD instrument (parent is a `PixelWorkspace`) or a `MatrixWorkspace` generated from `SofQW3` if the parent is from a non-PSD instrument (parent is also a `MatrixWorkspace`).

martyngigg: My bad for not remember that!
Should we really check the dimensionality of the data though rather than rely on the fact that for some more instrument-specific check?

mducle: Yeah... a `getNumDims() == 2` check is probably better! It would still be valid in future if we should change the code somehow.

mslice/cli/_mslice_commands.py
aybamidele: Have you found anything that could help with this?

martyngigg: Not fully. I think we need to set the name for the `InputWorkspace` argument as well as the pointer variable but I can't see a clean way to do this yet.

aybamidele: Is it okay to say that this is what we're waiting for for this pull request to be closed. I have opened an issue to add tests.

mslice/plotting/plot_window/slice_plot.py
```diff
@@ -111,6 +111,7 @@ def setup_connections(self, plot_window):
         plot_window.action_gen_history.triggered.connect(partial(generate_script, self.ws_name, None, self,
```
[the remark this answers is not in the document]
aybamidele: Yes, I've spotted that. I'm about to push some changes.

aybamidele:
@mducle Is it possible to have two slices on the same plot? Does mslice not clear the figure each time a new one is plotted?

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
-        script_lines = ["\nws = mc."] + cleanup(GeneratePythonScript(ws).split('\n'))
+        script_lines += generate_script_lines(ws, ws_name)
 
     return script_lines
+
+
+def generate_script_lines(raw_ws, workspace_name):
+    lines = []
+    ws_name = workspace_name.replace(".", "_")
+    alg_history = raw_ws.getHistory().getAlgorithmHistories()[::-1]
```
martyngigg: There is a function in python called `reversed` which is handy for doing this:
```python
alg_history = raw_ws.getHistory().getAlgorithmHistories()
for algorithm in reversed(alg_history)
...
```
It makes it easier to read as it's easy to forget what `[::-1]` does.

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
+    for algorithm in alg_history:
+        alg_name = algorithm.name()
+        kwargs = get_algorithm_kwargs(algorithm, ws_name)
+        if alg_name != 'Load':
+            lines += ["ws_{} = mc.{}({})\n".format(ws_name, alg_name, kwargs)]
```
martyngigg: I think this line is exactly the same in both blocks? Can it be above the `if` and then just the break is in the `if`?

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
+        else:
+            lines += ["ws_{} = mc.{}({})\n".format(ws_name, alg_name, kwargs)]
+            break
+    return lines[::-1]
```
martyngigg: Again I think we can use `reversed(lines)` here?

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
+
+
+def get_algorithm_kwargs(algorithm, workspace_name):
+    arguments = []
+    if algorithm.name() == "Load":
+        for property in algorithm.getProperties():
```
martyngigg: I would avoid the use of the variable name `property` as this is a keyword in Python. `prop` should be fine.

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
+            if property.name() == "Filename":
```
martyngigg: I think it might be better to blacklist the properties we don't want, i.e skip `LoaderName`, `OutputWorkspace` and any that are `isDefault` so that any added in the future would not be trapped by this.

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
+                elif property.name() == "OutputWorkspace" or property.name() == "Limits":
+                    pass
+                else:
+                    arguments += ["{}='{}'".format(property.name(), property.value())]
+    else:
+        for property in algorithm.getProperties():
+            if not property.isDefault():
+                if property.name() == "Filename":
+                    arguments = ["{}='{}'".format(property.name(), property.value())]
+                else:
```
martyngigg: Is this guaranteed to be true for all argument types?

mslice/scripting/__init__.py
```diff
@@ -22,15 +21,57 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None):
...
+                    arguments += ["{}=ws_{}".format(property.name(), property.value())]
+    return ", ".join(arguments)
```
martyngigg: Good use of join :+1:

mslice/scripting/helperfunctions.py
```diff
@@ -157,7 +147,7 @@ def add_cut_lines(script_lines, plot_handler):
 
 def add_cut_lines_with_width(errorbars, script_lines, cuts, plot_handler):
     """Adds the cut statements for each interval of the cuts that were plotted"""
-    more_than_one_ws = len(plot_handler._cut_plotter_presenter._cut_cache) > 1
+    #more_than_one_ws = len(plot_handler._cut_plotter_presenter._cut_cache) > 1
```
martyngigg: From the removal of the `if` below I assume this is no longer needed. Can it be removed?

mducle:
> @mducle Is it possible to have two slices on the same plot? Does mslice not clear the figure each time a new one is plotted?

@aybamidele No. It should clear the slice figure each time a new slice is plotted so there should only be one slice at a time in the slice plot window.

aybamidele:
@martyngigg The problem @mducle  mentioned is caused by showEvent not being called the time the second plot is made.

aybamidele:
@mducle Please, could you confirm that the bugs you found are no longer present in the latest push to the branch.

martyngigg:
@aybamidele We just have 1 failing test and then we can merge this.

PR #436, opened 2019-02-18
Waterfall plots

### LINKED ISSUE #433: Waterfall plots for cuts.
An option to produce a "waterfall" plot - where in a plot with multiple 1D cuts each cut is offset in `y` and/or `x` from other cuts by a constant amount - was one of the original requirements of MSlice, but its priority was downgraded in favour of the implementation of the CLI.

The caching mechanism implemented for the CLI should, however, now make implementing the waterfall plots much easier.

mducle:
This PR adds a waterfall plot option to `Cut` figures.

**To test:**

1. Load a dataset and make several cuts - either using the `width` option or with `Plot Over`.
2. Click on the `Waterfall` button on the plot toolbar. You should see two line-edit boxes appear to the left of the waterfall button with zeros in them - this is the *x* and *y* offsets, in data coordinates
3. Put in some numbers in the edit boxes and check that the plotted curves are shifted by the appropriate amounts.
4. Click on the `Waterfall` button again. Check that the curves are back to their original positions (e.g. the plot is no longer a waterfall) and that the *x* and *y* offset line edit boxes have disappeared.
5. Plot another cut on the same figure (using `Plot Over`), and click on `Waterfall` again - you should see that not only the original cuts have been offset, but also the new cut. 
6. Plot yet another cut on the same figure and check that it too is appropriately displaced.
7. In the commandline (IPython console) run `gca().set_waterfall(False)` and check that the waterfall effect is turned off.
8. In the commandline run `gca().set_waterfall(True)` to turn it on again, or specify a different offset using e.g. `gca().set_waterfall(True, x_offset=1, y_offset=2)` or similar.
9. In the cut plot window, enable the waterfall plot again and then generate a script file of the plot. Look in the script file and check that there is a call to `set_waterfall` with the correct offsets. Also check that the generated file runs ok and produces the correct waterfall plot.
10. Go back to the main GUI and plot a slice. Click `Interactive Cut` on this slice and select an area of the Slice plot. Check that the original cut window has now changed to an interactive cut window (and that the `Waterfall` button is hidden). Click on `Interactive Cut` on the slice window to disable the interactive cut and check that the `Waterfall` button is visible again with its previous setting (e.g. if it was previously unchecked that it is still unchecked and if it was previously checked that the *x* and *y* offset boxes are now visible with their previous values).

Fixes #433


martyngigg:
@mducle We've got some conflicts after merging a couple of other PRs

mducle:
@martyngigg - Yes, doing the waterfall plot on a single cut should not shift it at all. The callback still runs but the offset will always be zero. I thought that it didn't make any sense to translate the first curve only subsequent curves.

martyngigg:
@mducle Thanks for the information. That make sense I think. Do you think we should disable the waterfall button if there is only a single curve so it is obvious it is not supposed to do anything?

mducle:
@martyngigg Yeah... that's an idea. I'll try to put that in the code.

mducle:
@martyngigg I've put code to disable the waterfall if there is only one curve. The previous state of the waterfall action is still remember though so if you plotted several curves on a particular plot window and enabled waterfall with a non-zero offset; then plot a single curve it will disable the waterfall button and hide the edit boxes for the single curve; but then if you overplot something else it will re-enable the waterfall button *and* offset the new curve by the amount previously specified.

martyngigg:
@mducle That sounds reasonable to me. Looks good now!

I just fixed the tiny flake8 issue that was there in that last commit.

mducle:
@martyngigg Ah thanks - that's weird, I had the `and` on the second line originally and ran flake8 on it and it complained about `W503 line break before binary operator` so I then put it on the first line...   ¯\_(ツ)_/¯

martyngigg:
Actually it looks like there are conflicting flake8 issues between python 2 and 3 because the changed :man_facepalming:. I've disabled the check globally.

mducle:
Thanks for sorting it out... I only tend to test on python2.

PR #442, opened 2019-02-28
Added energy units

### LINKED ISSUE #438: Energy units conversion
Both cuts and slices should have the option to convert the energy transfer axis from `meV` (default) to `cm` (wavenumber) at least (and perhaps other units such as `THz` and `K` also).

mducle:
This PR adds an energy units conversion framework to MSlice. Only two units are implemented: `meV` (`DeltaE` in Mantid notation, and the actual unit used in all calculations in MSlice) and `cm-1` (`DeltaE_inWavenumber` in Mantid notation).

A new class, `EnergyUnits` has been added to facilitate conversion to/from `meV` before each workspace operation that needs it. The `Axis` class has been modified to be aware of the different energy units and to call `EnergyUnits` for the necessary conversions.

New comboboxes on the `Slice` and `Cut` tabs have been added so users can specify the desired unit. If this is not `meV`, then the output workspaces generated by the `Slice` and `Cut` algorithms are changed (rescaled) to the new unit using either `ScaleX` (for `non-PSD` workspaces) or `TransformMD`. The Mantid `units` label is still left as `DeltaE` however, so to preserve knowledge of the actual energy units, the `Axis` objects used by `Slice` and `Cut` is included in the (MSlice) workspace objects as attributes. 

As the `algorithms` only work with pure Mantid workspaces these attributes would be lost on re-wrapping, so a mechanism to `pickle` these `Axis` attributes via the `comment` attribute of the Mantid workspace is used (TODO: change from using `comment` to using `log`).

The energy units combo box in the `Powder` tab has been removed - the conversion (using `MakeProjection`) to `PixelWorkspace` (`MDEventWorkspace` in Mantid terms) is now done only in `meV` - but `Slice` and `Cut` from the `PixelWorkspace` can convert the energy units to something else.

Finally a new menu option has been added which sets the default energy unit used by MSlice - this basically resets the energy units combo box to the default unit if it encounters a new workspace without saved cut parameters, so that users don't have to keep clicking on the combo box if they want to regularly use a unit that is not `meV`.

TODO: sort out file saving mechanisms for different energy units.

**To test:**

Code review.

Run MSlice and load some workspaces. Plot slices and cuts with `cm-1` as the unit and check they look equivalent to the cuts with `meV` - not that changing from `meV` to `cm-1` in the combo box should trigger a callback that will rescale the parameters in the `DeltaE` cut or slice parameter fields.

Check that when `cm-1` is selected as the default unit in the `Options` menu that when clicking on a new workspace or new cut axis that the energy units is set to `cm-1`.

Check that the overplot information (recoil and powder lines) still look ok in `cm-1`. 

Check that interactive cuts work ok in `cm-1`.

Check that script generation with plots in `cm-1` works and the scripts reproduce the plots.

More detailed test instructions to follow.

Fixes #438

### DIFF
--- a/mslice/app/mainwindow.py
+++ b/mslice/app/mainwindow.py
@@ -12,6 +12,8 @@ from mslice.widgets.workspacemanager import TAB_2D, TAB_EVENT, TAB_HISTO, TAB_NO
 from mslice.widgets.workspacemanager.command import Command as ws_command
 from mslice.widgets.cut.command import Command as cut_command
 
+from functools import partial
+
 TAB_SLICE = 1
 TAB_CUT = 2
 TAB_POWDER = 0
@@ -79,8 +81,8 @@ class MainWindow(MainView, QMainWindow):
         self.wgtPowder.busy.connect(self.show_busy)
         self.data_loading.busy.connect(self.show_busy)
         self.action_quit.triggered.connect(self.close)
-        self.actionmeV.triggered.connect(self.set_meV_default)
-        self.actioncm.triggered.connect(self.set_cm_default)
+        self.actionmeV.triggered.connect(partial(self.set_energy_default, self.actionmeV, self.actioncm))
+        self.actioncm.triggered.connect(partial(self.set_energy_default, self.actioncm, self.actionmeV))
 
     def setup_save(self):
         menu = QMenu()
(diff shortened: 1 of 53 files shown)


mducle:
@martyngigg 

1. Yes. I'll change the labels to both be `en`.
2. Thanks for reminding me. I had thought about this but then forgot about it...

mducle:
@martyngigg I've sorted out the issues you mentioned. MSlice workspace attributes (axes) are now saved to the Mantid raw workspace using a log entry rather than the comment, and overplotting a different energy unit will automatically rescale all previous cuts (the cuts are recomputed - so the saved cuts in the `MDHisto` tab are actually converted). This option can be disabled in the options menu.

Could you have a look at this, and then merge it to nightly, please?

martyngigg:
@mducle There seems to be an issue with the generating a script for the cuts, the slices are okay. If I plot a cut from the GUI and then try and run the generated script then I get
```python
  File "test-cut.py", line 5
    ws_MAR21335_Ei60_00meV_cut(-100_000,100_000) = mc.Load(Filename='/home/dmn58364/mslice/MAR21335_Ei60.00meV.nxs')
                                      ^
SyntaxError: invalid syntax
```
I guess it's an issue with the naming of the variables in the script.

mducle:
@martyngigg I think it's all done now...

PR #444, opened 2019-03-01
Cut plotter bugsfixes

### LINKED ISSUE #443: Cut plotter bugs
Some bugs in the cut functionalities were discovered during implementation of energy unit translation for PR #442:

1. In the interactive cuts, the buttons to `Save Cut to Workspace` and `Flip Integration Axis` do nothing (the console returns an `AttributeError` because the `get_icut` method returns `None` instead of an `InteractiveCut` instance.

2. Script generation functionality only really works for a single cut figure - e.g.:
   a. Load a workspace and make several over-plot cuts on a single figure. Save a script for this figure - it should work as expected. Now `Keep` that figure.
   b. Make several cuts on second figure. Go back to the first figure (with `Keep` set) and save another script for this figure from its file menu. Both the first and second scripts generated should be identical. However, whilst the `label` for this second file is correct, it uses the axes limits for curves in the *second* (`Current`) figure instead of the first. 

3. Restart MSlice and `import` / run the first generated script. Now try to generate another script from the resulting figure - the commands for workspace loading / make projection etc and the axes labels etc are there but the actual cut and plot (`errorbar`) commands are not.

All these bugs stem from the cache-mechanisms used by the `CutPlotterPresenter`.

mducle:
This is a small PR to fix some bugs identified during implementation of the energy units conversion framework - because the fixes here don't really fit into that work.

This PR fixes the first two issues mentioned in #443 (buttons doing nothing in the interactive cut window, and script generation getting things wrong if there are multiple cut windows and the user switches back and forth between them). The third issue cannot be addressed easily but the machinery for it is included in PR #442 (inclusion of MSlice `Axis` instances within `Workspaces` and using workspace comments to propagate them through Mantid algorithms). This is because the CLI does not really use the `CutPlotterPresenter` which is used by the GUI to cache cut information. So calls to cache cut information from the CLI must be in the `errorbar` function itself.

**To test:**

1. Load a workspace, make a slice from it and select `Interactive Cut`. In the interactive cut window (with the 1D line plot) click on either of the `Save workspace` or `Flip axis` buttons and check that they work as advertise (e.g. a workspace is created in the `MDHisto` tab and the integration axis is changed).

2. Make a series of cuts using `Plot Over` then `Keep` that figure. Make another series of cuts and go back to the kept (first) figure. Click on its `File` menu and select `Generate Script`. Check that the generated script is correct for this figure window. Go the the second window (the `Current` one) and generate a script for that one and also check that it's ok.

Fixes #443 .

### DIFF
--- a/mslice/cli/helperfunctions.py
+++ b/mslice/cli/helperfunctions.py
@@ -40,7 +40,7 @@ _intensity_to_workspace = {
 
 def _update_cache(cut_presenter, ws_name, CutAxis, IntegrationAxis, NormToOne):
     """Creates a list of all cuts used to create a particular cut. This is required when plot over is used."""
-    cut_list = cut_presenter._cut_cache_list
+    cut_list = cut_presenter._cut_cache_dict
     cut_cache = cut_presenter._cut_cache
     int_axis = Axis(*IntegrationAxis.split(','))
     cut_axis = Axis(*CutAxis.split(','))
--- a/mslice/cli/helperfunctions.py
+++ b/mslice/cli/helperfunctions.py
@@ -40,29 +40,13 @@ _intensity_to_workspace = {
 
 def _update_cache(cut_presenter, ws_name, CutAxis, IntegrationAxis, NormToOne):
     """Creates a list of all cuts used to create a particular cut. This is required when plot over is used."""
-    cut_list = cut_presenter._cut_cache_dict
-    cut_cache = cut_presenter._cut_cache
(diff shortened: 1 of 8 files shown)


mducle:
@martyngigg - no sorry it's a bug. I decided to rename the variable from `_cut_cache_list` to `_cut_cache_dict` and just missed this instance... I think for some reason I didn't use the refactoring tool but decided to do it manually :(

PR #448, opened 2019-03-05
Updated docs and version to 1.2

mducle:
Updates the documentation with information about the CLI and energy units conversion.

Also updates the version information to `1.2`

**To test:**

Read through and see if the docs make sense!

No related issue.


mducle:
@martyngigg Ah yes... I forgot about the index page. I've updated it now.

PR #453, opened 2019-03-11
Fixed bug to allow plotting indirect nxspe files

### LINKED ISSUE #451: Unable to plot IRIS/OSIRIS nxspe files
The indirect spectrometers IRIS and OSIRIS can save reduced datafiles into both `nxs` and `nxspe` formats. MSlice is able to read both formats but only plots cuts and slices from the `nxs` files. That is, for the `nxspe` file, the `Slice` and `Cut` are enabled (it recognises the data correctly as `non-PSD`) but all the edit fields and buttons in the `Slice`/`Cut` tabs are disabled.

This is because the `nxspe` format does not save the fixed final energy (`Ef`) of each detector element in the file. MSlice used to work around this by asking the user for a single `Ef` value which it uses for all detectors (which is the case for IRIS and OSIRIS but not necessarily true for other indirect geometry spectrometers). The popup window asking the user for `Ef` still comes up but somewhere in the refactoring of the code this information has not be propagated through, so that other parts of MSlice still think that `Ef` is not defined. 

Test files here: [iris_reduced_data.zip](https://github.com/mantidproject/mslice/files/2945457/iris_reduced_data.zip) - they are both of the same data from the same reduction, but one is in `nxs` and the other in `nxspe` format.

mducle:
Fixes bug in loading an indirect instrument `nxspe` file which prevented it from plotting

**To test:**

Load the nxspe file in [iris_reduced_data.zip](https://github.com/mantidproject/mslice/files/2945457/iris_reduced_data.zip) and check that you can plot slices and cuts from it.  

Fixes #451

### DIFF
--- a/mslice/models/cut/cut_algorithm.py
+++ b/mslice/models/cut/cut_algorithm.py
@@ -2,7 +2,7 @@
 from mantid.api import PythonAlgorithm, WorkspaceProperty
 from mantid.kernel import Direction, PropertyManagerProperty, StringMandatoryValidator
 from mantid.simpleapi import BinMD, ConvertSpectrumAxis, CreateMDHistoWorkspace, Rebin2D, SofQW3, TransformMD, \
-    ConvertToMD, DeleteWorkspace
+    ConvertToMD, DeleteWorkspace, CreateSimulationWorkspace, AddSampleLog, CopyLogs
 
 from mslice.models.alg_workspace_ops import fill_in_missing_input, get_number_of_steps
 from mslice.models.axis import Axis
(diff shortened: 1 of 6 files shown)


mducle:
@martyngigg Sorry... I forgot about cuts altogether on this... It's fixed now. There is also a very small fix to the plotting Bragg lines from CIF files in here too... hope that's ok...

mslice/models/cut/cut_algorithm.py
```diff
@@ -90,10 +90,15 @@ def _compute_cut_nonPSD(selected_workspace, cut_axis, integration_axis, emode):
     extents = " ,".join(map(str, (xdim.getMinimum(), xdim.getMaximum())))
 
     # Hack to (deep) copy log data (ExperimentInfo)
-    _tmpws = ConvertToMD(ws_out, EnableLogging=False, StoreInADS=False, PreprocDetectorsWS='-')
```
martyngigg: Could you help me understand what is happening here? I'm struggling a little..

mducle: Yeah... it's a bit over complicated. The main issue is to be able to copy the sample logs from `ws_out` which is a `Workspace2D` to the final output which is an `MDHistoWorkspace`. The cleanest solution would be to use the `CopyLogs` algorithm and use that but at the moment `CopyLogs` does not work on MD workspaces. So, instead I'm using the MDWorkspace's `copyExperimentInfos` method but this requires another MDWorkspace. This is created using `ConvertToMD` but for `dEAnalysisMode='Indirect'`, `ConvertToMD` requires each detector to have the fixed `E_f` defined. This is the case for `nxs` files but not for `nxspe` files. (`SofQW` allows you to specify a value for `E_f` to override that in the detectors / or in case the values in the detectors don't exist, but `ConvertToMD` doesn't).

So the main solutions are:

1. Use this complicated work-around / hack which creates a blank `MatrixWorkspace`, copies the logs, converts it to MD and then uses `copyExperimentInfos`. The only advantage of this is that it is pure python and doesn't need changes to the C++ code (and would allow MSlice master to be backwards compatible with older Mantid versions).
2. Change `CopyLogs` to allow it to copy logs from `MatrixWorkspace` to MD workspaces.
3. Change `ConvertToMD` to allow specifying a value for `E_f`
4. Change the `copyExperimentInfos` method to allow it to accept a `MatrixWorkspace` too.

If we don't like the current implementation, I think changing `CopyLogs` is the best option. It would also allow some other python code in Mantid (mostly in various workflow algorithms) to be refactored.

martyngigg: I see. Thanks for the explanation. I'm happy keeping this as a solution for the release but I agree I think the better long-term solution is to update `mantid`  probabaly with all 3 of your final points above.
Could you add this comment to that code so we don't forget what is a happening?

PR #455, opened 2019-03-12
Fix 2theta units naming bug

### LINKED ISSUE #454: Incorrect `2theta` slices / cuts
* Make a slice of the standard file `MAR21335_Ei60.00meV` with `DeltaE` as the `x`-axis and `2Theta` as the `y`-axis - everything appears to be ok (the plot appears and shows the correct data, with Bragg peaks (bright spots) around 0meV.
* Now make the same slice but with `2Theta` as the `x`-axis and `DeltaE` as the `y`-axis - the plot appears but with the colormap in the top left corner and with the incorrect data (e.g. no Bragg peaks around 0meV).
* Make cut of the same data along `2Theta` from 0 to 135 degrees and integrating from -5 to +5 meV in `DeltaE` - you should see a series of spikes corresponding to the Bragg peaks at different angles. However, what appears looks more like a cut along energy (peaks at 0 and ends at around 60).

mducle:
This fixes a bug where the label for a 2-theta cut is `2Theta` rather than the expected `Degrees`. It seems that `MDWorkspaces` and `MatrixWorkspace` have different conventions: `MatrixWorkspace` expects `Degrees` which was what MSlice had but `MDWorkspace`s are more flexible and in the GUI in MSlice we call that axis `2Theta` which is then passed to the different algorithms which do not recognise this (they expect `Degrees`). 

This PR updates the relevant checks in the `Cut` and `Slice` algorithms so that they will now accept both `Degrees` and `2Theta`.

**To test:**

Repeat the cuts / slices detailed in issue #454 and check that they are correctly plotted now (e.g. `DeltaE` vs `2Theta` and `2Theta` vs `DeltaE` slices are consistent and cuts along `2Theta` work as described. Also check that overplotting lines work on the slices.

Fixes #454 .

### DIFF
--- a/mslice/models/cut/cut_algorithm.py
+++ b/mslice/models/cut/cut_algorithm.py
@@ -77,7 +77,7 @@ def _compute_cut_nonPSD(selected_workspace, cut_axis, integration_axis, emode):
         idx = 1
         unit = 'MomentumTransfer'
         name = '|Q|'
-    elif cut_axis.units == 'Degrees':
+    elif cut_axis.units == 'Degrees' or cut_axis.units == '2Theta':
         ws_out = _cut_nonPSD_theta(cut_binning, int_binning, selected_workspace)
         idx = 1
         unit = 'Degrees'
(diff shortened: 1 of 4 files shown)


mducle:
@martyngigg I've refactored this to use `labels.py`. I've defined a couple of helper functions though so that it will search substrings... I don't think this is needed but some of the code do that instead of checking for exact equality.

PR #469, opened 2019-04-30
Migration from QtAwesome to get_icon

Pasarus:
Description of work.
Mantid is moving away from using QtAwesome, so the reliance in msclice will no longer be possible. QtAwesome was replaced by a new mantid library dedicated to icons, for both C++ and Python. This has been used in it's place. The icons are very close if not identical to the previous icons.

**To test:**
- Open Mslice
- Load some data
- Get the data to the state in which a plot window can be opened
- Check out the new icons, including the interactive cut circular arrows (retweet) icon.

### DIFF
--- a/mslice/plotting/plot_window/plot_window.py
+++ b/mslice/plotting/plot_window/plot_window.py
@@ -1,7 +1,7 @@
 from __future__ import (absolute_import, division, print_function)
 
 from matplotlib.figure import Figure
-import qtawesome as qta
+from mantidqt.icons import get_icon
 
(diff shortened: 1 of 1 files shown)


martyngigg:
We have a chicken and egg problem with the builds because the nightly build is used to test this and that doesn't have the icon changes yet.

I have run all of the tests locally and they passed. I have also imported the new version into MantidPlot and Workbench and all seems fine. I am therefore going to merge this despite the failing tests to we can get the fixed version into mantid.

PR #476, opened 2019-07-03
SaveToADS button now enabled in workbench

### LINKED ISSUE #473: Save to ADS should also be enabled for Workbench
In the `MD Histo` tab there is a button to export 1D cuts to the ADS (after conversion from `MDHistoWorkspace` to `MatrixWorkspace`) for fitting. It is only enabled and visible when MSlice is run from within MantidPlot since it doesn't make sense for it to be visible if run from the command line as it would appear to do nothing.

However, it is now also not enabled/visible in the Workbench - which it should be.

Report to: G. Granroth (SNS)

mducle:
Changes the check for `in_mantidplot` to include the workbench.

**To test:**

Run MSlice in the workbench - go to the `MD Histo` tab and check the `Save to Mantidplot` button is visible.

Run MSlice in Mantidplot -  go to the `MD Histo` tab and check the `Save to Mantidplot` button is visible.

Run MSlice using the `mslicedevel.sh` script outside of Mantidplot and workbench. Go to the `MD Histo` tab and check the `Save to Mantidplot` button is *not* visible.

Fixes #473


mducle:
@martyngigg I've changed it so it checks whether it is `MantidPlot` or `Workbench` and sets the button text accordingly.

PR #477, opened 2019-07-04
Fix cut saving bugs

### LINKED ISSUE #472: Bug: saving cut only works for first cut plotted to that window
Reported by a user in the Forum: https://forum.mantidproject.org/t/saving-data-from-mslice/541

Report to: MJCliff on Forum

To reproduce:

* Load a data file.
* Plot a cut
* Save it from the save button on the plot window
* Plot another cut (maybe with different integration range)
* Click the save button and give it a different filename
* Compare the two saved files - they are the same: should be different cuts.

Alternatively:

* Load a data file
* Plot a cut with a non-blank `Width` parameter (multiple cuts should appear in the plot window)
* Click the save button
* Save generates only a saved file of the first cut - should have generated a file for each cut.

mducle:
Fixes a but where only the first cut plotted on a window is saved.

**To test:**

Load data and make a cut. Click on the save icon in the cut window and save to a file. Make another cut, plot it (not plot over) and save that to a file also. Before this fix the two files are the same - and show the data from the first cut made. Now they should be different.

Make multiple (plot over) cuts on a single plot window. Click save. If you select a data file format (e.g. `ascii`) it will save multiple files for each cut, appending a serial number to the files to distinguish them.

Fixes #472

### DIFF
--- a/mslice/models/workspacemanager/file_io.py
+++ b/mslice/models/workspacemanager/file_io.py
@@ -38,7 +38,8 @@ def get_save_directory(multiple_files=False, save_as_image=False, default_ext=No
             file_dialog.selectNameFilter(ext_to_qtfilter[default_ext])
         if (file_dialog.exec_()):
             path = str(file_dialog.selectedFiles()[0])
-            if '.' not in path: # add extension unless there's one in the name
+            filename = os.path.basename(path)
+            if '.' not in filename: # add extension unless there's one in the name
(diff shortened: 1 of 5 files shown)


mducle:
@martyngigg I was going to write unit tests for this but it seems we need unit tests for everything in `plot_figure_manager.py` so I'll defer that to later...

PR #495, opened 2019-07-12
Keep dialog open

### LINKED ISSUE #460: Keep dialog open option
GUI dialogs like for `x`-axis, `y`-axis, `color`-axis etc should have a `Keep open` check box (like the algorithm dialogs in MantidPlot).

Report to: ISIS / T. Guidi

mducle:
Changes the `QuickAxisOptions` and `PlotOptions` classes to be non-modal and to have a `Keep open` check box. If this is not checked the `Ok` button will close the dialog. If it is checked then `Ok` will submit the new values entered to the plot but the dialog will stay open. This is done by connecting the options dialog to both the update function in the presenter and the `_canvas.draw()` function of the plot window. When the `Ok` button is clicked, a Qt `Signal` is emitted to the presenter to update the model, and then another signal is emitted to the canvas to redraw the graph. (I think there is no guarantee that these signals will be processed one after the other so there may be a race condition but in testing this didn't not seem to be a problem and I did not want to add a time delay between them).

**To test:**

Make a plot (slice or cut). Double click on one of the axes - the quick options dialog will pop up. Check the `Keep open` box and type a new value for the limits. Hit `enter` or click `Ok` and check that the plot updates but the dialog stays open. Uncheck the `Keep open` box, and change the value again, and hit `enter` / click `Ok`. Check that the plot updates correctly and the dialog is closed. Do the same for the main plot options (click on the cogwheel icon).

Fixes #460

### DIFF
--- a/mslice/presenters/quick_options_presenter.py
+++ b/mslice/presenters/quick_options_presenter.py
@@ -8,7 +8,7 @@ def quick_options(target, model, has_logarithmic=None, redraw_signal=None):
     if isinstance(target, text.Text):
         quick_label_options(target)
     elif isinstance(target, string_types):
-        quick_axis_options(target, model, has_logarithmic, redraw_signal)
+        return quick_axis_options(target, model, has_logarithmic, redraw_signal)
     else:
         quick_line_options(target, model)
 
@@ -26,6 +26,7 @@ def quick_axis_options(target, model, has_logarithmic=None, redraw_signal=None):
     view = QuickAxisOptions(target, getattr(model, target), grid, has_logarithmic, redraw_signal)
     view.ok_clicked.connect(lambda: _set_axis_options(view, target, model, has_logarithmic, grid))
     view.show()
+    return view
 
 
 def quick_line_options(target, model):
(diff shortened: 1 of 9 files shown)


mslice/presenters/quick_options_presenter.py
```diff
@@ -39,6 +41,7 @@ def _run_quick_options(view, update_model_function, *args):
 
 
 def _set_axis_options(view, target, model, has_logarithmic, grid):
+    print('setting axis options')
```
martyngigg: Is this a left over debugging statement?

mducle: Yes. Doh! I thought I removed it but I guess I forgot to do `git add` before `git commit --amend`...

mducle:
> 
> 
> This works well. Do we also want to include a keep open box for the quick-axis-label editor box?

I didn't think it was necessary - the axes options is because if you first don't get the right level you want to modify it again without double-clicking on the axes. Maybe we can change it after getting some feedback from users.

mslice/presenters/quick_options_presenter.py
mducle: The print statement was because I couldn't seem to get PyCharm to stop at break points when running the tests through `nose` - not sure why...

Issue #504, opened 2019-07-15
Running script created from waterfall plot crashes MantidPlot / Workbench


mducle:
### To reproduce

Run MSlice within MantidPlot / Workbench. Load a dataset and create a plot with multiple cuts. Click on `Waterfall` and set the x- and y- offsets to be non-zero. Save the script for this plot using the menu option File->Generate Script File. Open the script in the Python window and run it. It will cause MantidPlot / Workbench to freeze and then crash (the unhandled exception / sad mantid window will not pop up). 

### Analysis

Running the script outside MantidPlot / Workbench works fine.

Also running the script with the `ax.set_waterfall` line commented out does not crash MantidPlot / Workbench.

The waterfall feature works by getting all the Matplotlib `Line2D` elements in a plot and then applying an offset to them, so it could be an issue with Matplotlib within MantidPlot / Workbench somehow.

mducle:
The issue seems to be in the handling of the depressed `Waterfall` button - running it through `gdb` shows that it crashes in `QFontEngineFT::loadGlyphSet(QTransform const&) () from /lib64/libQt5XcbQpa.so.5` when trying to update the text for the edit boxes which are enabled by the `Waterfall` button. When running directly from the GUI this works fine but from the CLI it crashes - probably something to do the being in different threads (just before this the system also says `QCoreApplication::exec: The event loop is already running`). 

A quick work-around for the release is to disable the waterfall button for CLI generated graphs (but to still plot the curves with the waterfall offsets).

PR #508, opened 2019-07-16
Workaround for CLI waterfall crash

mducle:
This is a quick work-around to avoid a hard crash of the Workbench when a script generated from a waterfall plot is run in the Workbench script window. 

As explained in the issue #504 - the crash appears to be in some Qt threading issue - it only occurs when the state of the Qt window is changed from within the script in the MSlice CLI - it occurs when the `Waterfall` button (which is a checkable button in a `QToolbar`) is enabled which also triggers two `QLabels` and two `QLineEdits` to be made visible. Even disabling making these visible also triggers the crash so it might be the "depressed" `Waterfall` state which is responsible. In any case `gdb` traces the segfault to a function in `QFontEngineFT`. 

The workaround is to make the `Waterfall` button invisible so these triggers have no effect, and thus avoids the crash. A longer term fix will be made for the next release.

**To test:**

Run MSlice within the Workbench. Make multiple cuts on the same graph and enable a waterfall with some non-zero offset. Make a script of the window and load it in the workbench script window. Run the script - it should reproduce the plot with the waterfall offset but the `Waterfall` button will be invisible. 

Run the script outside the workbench - check that the `Waterfall` button still works as before.

### DIFF
--- a/mslice/cli/__init__.py
+++ b/mslice/cli/__init__.py
@@ -12,11 +12,13 @@ from mslice.cli.helperfunctions import (_check_workspace_name, _check_workspace_
 from mslice.models.workspacemanager.workspace_provider import get_workspace_handle
 from mslice.plotting.globalfiguremanager import GlobalFigureManager
 from mslice.workspace.histogram_workspace import HistogramWorkspace
+from mslice.app import in_mantid as is_in_mantid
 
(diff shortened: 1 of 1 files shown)


martyngigg:
Closed in favour of #509

PR #541, opened 2020-08-27
Force figuremanager calls to be on same thread as QApplication object

### LINKED ISSUE #535: import mslice.cli crashes Mantid sometimes
Sometimes (but not always) running `import mslice.cli` in the *script window* (not the IPython console) in either MantidPlot or MantidWorkbench causes a hard crash of Mantid.

martyngigg:
Description of work.

This should be a more robust method for ensuring future updates of matplotlib don't introduce further calls that need to be forced onto the GUI thread when run from mantid's script editor.

**To test:**

Run any mslice script in the workbench script editor.

Fixes #535.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,6 @@ ENV/
 # Vim files
 *.swp
 *.swo
+
+# vscode files
+.vscode/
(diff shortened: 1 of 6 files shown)


martyngigg:
Merging after mantidproject/mantid#29242 verified it's good.

PR #546, opened 2020-09-17
Prevent crashes on cancel for rename- and add-workspace dialogs

### LINKED ISSUE #542: MSlice Unhandled ValueError exception when rename cancelled
**Original reporter:** David Voneshen via Error Reporter

### Expected behavior
When renaming a file in the MSlice interface, clicking cancel should close the dialog box and not cause an error

### Actual behavior
An error is thrown when the rename dialog box is closed

### Steps to reproduce the behavior
1. Open the MSlice interface (Direct -> MSlice)
2. Load a file (e.g. 164198.nxs from the training data)
3. Click rename to open the dialog, then click cancel

### Platforms affected
Confirmed on windows, assumed on all

joseph-torsney:
Fixes crashes associated with the rename workpsace and add workspace dialogs when the user clicked cancel. This was fixed by raising `RuntimeErrors` when the dialogs were cancelled, which are caught by the `workspace_manager_presenter`.

**To test:**
1. Load a workspace, e.g. `164198.nxs` in `TrainingCourseData`
2. In the Workspace Manager tab, click add. 
3. Cancel the dialog, check MSlice doesn't crash.
4. Repeat for rename workspace.

Fixes #542

### DIFF
--- a/mslice/presenters/workspace_manager_presenter.py
+++ b/mslice/presenters/workspace_manager_presenter.py
@@ -157,8 +157,15 @@ class WorkspaceManagerPresenter(WorkspaceManagerPresenterInterface):
         if not selected_ws:
             self._workspace_manager_view.error_select_one_or_more_workspaces()
             return
-        if len(selected_ws) == 1:
-            selected_ws.append(self._workspace_manager_view.add_workspace_dialog())
+        if len(selected_ws) != 1:
+            return
+        try:
+            new_ws = self._workspace_manager_view.add_workspace_dialog()
+        except RuntimeError as e:
+            if str(e) == 'dialog cancelled':
+                return
+            raise RuntimeError(e)
+        selected_ws.append(new_ws)
         try:
             add_workspace_runs(selected_ws)
         except ValueError as e:
--- a/mslice/presenters/workspace_manager_presenter.py
+++ b/mslice/presenters/workspace_manager_presenter.py
@@ -127,7 +127,12 @@ class WorkspaceManagerPresenter(WorkspaceManagerPresenterInterface):
             self._workspace_manager_view.error_select_only_one_workspace()
             return
         selected_workspace = selected_workspaces[0]
-        new_name = self._workspace_manager_view.get_workspace_new_name()
+        try:
+            new_name = self._workspace_manager_view.get_workspace_new_name()
+        except RuntimeError as e:
+            if str(e) == 'dialog cancelled':
+                return
+            raise RuntimeError(e)
         rename_workspace(selected_workspace, new_name)
         self.update_displayed_workspaces()
 
--- a/mslice/presenters/workspace_manager_presenter.py
+++ b/mslice/presenters/workspace_manager_presenter.py
@@ -127,12 +127,9 @@ class WorkspaceManagerPresenter(WorkspaceManagerPresenterInterface):
             self._workspace_manager_view.error_select_only_one_workspace()
             return
         selected_workspace = selected_workspaces[0]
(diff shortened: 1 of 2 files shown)


mslice/widgets/workspacemanager/workspacemanager.py
```diff
@@ -118,7 +118,8 @@ def add_workspace_dialog(self):
         dialog.setLabelText("Choose a workspace to add:")
         dialog.setOptions(QInputDialog.UseListViewForComboBoxItems)
         dialog.setComboBoxItems(items)
-        dialog.exec_()
+        if dialog.exec_() != QInputDialog.Accepted:
+            raise RuntimeError('dialog cancelled')
```
robertapplin, line 122: Would it be slightly better to `return None` instead of raising an error, because cancelling a dialog isn't a runtime error?

mslice/widgets/workspacemanager/workspacemanager.py
```diff
@@ -166,10 +167,10 @@ def get_workspace_to_load_path(self):
         return paths[0] if isinstance(paths, tuple) else [str(filename) for filename in paths]
 
     def get_workspace_new_name(self):
-        name, success = QInputDialog.getText(self,"Workspace New Name","Enter the new name for the workspace :      ")
+        name, success = QInputDialog.getText(self, "Workspace New Name", "Enter the new name for the workspace :      ")
         # The message above was padded with spaces to allow the whole title to show up
         if not success:
-            raise ValueError('No Valid Name supplied')
+            raise RuntimeError('dialog cancelled')
```
robertapplin, line 173: Same here, `return None` might be better

mslice/presenters/workspace_manager_presenter.py
```diff
@@ -127,7 +127,12 @@ def _rename_workspace(self):
             self._workspace_manager_view.error_select_only_one_workspace()
             return
         selected_workspace = selected_workspaces[0]
-        new_name = self._workspace_manager_view.get_workspace_new_name()
+        try:
+            new_name = self._workspace_manager_view.get_workspace_new_name()
```
robertapplin, line 131: The `try-except` statements could then be replaced by a check to see if `new_name` is `None`, and if it is then return.

mslice/presenters/workspace_manager_presenter.py
```diff
@@ -152,8 +157,15 @@ def _add_workspaces(self):
         if not selected_ws:
             self._workspace_manager_view.error_select_one_or_more_workspaces()
             return
-        if len(selected_ws) == 1:
-            selected_ws.append(self._workspace_manager_view.add_workspace_dialog())
+        if len(selected_ws) != 1:
+            return
+        try:
+            new_ws = self._workspace_manager_view.add_workspace_dialog()
```
robertapplin, line 163: Same here as above ^

robertapplin (changes requested):
The bug is no longer present - I have made a few code change suggestions

mslice/widgets/workspacemanager/workspacemanager.py
joseph-torsney: I agree! I thought about this vs. following the style that was already used in the file, but returns are just nicer

robertapplin (approved):
Good job, this fixes the issue in #542 👍

martyngigg:
Thanks @robertapplin @joseph-torsney for fixing the recent issues. @robertapplin Could you create the PR on `mantid` to pull in the SHA from the current value of `master` as we discussed?

PR #552, opened 2020-10-21
Prevent MSlice crash by stopping the entry of comma's into QLineEdits expecting double's

### LINKED ISSUE #551: Crash when entering a comma in the Scale box
MSlice crashes when accidentally entering a comma into the `Scale` QDialog box.

This should be prevented.

https://mantid.slack.com/archives/CPXMTE49L/p1603286835025800

robertapplin:
**Description of work:**
Fixes the possibility of a crash by restricting the input of the line edits used for entering doubles. 

By default, a `QDoubleValidator` allows the entry of a comma. This will cause a crash when attempting to convert the entered data to a float. This PR prevents the entry of comma's into the relevant line edits thereby preventing a crash.

**To test:**
1. Open MSlice.
2. Load any data (e.g. `MAR25087`)
3. Go to the `Workspace Manager` tab
4. Click the down arrow next to `Compose` and select `Scale`
5. Attempt to enter a comma into the line edit. This should not be allowed.
6. Also try to enter a comma into the line edits seen on the `Slice` and `Cut` tabs. This should not be allowed.

Fixes #551.

### DIFF
--- a/mslice/plotting/plot_window/plot_window.py
+++ b/mslice/plotting/plot_window/plot_window.py
@@ -4,7 +4,8 @@ from matplotlib.figure import Figure
 from mantidqt.icons import get_icon
 
 from mslice.plotting.backend import get_canvas_and_toolbar_cls
-from mslice.util.qt import QtCore, QtWidgets, QtGui
+from mslice.util.qt import QtCore, QtWidgets
+from mslice.util.qt.validator_helper import double_validator_without_separator
 
 FigureCanvas, NavigationToolbar2QT = get_canvas_and_toolbar_cls()
 
@@ -158,7 +159,7 @@ class PlotWindow(QtWidgets.QMainWindow):
         sz.setWidth(60)
         self.waterfall_x_edt.setMaximumSize(sz)
         self.waterfall_y_edt.setMaximumSize(sz)
-        validator = QtGui.QDoubleValidator(self)
+        validator = double_validator_without_separator()
         self.waterfall_x_edt.setValidator(validator)
         self.waterfall_y_edt.setValidator(validator)
         self.waterfall_x_lbl_act = parent.addWidget(self.waterfall_x_lbl)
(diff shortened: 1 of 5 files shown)


mducle (changes requested):
Looks good to me. Now, you _cannot_ type `,` in the edit boxes - if you type `,` it ignores it so if you type `0,5` what appears is `05`. 

The root cause of this is that in some locales (e.g. Italy, Germany, France, etc) `,` is used as the radix separator, so a half is `0,5` not `0.5`; and a thousand and a half is `1.000,5` not `1,000.5`. It is possible to use the `locale` module to interpret strings based on locale in this way, but since Python float literals always *have* to use `.` as the radix separator, it is better to force users to use `.` in the text boxes too (unless we get push back later... - if we do, the instead of `float(v)` we can use `import locale; locale.atof(v)` which will depend on locale). 

However, we can give users a hint that they should use `.` by giving a default value for the scale of `1.0` (as a string). E.g.

```diff
diff --git a/mslice/widgets/workspacemanager/input_boxes.py b/mslice/widgets/workspacemanager/input_boxes.py
index 519336e..84b602a 100644
--- a/mslice/widgets/workspacemanager/input_boxes.py
+++ b/mslice/widgets/workspacemanager/input_boxes.py
@@ -41,6 +41,7 @@ class ScaleInputBox(QDialog):
             self.text1.setText('Scale Factor')
             layout.addRow(self.text1)
             layout.addRow(self.edit1)
+            self.edit1.setText('1.0')
             self.text2.hide()
             self.edit2.hide()
         okbtn = QPushButton('OK')
```

martyngigg:
Is this critical enough to consider putting into the upcoming patch?

mducle (approved):
Thanks @robertapplin . I approve this PR.

@martyngigg I'm not sure that it's "critical" (not very many users have reported this bug) but it does prevent a soft crash (sad Mantid screen) so if it's not too much trouble I would like to include it in the patch.

robertapplin:
> Is this critical enough to consider putting into the upcoming patch?

I think it might be worth putting it in the patch because the crash is easy to cause. I also thing the changes are fairly low-risk.

martyngigg:
@mducle That should be fine. The patch will most likely go out either tomorrow or Monday now.

Issue #556, opened 2021-01-04
Crash when entering a negative value or a standalone 'e' in width box for Cut.


SilkeSchomann:
**Expected behavior**
The width box for Cut should not accept a negative value or a standalone 'e' (as opposed to a double in e notation).

**Actual behavior**
MSlice crashes when either a negative value or a standalone 'e' is entered into the width box for Cut.

**Steps to reproduce the behavior**
1. Open MSlice.
2. Load any data 
3. Go to the Workspace Manager tab
4. Click on the Cut tab
5. Attempt to enter a negative value into the width box. (This should not be allowed at the moment it causes a crash.)
6. Also try to enter a standalone 'e' into the width box. (This should not be allowed at the moment it causes a crash.)

PR #564, opened 2021-01-06
Fixed MSlice crash when entering unexpected values into width box

SilkeSchomann:
Modified width function to not accept negative values or a standalone 'e'. Double values in proper e notation are still allowed.

**To test:**
1. Open MSlice.
2. Load any data
4. Click on the Cut tab
5. Attempt to enter a negative value into the width box. 
6. Also try to enter a standalone 'e' into the width box.

Fixes [#556](https://github.com/mantidproject/mslice/issues/556)

### DIFF
--- a/mslice/models/cut/cut.py
+++ b/mslice/models/cut/cut.py
@@ -91,7 +91,7 @@ class Cut(object):
 
     @width.setter
     def width(self, width_str):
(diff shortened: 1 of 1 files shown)


robertapplin (approved):
There is no longer any crashes, good job! I'm not sure if its worth adding a test for this, so I'm happy to approve

PR #565, opened 2021-01-07
Implement floating/ontop window behaviour for mslice

joseph-torsney:
Implements floating/ontop window behaviour (that can be set in workbench settings) on mslice, as described in [this PR on the mantid repo](https://github.com/mantidproject/mantid/pull/30129).

**To test:**
- Launch mslice and check it opens correctly, and that without mantid the window is 'floating'.
- If possible, test the behaviour in workbench. "Floating" means that mslice would disappear behind the workbench window when the user changes the window in focus. "On top" means mslice would stay on top of workbench.

### DIFF
--- a/mslice/app/__init__.py
+++ b/mslice/app/__init__.py
@@ -1,6 +1,8 @@
 """Package defining top-level MSlice application
 and entry points.
 """
+import sys
+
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
Works fine both on Windows and Linux.

PR #567, opened 2021-01-08
Changed buildscript to QT5

SilkeSchomann:


martyngigg:
Confirmed in logs that Qt4 errors are gone but an api change in mantid will be dealt with separately.

PR #568, opened 2021-01-08
Changed API call from getNumberBins to blocksize

SilkeSchomann:


martyngigg (approved):
This looks good and fixes the API breakage.

Issue #559, opened 2021-01-05
Adding several workspaces does not work as described in the documentation


SilkeSchomann:
1.8. Manipulating Workspaces describes selecting several workspaces and clicking on the Add button in order to sum the selected workspaces. In reality, it is only possible to select one workspace. After clicking on the Add button a second workspace can be selected in the Add Workspace dialog.

For subtracting it is possible to select more than one workspace though.

PR #569, opened 2021-01-08
Allow selection of more than one workspace for addition in workspace panel of MSlice

SilkeSchomann:
Fixed bug that only one workspace can be selected from the workspace panel so the Add Workspace dialog had to be used to select additional workspaces. Now the Add Workspace dialog only opens when one workspace was selected from the workspace panel. In case of a selection of more than one workspace clicking the Add button is enough to add the selected workspaces.

**To test:**
1. Open MSlice.
2. Load any data
4. Select more than one workspace
5. Click on Add
6. Check that new workspace with suffix '_sum' is created without opening the Add Workspace dialog

Fixes [#559](https://github.com/mantidproject/mslice/issues/559)


robertapplin (approved):
Works as described and changes look good 👍

PR #570, opened 2021-01-12
Fix for colorcycle problem for multiple curves on one plot

SilkeSchomann:
Adding curves via Python script with colors that are not in the pre-defined color list caused an unhandled exception.
Now these additional colors are temporarily added to the color list used in settings.

**To test:**
Follow instructions in issue description.

Fixes [#550](https://github.com/mantidproject/mslice/issues/550)


DanielMurphy22 (approved):
Works really well! Nice work!

Issue #528, opened 2020-04-07
Minor issues in Workbench integration


mducle:
A few issues have arisen when using MSlice in Workbench:

1. `RuntimeError` from deleting `Workspace`s which have already been deleted - seemed to have cropped up more often in Mantid 5.0. The MSlice `Workspace` wrappers have `__del__` methods which tries to delete MSlice-temporary workspaces (which are actually in the ADS but with hidden names beginning with `__`). However, sometimes Mantid sees that these have gone out of scope and deletes them before the Python `__del__` method is called, giving a `RuntimeError`.

2. `IndexError` when plotting the first figure if it is a `Cut` (rather than a `Slice`), because the options handler tries to determine the colorbar label when no colorbar exists. This error was also in MantidPlot, but in Workbench it triggers the sad-face unhandled exceptions dialog whilst in MantidPlot it did not.

mducle:
For issue 2, the problem is with using `mslice.cli.Show()` (which has to access the GlobalFigureManager to try to find the current figure). If using `fig.show()` where `fig = mslice.plotting.pyplot.figure()` then everything is ok.

mducle:
Issue 2 no longer exists in Mantid 5.1.1.

Issue #544 is a duplicate of issue 1 here.

PR #571, opened 2021-01-13
Fixed runtime error when trying to delete workspace for the second time

SilkeSchomann:
Occasionally temporary workspaces are already deleted when the clean up function is called. If not caught this causes runtime errors.
Now there is a check if the respective workspace is still present.

**To test:**

1. Open MSlice.
2. Load any data
4. Select a workspace
5. Click on Subtract
6. In the Subtract Workspace dialog click on Ok
7. Select the initial workspace once more
8. Click on Subtract
9. In the Subtract Workspace dialog click on Ok

Fixes [#528](https://github.com/mantidproject/mslice/issues/528)

### DIFF
--- a/mslice/workspace/histogram_workspace.py
+++ b/mslice/workspace/histogram_workspace.py
@@ -55,7 +55,7 @@ class HistogramWorkspace(HistoMixin, WorkspaceOperatorMixin, WorkspaceMixin, Wor
 
     def __del__(self):
         try:
-            if hasattr(self, '_raw_ws') and self._raw_ws.name().endswith('_HIDDEN') and self._raw_ws is not None:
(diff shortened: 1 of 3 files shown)


mducle (commented):
Thanks this is looks like a good implementation but I think the check for `self._raw_ws is not None` should be second - I think sometimes the error is in the `name()` call if the workspace is already deleted. So it should be:

```
if hasattr(self, '_raw_ws') and self._raw_ws is not None and self._raw_ws.name().endswith('_HIDDEN') :
```

Since the second if should short-circuit and the third is not evaluated.

SilkeSchomann:
Good point. I will change the sequence in the respective if conditions as suggested.

mducle (approved):

Issue #576, opened 2021-01-26
Subtraction of one workspace from another does not result in expected workspace


SilkeSchomann:
**Expected Behavior**
When scaling one workspace to match another workspace and then subtracting the two the result should be 0 (for the Y values).

**Actual Behavior**
When scaling one workspace to match another workspace and then subtracting the two the result is not 0 (for the Y values).

**Steps to reproduce the behavior**
There is a test script called Mslice_IRIS_scaling.py and corresponding data on babylon under Silke.

SilkeSchomann:
Code is merged into release-next

PR #577, opened 2021-01-26
Fixed bug that overwrote selection of background workspace when subtracting

SilkeSchomann:
Subtraction of one workspace from another did not result in the expected workspace in MSlice

Description of work.
Modified user_input function in input_boxes.py to return the workspace selected by the user as the background workspace.

**To test:**

1. Open MSlice in IDAaaS
2. Load two datasets that result in compatible workspaces
4. Scale the first workspace with a random scale factor
5. Select the second workspace
6. Click on Subtract
7. Select the scaled workspace
8. Ensure that the subtracted workspace uses the scaled workspace as the background workspace

Fixes [#576](https://github.com/mantidproject/mslice/issues/576).

### DIFF
--- a/mslice/widgets/workspacemanager/input_boxes.py
+++ b/mslice/widgets/workspacemanager/input_boxes.py
@@ -14,7 +14,7 @@ class SubtractInputBox(QDialog):
         self.listWidget.setCurrentRow(0)
 
     def user_input(self):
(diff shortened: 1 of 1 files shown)


DanielMurphy22 (approved):
Well done! This simple fix stops the first workspace being automatically selected and now behaves as expected.

Issue #573, opened 2021-01-21
User bug: Mslice crashes when converting Intensity from S(Q,E) to GDOS


SilkeSchomann:
**Expected Behavior**
Adjust the plot according to the provided sample temperature.

**Actual Behavior**
MSlice crashes.

**Steps to reproduce the behavior**

1. Open MSlice in IDAaaS.
2. Load any data
3. Go to the Workspace Manager tab
4. Click on the Slice tab and then on Display.
5. Click on Intensity
6. Select GDOS
7. Select MSlice from drop down menu (alternatively enter double with unit)

PR #583, opened 2021-02-22
Added exception handling for different input values for GDOS to prevent MSlice crashes.

SilkeSchomann:
Mslice crashed often when converting Intensity from S(Q,E) to GDOS. One of the problems are unexpected input values.
Also, temperature input in Kelvin will not always have the unit appended. This fix allows both values with and without unit.
When selecting a pre-configured setting for GDOS, for instance 'MSlice', this also often resulted in crashes whenever there is no valid entry for temperature. 

- When entering an invalid temperature an error message is displayed. 
- The function trying to convert an input value with or without 'K' as a unit is more robust now (get_sample_temperature_from_string).
- Selecting a pre-configured setting for GDOS that does not have a valid temperature now throws the exception 'sample temperature not found' instead of crashing MSlice.

**To test:**

1. Open MSlice
2. Load any data
4. Click on the Slice tab and then on Display.
5. Click on Intensity
6. Select GDOS
7. Try different selections from drop down menu. 'MSlice' seemed to cause crashes quite often, now it should at most throw an exception. 
8. Also, try different values as temperature input. A double or integer with or without 'K' appended should work. Any other entry should result in a pop up error message, but no crash.

Fixes #[573](https://github.com/mantidproject/mslice/issues/573).

### DIFF
--- a/mslice/plotting/plot_window/slice_plot.py
+++ b/mslice/plotting/plot_window/slice_plot.py
@@ -342,19 +342,22 @@ class SlicePlot(IPlot):
                 self._slice_plotter_presenter.add_sample_temperature_field(temp_value)
                 self._slice_plotter_presenter.update_sample_temperature(self.ws_name)
             else:
-                try:
-                    temp_value = float(temp_value)
-                    if temp_value < 0:
-                        raise ValueError
-                except ValueError:
-                    self.plot_window.error_box("Invalid value entered for sample temperature. Enter a value in Kelvin \
+                if temp_value is not None and temp_value.strip():
+                    if temp_value.endswith('K'):
+                        temp_value = temp_value[:-1]
+                    try:
+                        temp_value = float(temp_value)
+                        if temp_value < 0:
+                            raise ValueError
+                    except ValueError:
+                        self.plot_window.display_error("Invalid value entered for sample temperature. Enter a value in Kelvin \
                                                or a sample log field.")
-                    self.set_intensity(previous)
-                    return False
-                else:
-                    self.default_options['temp'] = temp_value
-                    self.temp = temp_value
-                    self._slice_plotter_presenter.set_sample_temperature(self.ws_name, temp_value)
+                        self.set_intensity(previous)
+                        return False
+                    else:
+                        self.default_options['temp'] = temp_value
+                        self.temp = temp_value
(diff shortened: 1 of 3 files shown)


srikanthravipati:
The fix handles the invalid input for temperature in a robust way, and displays error in such a case and prevents crash. I would recommend adding a few tests for the get_sample_temperature_from_string. If a function like raise_temperature_error, following my comment, is added, tests for that function also would be great.

mslice/plotting/plot_window/slice_plot.py
```diff
@@ -342,19 +342,22 @@ def _run_temp_dependent(self, slice_plotter_method, previous):
...
+                if temp_value is not None and temp_value.strip():
+                    if temp_value.endswith('K'):
+                        temp_value = temp_value[:-1]
+                    try:
+                        temp_value = float(temp_value)
+                        if temp_value < 0:
+                            raise ValueError
+                    except ValueError:
+                        self.plot_window.display_error("Invalid value entered for sample temperature. Enter a value in Kelvin \
                                                or a sample log field.")
```
srikanthravipati, line 354: Perhaps, we could use https://github.com/mantidproject/mslice/blob/7c5a6545548e54a9f965b841f30c63365a3f5157/mslice/models/slice/slice_functions.py#L145-L154
to get temperature 
``temp_value = get_sample_temperature_from_string(temp_value)``  
that leads to ``temp_value`` being either ``None`` or `` Float``. 

Then, we can check if there is an error and raise if needed
``if raise_temperature_error(temp_value):``
      ``self.plot_window.display_error("Invalid value entered for sample temperature. \``
                ``Enter a value in Kelvin or a sample log field.")``

where raise_temperature_error defined something like below

`def raise_temperature_error(temp_value):`
        `raise_error = (temp_value is None`)
        `if raise_error:`
               `pass`
         `else:`
               `raise_error = (temp_value < 0)`
 `return raise_error`

srikanthravipati (changes requested):
The fix handles an invalid input for temperature in a robust way, and displays error in such a case and prevents crash. I would recommend adding a few tests for the get_sample_temperature_from_string. If a function like raise_temperature_error, following my comment, is added, tests for that function also would be great.

srikanthravipati (approved):
Looks good :)

PR #595, opened 2021-04-07
Keep hidden cuts hidden while using exported script

### LINKED ISSUE #547: Bug: hidden curves in graph should still be hidden when exported to a script
### Expected behavior

Hidden curves in a plot stays hidden when running a script which is generated from that plot.

### Actual behavior

Hidden curves are shown.

### Steps to reproduce the behavior

Plot multiple line plots (using the `Cut` tab). Click on the wheel (settings) icon in the plot and hide one of the curves (unset one of the `Show Line` checkbox). Use `File->Generate Script File`. Run the generated script file - the hidden line will be plotted.

### Platforms affected

All

srikanthravipati:
Description of work.

While exporting to a script, the information regarding hidden cuts is lost and the cuts were being shown despite they were hidden. This fix adds the missing information to script and keeps the hidden cuts hidden even while using the exported script.

**To test:**

[hide_curves_start_script.txt](https://github.com/mantidproject/mslice/files/6272766/hide_curves_start_script.txt)
-> Update 'SET_CORRECT_PATH_TO_BUILD_PARENT' to parent directory of build directory in the attached 'hide_curves_start_script.txt'

-> Copy the python script from hide_curves_start_script.txt (make sure the above step is done to be able to read the input file correctly) and paste it in Mantid Workbench Jupyter IPython Console and Enter
-> Matplotlib figure with four lines and legends should show up.

-> Click on 'Plot Options (settings)' symbol on the figure. 
-> In the pop-up Qt window, right side column contains a tab for each and individual cut options
-> Uncheck 'show line' for one or more cuts and click 'Ok' at the bottom

-> Going back to the figure, check that lines, errorbars and legends are not shown for the cuts for which "show line" was unchecked
-> Click 'File-> Generate Script to Clipboard' on the top-left corner of the figure
-> Close the figure
-> Paste the script (in the buffer) in Jupyter IPython Console and Enter

-> Now a new matplotlib figure should show up. Make sure that the cuts for which "show line" was unchecked  are not shown 
-> Also verify that "show line" option for the cuts in "Plot Options (settings)" Qt window is unchecked.

Fixes #547.

### DIFF
--- a/mslice/cli/helperfunctions.py
+++ b/mslice/cli/helperfunctions.py
@@ -57,7 +57,6 @@ def _update_overplot_checklist(key):
 
 
 def _get_overplot_key(element, rmm):
-
     if element is not None and rmm is not None:
         raise RuntimeError('Cannot use both element name and relative molecular mass')
     if element is None and rmm is None:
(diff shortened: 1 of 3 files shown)


SilkeSchomann (changes requested):
Instead of adding the new helper functions to each script generated by MSlice I would add them to mslice.cli.helperfunctions permanently. Then append_visible_handle_and_label and hide_a_line_and_errorbars can be imported in _mslice_commands in mslice.cli and are available in the generated script.

srikanthravipati:
> Instead of adding the new helper functions to each script generated by MSlice I would add them to mslice.cli.helperfunctions permanently. Then append_visible_handle_and_label and hide_a_line_and_errorbars can be imported in _mslice_commands in mslice.cli and are available in the generated script.

Thanks.

SilkeSchomann (approved):
Thanks for the changes. This looks good now.

Issue #610, opened 2021-05-17
MSlice problems when creating empty plots without defining Cuts or Slices


SilkeSchomann:
**Describe the bug**
When running the following script Mantid crashes:

```
import mslice.cli as mc
import mslice.plotting.pyplot as plt

ws_MAR28230_Ei149_11meV = mc.Load(Filename='/instrument/MARI/CYCLE20211/RB2010456/autoreduced/MAR28230_Ei149.11meV.nxspe', OutputWorkspace='MAR28230_Ei149.11meV')
ws_MAR28229_Ei149_11meV = mc.Load(Filename='/instrument/MARI/CYCLE20211/RB2010456/autoreduced/MAR28229_Ei149.11meV.nxspe', OutputWorkspace='MAR28229_Ei149.11meV')
ws_MAR28228_Ei149_10meV = mc.Load(Filename='/instrument/MARI/CYCLE20211/RB2010456/autoreduced/MAR28228_Ei149.10meV.nxspe', OutputWorkspace='MAR28228_Ei149.10meV')

fig = plt.gcf()
ax = fig.add_subplot(111, projection="mslice")
ax.set_title('')
ax.set_ylabel(r'')
ax.set_xlabel(r'')
ax.set_ylim(bottom=0.0, top=1.0)
ax.set_xlim(left=0.0, right=1.0)
mc.Show()
```

The error message is

Traceback (most recent call last):
  File "/opt/mantidnightly/scripts/ExternalInterfaces/mslice/plotting/plot_window/plot_window.py", line 43, in showEvent
    self.canvas.manager.plot_handler.save_default_options()
AttributeError: 'NoneType' object has no attribute 'save_default_options'

**To Reproduce**
Run the script with the data in \\olympic\Babylon5\Public\Guidi\Tmp\Daniel\autoreduced.

**Expected behavior**
An empty plot window should appear (there are no cuts or slices created with the script). When closing the plot window by clicking on X in the upper right corner the plot should close without an error message.

PR #611, opened 2021-05-17
Fixes problems for empty plots

SilkeSchomann:
There are two fixes combined in here:
1. When displaying an empty plot window the AttributeError when calling save_default_options() without having a Cut or Slice window is caught.
2. When destroying a plot window that is neither Cut nor Slice window this window is removed from _unclassified_figures but not treated as a plot window with a category which would cause an exception and crash Mantid.

**To test:**

Use data and script from original issue: https://github.com/mantidproject/mslice/issues/610 and the current beta version from Mantid (with Mantid 6.0 this will be masked by a different bug).

Fixes #[610](https://github.com/mantidproject/mslice/issues/610).

### DIFF
--- a/mslice/plotting/globalfiguremanager.py
+++ b/mslice/plotting/globalfiguremanager.py
@@ -69,6 +69,9 @@ class GlobalFigureManager(object):
         """
         if not cls.has_fignum(num):
             return
+        if num in cls._unclassified_figures:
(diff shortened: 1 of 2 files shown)


srikanthravipati (approved):
The problem came up during beta testing. This PR fixes the issue and avoids a hard crash.

PR #613, opened 2021-05-20
Raise or bring forward plot window for plotover

### LINKED ISSUE #612: Make the (over)plot window visible at the top/front
`MSlice` allows overplotting cuts. If a user over-plots another cut on an existing plot window that is minimized or in the background, it would be useful to raise the corresponding window to the top if minimized or bring it to the front if in the background.

srikanthravipati:
**Description of work.**

When a `cut plot` is requested via the `plotover` option on an existing plot window, this fix brings the `plot window` to the top if it was minimized or forward if it was in the background.

The fix is inspired by

https://github.com/mantidproject/mslice/blob/33239062a80caf9d61f2b8167f9551e8b8768f82/mslice/plotting/plot_window/plot_figure_manager.py#L55-L58

**To test:**

-> Make a cut plot using `MSlice`
-> Minimize the above cut plot window (or bring forward some other window so that the plot window goes to the background)
-> Plot another cut using the plotover option 
-> Check if the plot window raises to the top if it was minimized or comes forward if it was in the background

Fixes #612.

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -250,7 +250,9 @@ class CutPlot(IPlot):
         self.plot_window.action_gen_script.setVisible(not is_icut)
         self.plot_window.action_waterfall.setVisible(not is_icut)
 
-        self.plot_window.show()
+        self.plot_window.showNormal()
+        self.plot_window.activateWindow()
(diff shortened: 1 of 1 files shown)


SilkeSchomann (approved):
Works as expected

PR #620, opened 2021-06-02
Update __init__.py

SilkeSchomann:
Changed version number to next version with string "dev0" to indicate ongoing development


DanielMurphy22 (approved):
Let's see if this works!

PR #623, opened 2021-07-01
Restore checked over-plot lines on a re-plotted slice

### LINKED ISSUE #515: Overplot info on slices lost after replot
When a slice is plot and a set of Bragg peaks or recoil lines are overplot on top ("Informational Overplot"), and then the slice is replotted (e.g. to change the colormap), the information overplot lines are cleared - but they are still shown as plotted in the menu. 

In addition, on Windows, after the information overplot, the response of the slice window is significantly slower than if there was no overplotting.

Desired behaviour is that after a slice replot, the information overplot lines should remain on the plot.

srikanthravipati:
**Description of work.**

Let say over-plots lines are plotted on an existing slice plot and then
the slice plot is re-plotted on the same plot window.

Previously plotted (checked) over-plot lines do not appear on the new slice plot.

This fix plots previously over-plotted lines on the re-plotted slice.

**To test:**

--> Plot a slice of a workspace (using Display)
--> Over-plot lines using Information/ recoil lines (or) Bragg  peaks from the menu
--> Re-plot slice (using Display)
--> Without this fix, only the slice would appear when re-plotted
--> With this fix, all the previously over-plotted lines should also appear when the slice is re-plotted

Fixes #515.

### DIFF
--- a/mslice/views/slice_plotter.py
+++ b/mslice/views/slice_plotter.py
@@ -36,6 +36,8 @@ def _show_plot(slice_cache, workspace):
     if cur_fig.canvas.manager.plot_handler.icut is not None:
         cur_fig.canvas.manager.plot_handler.icut.rect.ax = ax
 
+    cur_fig.canvas.manager.plot_handler._update_lines()
+
     cur_fig.canvas.draw_idle()
     cur_fig.show()
 


SilkeSchomann (approved):
Works as expected :-)

PR #624, opened 2021-07-12
Enable powder(Bragg) peaks on cut plots

### LINKED ISSUE #602: Feature request: Powder peaks in 1D cuts in Q/2theta
At the moment you can do an "Information overplot" on 2D slices, which includes recoil lines and powder lines.

Recoil lines don't make sense for 1D plots, but powder lines (peak positions in |Q| or 2theta) do make sense and would be a useful additional functionality.

Original suggester: @tatgu

srikanthravipati:
**Description of work**

Powder(Bragg) peaks are enabled on cut plots in addition to slice plots. 
Overplot interface is created taking material from powder peaks implementation for slice plots.

**To test:**

Create a slice plot
Make sure Information over-plots on slice plots work without any problems (The existing feature is not corrupted)

Create a cut plot
Then check that powder(Bragg) peaks (in Information) can be invoked on a cut plot

Fixes #602 .
PowderPeaks

### DIFF
--- a/mslice/models/powder/powder_functions.py
+++ b/mslice/models/powder/powder_functions.py
@@ -64,6 +64,4 @@ def _crystal_structure(ws_name, element, cif_file):
         LoadCIF(Workspace=ws, InputFile=cif_file)
         return ws.sample().getCrystalStructure()
     else:
-        return CrystalStructure(crystal_structure[element][0],
-                                crystal_structure[element][1],
-                                crystal_structure[element][2])
+        return CrystalStructure(*crystal_structure[element])
--- a/mslice/models/powder/powder_functions.py
+++ b/mslice/models/powder/powder_functions.py
@@ -1,0 +1,69 @@
+import numpy as np
+
+from scipy import constants
+
+from mantid.geometry import CrystalStructure, ReflectionGenerator, ReflectionConditionFilter
+
+from mslice.models.labels import is_momentum, is_twotheta
+from mslice.models.workspacemanager.workspace_provider import get_workspace_handle
+from mslice.util.mantid.mantid_algorithms import LoadCIF
+
+# energy to wavelength conversion E = h^2/(2*m_n*l^2)
+E2L = 1.e23 * constants.h ** 2 / (2 * constants.m_n * constants.e)
+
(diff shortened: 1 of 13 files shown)


mducle (changes requested):
It looks good... but I found a couple of issues. Those with suggested code changes (in the review comments) are:
* Tick marks can be too small to see if the intensity scale (counts) is large. This is because the code for the marks on the 2D slices scales the ticks by the incident (maximum) energy as the y-axis is energy in the 2D slices. For the 1D cuts, the y-axis is the intensity (counts) so this could be quite different.
* Sometimes the wrong tick marks are plotted - e.g. if you plot a 2theta cut and then plot a Q-cut and plot the powder lines on the Q-cut it might plot the tick marks as if the cut was actually 2theta.

An issue which I've not put in a code suggestion for is:
* When you plot the powder line on a figure and then replot another on the same figure (not overplotting), the powderline check-box is still checked but the actual line has been removed. Now, clicking the check-box it to uncheck it gives a `ValueError` - in `remove_line()` in `overplot_interface.py` (line 68).

mslice/models/powder/powder_functions.py
```diff
@@ -0,0 +1,67 @@
...
+    hkls = generator.getUniqueHKLsUsingFilter(d_min, d_max,
+                                              ReflectionConditionFilter.StructureFactor)
+    dvalues = np.sort(np.array(generator.getDValues(hkls)))[::-1]
+    return dvalues
+
+
+def _compute_powder_line_momentum(ws_name, q_axis, element, cif_file):
+    two_pi = 2.0 * np.pi
+    d_min = two_pi / q_axis.end
+    d_max = two_pi / q_axis.start
```
mducle, line 31:
```suggestion
    d_max = two_pi / np.max([q_axis.start, 0.01])
```

The original code gives a `ZeroDivisionError` if the user makes a cut starting at Q=0.

mducle (changes requested):
Sorry... just did one last check and found another issue...

mducle (approved):
Looks good now, and the new feature works as described. :shipit:

Issue #629, opened 2021-09-09
MSlice opens to plot page


AnthonyLim23:
MSlice opens to the plot tab instead of the loading tab. This is annoying for users

PR #631, opened 2021-09-14
Changed default tab to load tab

SilkeSchomann:
Changed the default tab from plot to load.

**To test:**

Open MSlice and make sure that the load tab is opened and not the plot tab.

Fixes #[629](https://github.com/mantidproject/mslice/issues/629).


sf1919 (approved):
I can confirm the Data Loading Tab now opens as default.

Issue #628, opened 2021-08-24
An error happens when generating a script to clipboard


robertapplin:
**Describe the bug**
An error appears after attempting to generate a script to clipboard from an MSlice plot.

**To Reproduce**
1. Open Mslice

2. Load MAR25087 in the Data Loading tab (provided below)
3. On the `Cut` tab, enter the following data before clicking `Plot`.

![image](https://user-images.githubusercontent.com/40830825/130593532-c60af6aa-3d12-4e8f-b50c-41e35e0bdc80.png)

4. A new window with a plot should open.
5. On the plot window, click `File`->`Generate Script to Clipboard`
6. I see this error in the log:

![image](https://user-images.githubusercontent.com/40830825/130595899-eea70341-1fae-41f4-a726-21204b562f36.png)

**Data Files**
[MAR25087_Ei40.00meV.zip](https://github.com/mantidproject/mantid/files/7037702/MAR25087_Ei40.00meV.zip)

**Mantid Version (please complete the following information):**
 - Windows Mantid nightly version from 19th August 2021.

**Additional context**

AnthonyLim23:
this is now a full crash

PR #632, opened 2021-09-14
Save default options for cut plot

SilkeSchomann:
Without saving the default options for cut plots scripts cannot be generated successfully.

**To test:**

Create cut plot. On the plot window, click File->Generate Script to Clipboard

Fixes #[628](https://github.com/mantidproject/mslice/issues/628).


sf1919 (approved):
I can confirm that following the steps in #628 that the error no longer appears. The code looks fine and seems to be a good solution. Happy to approve.

Issue #627, opened 2021-08-24
A crash happens when selecting a Bragg Peak option when there is no Bragg peak


robertapplin:
**Describe the bug**
A crash happens after selecting a Information->Bragg Peak->Copper when there is no Bragg peak in the data.

**To Reproduce**
1. Open Mslice

2. Load MAR25087 in the Data Loading tab (provided below)
3. On the `Cut` tab, enter the following data before clicking `Plot`.

![image](https://user-images.githubusercontent.com/40830825/130593532-c60af6aa-3d12-4e8f-b50c-41e35e0bdc80.png)

4. A new window with a plot should open.
5. Click `Information`->`Bragg Peaks`->`Copper`
6. The sad mantid appears and I see this error in the log:
![image](https://user-images.githubusercontent.com/40830825/130593898-8762a444-6d72-4c19-a327-ef09b84b9551.png)

**Data Files**
[MAR25087_Ei40.00meV.zip](https://github.com/mantidproject/mantid/files/7037702/MAR25087_Ei40.00meV.zip)

**Mantid Version (please complete the following information):**
 - Mantid Version [e.g. 6.0.0]

**Additional context**

PR #633, opened 2021-09-14
Added exception for bragg peak not found

SilkeSchomann:
Added exception to prevent crash when trying to display a Bragg peak when there is no Bragg peak in the data.

**To test:**

Follow the steps described in https://github.com/mantidproject/mslice/issues/627. Instead of a crash there is a warning 'No Bragg peak found.'

Fixes #[627](https://github.com/mantidproject/mslice/issues/627).


DavidFair:
This prevents the crash but doesn't completely solve the problem:
- Copper is still ticked despite the fact it's not selected in the information drop down
- There is no warning (apart from the console) that there is anything wrong. From the user POV nothing happens now

Lets get this shipped as "nothing happening" is far better than a crash

DavidFair:
Also I'm not a member of this repo so I cannot explicitly approve, but feel free to add an approval on my behalf whilst merging

DavidFair (approved):
Approving now that I can
There's some outstanding minor UI bugs, but the crash is much more important to solve

Issue #544, opened 2020-09-16
Error message for Ignored Exception and Runtime error when creating slice/loading files


sf1919:
As found during Beta testing

**Expected behaviour**
When creating a slice the messages should show successful Slice, Transpose and DeleteWorkspace.

**Actual behaviour**
After a successful slice an Ignored Exception and Runtime error occurs before as successful Transpose. User was accessing MSlice through IDAaaS.

![Mslice_error](https://user-images.githubusercontent.com/55837273/93321263-ea72af00-f809-11ea-80eb-d6bb79ff6343.JPG)

**Steps to reproduce**
It was not possible to reproduce this error. However the same error occurred when trying to open a file in MSlice with the same name as an existing workspace and choosing to overwrite it.

The steps to recreate this error are as follows
1) Open MSlice
2) Load file e.g. MAR21335_Ei60.00meV.nxs
3) Load file with same name but different file extension e.g. MAR21335_Ei60.00meV.nxspe
4) Click ok when Alert appears about Workspace name already being used and do you want to overwrite it
5) Error message appears in messages as per screenshot below

![error1](https://user-images.githubusercontent.com/55837273/93321620-463d3800-f80a-11ea-8d4a-392ac019b591.png)

**Platform affected**
Original issue involving taking slices found on IDAaaS Workbench Nightly version during Beta testing on 14/09/2020

Similar issue found when overwriting the Workspace occurred on Windows 10 using Mantid Nightly 5.1 version from 09/09/2020

sf1919:
Discussed with @martyngigg and decided this is a low priority as does not crash Mantid and in the case of taking a Slice does not appear to interfere with or affect the resulting plot. Therefore does not need to be fixed for Mantid 5.1 release,

SilkeSchomann:
The fix for https://github.com/mantidproject/mslice/issues/528 also fixes the runtime error described above. However, when following the steps to reproduce the problem the same runtime error is triggered in a different part of the code:

srikanthravipati:
Perhaps 5693a3fb77982701c9fa88a70d781aae1ef8a984 could be an inspiration; use of ``try`` and ``except``

PR #634, opened 2021-09-15
Fix for errors caused by small interactive cuts

SilkeSchomann:
In the MSlice matplotlib projection the matplotlib errorbar method should only be used when args[0] is not a HistogramWorkspace as otherwise parameters are either unknown or unsuitable (issue https://github.com/mantidproject/mslice/issues/626).

Another error occurred when trying to delete the same workspace twice. This problem was also described in https://github.com/mantidproject/mslice/issues/544. The solution is to handle the exception this causes.

**To test:**

Follow the instructions in https://github.com/mantidproject/mslice/issues/626 and https://github.com/mantidproject/mslice/issues/544. The described errors should not appear anymore.

Fixes #[626](https://github.com/mantidproject/mslice/issues/626) and #[544](https://github.com/mantidproject/mslice/issues/544).

### DIFF
--- a/mslice/cli/__init__.py
+++ b/mslice/cli/__init__.py
@@ -4,7 +4,7 @@ import mslice.util.mantid.init_mantid # noqa: F401
 from mslice.plotting.pyplot import *  # noqa: F401
 from matplotlib.axes import Axes
 from matplotlib.projections import register_projection
-from mslice.cli.helperfunctions import is_slice, is_cut
+from mslice.cli.helperfunctions import is_slice, is_cut, is_hs_workspace
 from ._mslice_commands import *  # noqa: F401
 from mslice.app import is_gui
 from mslice.cli.helperfunctions import (_check_workspace_name, _check_workspace_type, _get_overplot_key,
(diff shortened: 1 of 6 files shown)


DavidFair:
This correctly fixes #626 and #544

(As I can't leave formal reviews)
```python
mslice/cli/helper_functions.py
def is_hs_workspace(*args):
    """
    Checks if args[0] is a HistogramWorkspace
    """
    if isinstance(args[0], HistogramWorkspace):
        return True
    else:
        return False
```

Can be simplified to
```python
return isinstance(args[0], HistogramWorkspace)
```

And the try-except blocks that it repeated over and over could be collapsed into a method, so that any changes to it happen once rather than in 3+ places

DavidFair (approved):
I've re-tested this and it still works
@SilkeSchomann What's the policy on this repo for approved work: reviewer merges / selected list of maintainers / something else?

SilkeSchomann:
@DavidFair The reviewer can merge but in reality it is usually me.

PR #646, opened 2021-12-22
Fix for empty window in generated script

### LINKED ISSUE #639: Plot from generated script empty
**Describe the bug**
When generating a script from a plot window the plot produced from the script is empty when the original plot window was closed.

**To Reproduce**
1. In the ``Workspace Manager`` tab of the MSlice interface select the workspace ``MAR21335_Ei60meV``
2. Click ``Display`` in the ``Slice`` tab without changing the default values
3. Navigate to the ``File`` menu on the slice plot
4. Select ``Generate Script to Clipboard`` and paste the script into the Mantid editor.
5. Close the window with the slice plot
6. Run the script
7. The new slice plot window is empty

**Expected behavior**
In step 7. instead of an empty plot window the original slice plot should be displayed.

 - Mantid Version 6.2

SilkeSchomann:
This bug fix has to elements. The plot figure manager was updated to draw the canvas in the show method to ensure that the figure is visible. This bug was originally discovered for slice plots but occurred for both slice and cut plots.
The second element is removing the matplotlib pchanged call that activated the callback on_newplot for cut plots as this caused crashes. Now on_newplot is called directly for cut plots.

**To test:**

1. In the Workspace Manager tab of the MSlice interface load a data set (MAR21335_Ei60meV) and select the workspace 
2. Click Display in the Slice tab without changing the default values
3. Navigate to the File menu on the slice plot
4. Select Generate Script to Clipboard and paste the script into the Mantid editor.
5. Close the window with the slice plot
6. Run the script
7. The original slice plot should be displayed

Please repeat the above steps for the Cut tab and ensure that the original cut plot is displayed (and no crash occurs)

Fixes #639.

_A release note will be added for Mantid when this bug fix is made available for Mantid._

### DIFF
--- a/mslice/cli/plotfunctions.py
+++ b/mslice/cli/plotfunctions.py
@@ -71,7 +71,7 @@ def errorbar(axes, workspace, *args, **kwargs):
     if not cur_canvas.manager.has_plot_handler():
         cur_canvas.manager.add_cut_plot(presenter, workspace.name)
     cur_fig.canvas.draw()
-    axes.pchanged()  # This call is to let the waterfall callback know to update
+    cur_canvas.manager.update_axes(axes)
 
     cut = Cut(cut_axis, int_axis, intensity_min, intensity_max, workspace.norm_to_one, width='',
               algorithm=workspace.algorithm)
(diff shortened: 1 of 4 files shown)


martyngigg (approved):
Confirmed that the script in both modes now works and I see the same plot in scripg vs GUI

PR #658, opened 2022-01-10
Fix print button import error

### LINKED ISSUE #657: The print button on plots does not work anymore
**Describe the bug**
Clicking on the print button on a plot window raises the following error message:

  File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\plot_figure_manager.py", line 174, in print_plot
    printer = QtWidgets.QPrinter()
AttributeError: module 'qtpy.QtWidgets' has no attribute 'QPrinter'

mducle:
A quick fix to the imports for `QPrinter` and `QPrintDialog` because the module had changed from `QtWidgets` in QtPy4 when the code was originally written to `QtPrintSupport` in `qtpy` which we're using now.

**To test:**

Plot a cut, and click on the printer icon - it should show up a dialog. 
Print to a pdf - a file should be generated.
Previously it gave an `AttributeError`.

Fixes #657

### DIFF
--- a/mslice/plotting/plot_window/plot_figure_manager.py
+++ b/mslice/plotting/plot_window/plot_figure_manager.py
@@ -3,7 +3,7 @@ import weakref
 
 import six
 from qtpy.QtCore import Qt
-from qtpy import QtCore, QtGui, QtWidgets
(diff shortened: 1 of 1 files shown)


SilkeSchomann (approved):
Works as described

PR #664, opened 2022-01-11
Cleaned up File menu for interactive cuts

SilkeSchomann:
Disabled 'Generate Script to Clipboard' to match disabled 'Generate Script File'.


DavidFair (approved):
Interactive cut plots only have close now, so this works correctly :+1:

PR #666, opened 2022-01-11
Made lines for Bragg peaks editable

### LINKED ISSUE #655: Bragg peaks on cut plots are not editable
When clicking on a Bragg peak on a cut plot not a line editor opens but a KeyError is raised.

SilkeSchomann:
Description of work.

**To test:**

2. Navigate to the Cut tab
6. Navigate to the Information menu on the cut plot
7. Select Aluminium from the submenu for Bragg peaks. Green lines should appear on the cut plot with a respective legend entry.
8. Double click on the green line. A line editor should open. Change the colour of the line and check that plot changed accordingly.

Fixes #655.

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -170,11 +170,31 @@ class CutPlot(IPlot):
 
     def get_line_options(self, line):
         index = self._get_line_index(line)
-        return self.get_line_options_by_index(index)
+        if index >= 0:
+            return self.get_line_options_by_index(index)
+        else:
+            line_options = {
+                'label': line.get_label(),
+                'legend': None,
+                'shown': None,
+                'color': to_hex(line.get_color()),
+                'style': line.get_linestyle(),
+                'width': str(int(line.get_linewidth())),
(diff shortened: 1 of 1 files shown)


gemmaguest (approved):
Code changes look fine and this works as described :+1:.

I found some strange behaviour but this is not related to this PR; noting it here in case it needs looking at:
- after double clicking a line, the dialog that pops up always pops up a second time after you close it
- if you use the dialog to delete a line, it doesn't seem to work, and it gives an error in the terminal: `RuntimeError: wrapped C/C++ object of type QCheckBox has been deleted`

But as I say this is not related to this PR so let's :shipit:

SilkeSchomann:
> I found some strange behaviour but this is not related to this PR; noting it here in case it needs looking at:
> 
> * after double clicking a line, the dialog that pops up always pops up a second time after you close it
> * if you use the dialog to delete a line, it doesn't seem to work, and it gives an error in the terminal: `RuntimeError: wrapped C/C++ object of type QCheckBox has been deleted`
> 
Thanks for mentioning these! I will open issues for both of them.

PR #667, opened 2022-01-11
Prevent exception when generating script from plot from script

### LINKED ISSUE #652: Generating a script from a cut plot that is the result of a generated script causes crash
When trying to generate a script from a cut plot that was created through a script MSlice crashes with "AttributeError: 'TypeError' object has no attribute 'message'".

2. Navigate to the Cut tab
6. Navigate to the File menu on the cut plot
7. Select Generate Script to Clipboard and paste the script into the Mantid editor. 
8. Close the cut plot window
9. Run the script
10. Navigate to the File menu on the cut plot
11. Select Generate Script to Clipboard. Exception AttributeError: 'TypeError' object has no attribute 'message' appears.

Note: the same problem exists for slice plots as well.

SilkeSchomann:
When a plot is created by running a script generated from a plot the default options are not saved. This caused an exception when trying a to generate a new script from the plot. Now not existing default options are treated as unchanged options.

**To test:**

Follow instructions in original issue. Please note that this has been fixed both for slice and cut plots.

Fixes #652.

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -514,4 +514,6 @@ class CutPlot(IPlot):
         self.plot_window.waterfall_y = value
 
     def is_changed(self, item):
+        if self.default_options is None:
+            return False
         return self.default_options[item] != getattr(self, item)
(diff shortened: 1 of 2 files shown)


DavidFair (approved):
Could reproduce this crash without the changes
With the changes this crash is fixed

Issue #531, opened 2020-04-27
CLI not picking up workspace from names correctly


mducle:
The algorithm wrapper does not seem to pick up the correct workspace / workspace type from input workspace names. If given a workspace reference it works, but not if the workspace names are given.

For example:

```py
ws = mc.Load(Filename='MAR21335_Ei60.00meV.nxs', OutputWorkspace='MAR21335_Ei60.00meV')
ws1 = mc.Scale('MAR21335_Ei60.00meV', 2, OutputWorkspace='ws_scale')
ws2 = mc.Minus(LHSWorkspace='ws_scale', RHSWorkspace='MAR21335_Ei60.00meV', OutputWorkspace='ws_minus')
sl = mc.Slice('ws_minus', Axis1='|Q|', Axis2='DeltaE')
```

does not work, but:

```py
ws = mc.Load(Filename='MAR21335_Ei60.00meV.nxs', OutputWorkspace='MAR21335_Ei60.00meV')
ws1 = mc.Scale(ws, 2, OutputWorkspace='ws_scale')
ws2 = mc.Minus(LHSWorkspace=ws1, RHSWorkspace=ws, OutputWorkspace='ws_minus')
sl = mc.Slice(ws2, Axis1='|Q|', Axis2='DeltaE')
```

does work.

PR #670, opened 2022-01-13
Fixed bug in processing of input workspaces

### LINKED ISSUE #649: Exception when trying to scale a workspace
When using the 'Scale' button for a workspace a KeyError exception is thrown.

SilkeSchomann:
The fix for the bug https://github.com/mantidproject/mslice/issues/531 included returning workspace handles for all input workspaces that are provided by the workspace name. This caused problems when using the Scale button on Mslice. By being more selective when to return a workspace handle the original fix still works and the Scale button is usable again as well.

**To test:**

2. Click on the Scale button
3. Click ok on the pop up window. Check that no exception is thrown.

Also quickly check if https://github.com/mantidproject/mslice/issues/531 is still fixed. 

Fixes #649.

### DIFF
--- a/mslice/util/mantid/algorithm_wrapper.py
+++ b/mslice/util/mantid/algorithm_wrapper.py
@@ -12,10 +12,9 @@ from mslice.workspace.workspace import Workspace as MsliceWorkspace2D
 
 
 def _parse_ws_names(args, kwargs):
-    input_workspace = None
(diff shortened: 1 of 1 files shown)


DavidFair (approved):
Retested, this still works. Thanks for that :slightly_smiling_face:

PR #672, opened 2022-01-13
Added error message when attempting to load by file path

### LINKED ISSUE #671: "Invalid File Path" trying to load file by path directly
**Describe the bug**
This is either:
- A bug where files aren't handled correctly
Or
- A misleading error message

Trying to load a file by path in data loading doesn't work. Opening a directory does however (as expected).

**To Reproduce**
- Get the full file path to a file e.g.
`/home/david/mantidPrBuild/ExternalData/Testing/Data/UnitTest/MAR21335_Ei60meV.nxs`
- In data loading paste this whole path into the top bar
- Hit enter

**Expected behavior**
Either:
- The file loads
or
- An error message stating "file paths are not supported"

SilkeSchomann:
MSlice does not allow to load a file by entering the file path and name into the file path box on top of the data loader tab. In order to differentiate this case from entering an invalid file path a new error message was added: "Loading by file path is not supported"

**To test:**

Enter the full path to a file with the filename into the box on top of the data loader tab, e.g. C:\data\MAR21335_Ei60meV.nxs and hit enter

Fixes #671.

### DIFF
--- a/mslice/widgets/dataloader/dataloader.py
+++ b/mslice/widgets/dataloader/dataloader.py
@@ -5,7 +5,7 @@ import os
 from functools import partial
 
 from qtpy.QtWidgets import QWidget, QFileSystemModel, QAbstractItemView, QMessageBox
-from qtpy.QtCore import Signal, QDir, Qt
+from qtpy.QtCore import Signal, QDir, Qt, QFileInfo
 
 from mslice.presenters.data_loader_presenter import DataLoaderPresenter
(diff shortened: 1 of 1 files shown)


SilkeSchomann:
The unit tests for the data loader check that loading one ore more files work and all functionality around that (merge etc.) are ok, not the error messages for the file path box. I think it is probably not necessary.

DavidFair (approved):

PR #677, opened 2022-01-17
Changed Mantid into MSlice and added bug label

### LINKED ISSUE #650: Update template for bug
Currently the bug template asks for the Mantid version. This should be changed into MSlice.

Also, the label bug should be added automatically.

SilkeSchomann:
The bug template now asks for the MSlice version instead of the Mantid version. In addition, it automatically adds the 'bug' label.

**To test:**

Check the changes to the bug_report.md file.

Fixes #650.


mducle (approved):
Looks good to me.

Do we expect users to file bugs in github? If so, should we have an "About" help page with the version number on it?

Alternatively the template could be changed to give a hint, e.g. 

```
```

or something similar...?

mducle (dismissed):

mducle (approved):

PR #679, opened 2022-01-18
Moved some imports to top

### LINKED ISSUE #645: Move imports to top of file
To clean up the code move imports in methods to the top of the respective file.

SilkeSchomann:
To clean up the code a bit some of the imports were moved to the top of their respective files. Unfortunately it is not possible to move all imports in this way as it causes problems with circular dependencies.

**To test:**

Check that tests are still passing.

Fixes #645.

### DIFF
--- a/mslice/quickview/workspace_manager_quickview.py
+++ b/mslice/quickview/workspace_manager_quickview.py
@@ -1,4 +1,5 @@
 from __future__ import (absolute_import, division, print_function)
+from command import Command as c
 from quickview.QuickView import QuickView
 from WorkspaceManagerPresenter import WorkspaceManagerPresenter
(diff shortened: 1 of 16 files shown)


mslice/quickview/workspace_manager_quickview.py
```diff
@@ -17,6 +18,5 @@ def __getattribute__(self, item):
 
 
 if __name__ == '__main__':
-    from command import Command as c
```
mducle, line 20: Is this file used anywhere? (Actually it seems the whole `quickview` module is kind-of orphaned... it seems it was used to mock up the ui when MSlice was first written 6 years ago and then not touched...)

mducle (approved):

PR #683, opened 2022-01-21
Fix #682 Cut method:Integration scale bug

### LINKED ISSUE #682: Intensity factor of 100 smaller in MSlice powder cuts than SofQW3
**Describe the bug**

MSlice uses the Mantid `SofQWNormalisedPolygon` (aka `SofQW3`) algorithm to make cuts (rebins) of the data. However, plotting the data made by the `Cut` algorithm in MSlice *using the `Integration` method* [Menu:Options->Cut Algorithm] vs using `SofQW3` directly in Mantid shows a factor of 100x difference in the MSlice result vs the Mantid result. This is probably due to the conversion from `MatrixWorkspace` to `MDHisto` in [this section](https://github.com/mantidproject/mslice/blob/main/mslice/models/cut/cut_algorithm.py#L98-L114).

Edit: This seems to be because the data is a "Distribution" where the y-data is stored as frequencies (counts per bin-size). In the "Integration" cut method, MSlice first [rebins the data into 100 bins](https://github.com/mantidproject/mslice/blob/main/mslice/models/cut/cut_algorithm.py#L153) in the integration axis and then uses the `Integration` algorithm on this rebinned workspace. What appears to happen is that the rebinning causes the counts to be rescaled by the new bin sizes. Now `Integration` instead of just summing the stored frequencies, seems to assume they are actually counts, sums `y*binsize` [instead](https://github.com/mantidproject/mantid/blob/1d8dd9ba533a9226d402acf27405974210085894/Framework/Algorithms/src/Integration.cpp#L284) which applies another factor of 100 (since there are 100 bins).

A quick fix to this issue is to then scale the integrated data by the number of bins...

**To Reproduce**

Run this script:

```
# import mantid algorithms, numpy and matplotlib
from mantid.simpleapi import *
import matplotlib.pyplot as plt
import numpy as np

# Python Script Generated by Mslice on 2022-01-21 12:04:42
import mslice.cli as mc
import mslice.plotting.pyplot as mplt

ws = mc.Load(Filename='LET73024_3.70meV.nxspe', OutputWorkspace='LET73024_3.7meV')

fig = mplt.gcf()
ax = fig.add_subplot(111, projection="mslice")
cut_ws_0 = mc.Cut(ws, CutAxis="|Q|,0.0,5.0,0.02", IntegrationAxis="DeltaE,-0.5,0.5,0.0", NormToOne=False, Algorithm='Integration')
ax.errorbar(cut_ws_0, label="cut", color="#1f77b4", marker="o", ls="-", lw=1.5)

from mslice.cli.helperfunctions import hide_a_line_and_errorbars, append_visible_handle_and_label
from mslice.util.compat import legend_set_draggable

# hide lines, errorbars, and legends
handles, labels = ax.get_legend_handles_labels()
visible_handles = []
visible_labels = []

append_visible_handle_and_label(visible_handles, handles, visible_labels, labels, 0)

legend_set_draggable(ax.legend(visible_handles, visible_labels, fontsize='medium'), True)

ax.set_title('')
ax.set_ylabel(r'Signal/#Events')
mc.Show()

# Now do the pure Mantid rebinning
sqw_cut = SofQW3(ws.raw_ws, EMode='Direct', QAxisBinning='0,0.02,5', EAxisBinning='-0.5,10,0.5')
sqw_cut = Transpose(sqw_cut)
fig = plt.gcf()
ax = fig.add_subplot(111, projection="mantid")
ax.errorbar(sqw_cut)
fig.show()
```

If the `Algorithm='Integration'` is omitted (so the default 'Rebin' algorithm is used) then the two cuts are the same.

This should occur for any workspaces, but the nxspe file used here (from the user's bug report) is available on `/babylon/Public/duc/Ross\ nxspe\ files/`

**Expected behavior**
The two plots shown should be identical.

6.2.20220120.1649

mducle:
Adds a scaling by the number of bins which the workspace was rebinning into in the integration axis in the "Integration" method of `Cut`. For details, see original issue #682

**To test:**

Run the script in the issue #682 and check that the plots with MSlice and with pure Mantid are the same.

Fixes #682

### DIFF
--- a/mslice/models/cut/cut_algorithm.py
+++ b/mslice/models/cut/cut_algorithm.py
@@ -2,7 +2,7 @@ import numpy as np
 
 from mantid.api import PythonAlgorithm, WorkspaceProperty
 from mantid.kernel import Direction, PropertyManagerProperty, StringMandatoryValidator, StringListValidator
-from mantid.simpleapi import BinMD, ConvertSpectrumAxis, CreateMDHistoWorkspace, Rebin2D, SofQW3, TransformMD, \
+from mantid.simpleapi import BinMD, ConvertSpectrumAxis, CreateMDHistoWorkspace, Rebin2D, SofQW3, TransformMD, Scale, \
     ConvertToMD, DeleteWorkspace, CreateSimulationWorkspace, AddSampleLog, CopyLogs, Integration, Rebin, Transpose
 
 from mslice.models.alg_workspace_ops import fill_in_missing_input, get_number_of_steps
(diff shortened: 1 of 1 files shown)


SilkeSchomann (approved):
Plots are identical now when running the script from the original issue.

mducle:
@SilkeSchomann So actually, Ross came back to me on this, and after further investigation I found that it was actually a bug in the `Integration` algorithm in [Mantid#33407](https://github.com/mantidproject/mantid/issues/33407) - there is a PR [Mantid#33408](https://github.com/mantidproject/mantid/pull/33408) to fix it, and when that goes through in Mantid we should revert this PR in MSlice...

PR #684, opened 2022-01-21
Replace conda nose with pip nose

SilkeSchomann:
For some reason nose did not get installed properly anymore for GitHub workflows. By using pip instead of conda for installing nose this problem is fixed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,6 @@ dependencies:
   - ipython
   - matplotlib
   - mock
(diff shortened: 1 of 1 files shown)


mducle (approved):
CI now works again... :fireworks: 

Weirdly the nose version for both the successful and failed run is both `1.3.7` but the Conda package somehow was updated because the working version was 128KB and the failed on is 142KB...

PR #692, opened 2022-01-28
Changed default visibility for various menu items

SilkeSchomann:
Disabled some of the menu items and buttons on the MSlice plot window to allow plotting of data not from an MSlice workspace, for instance arrays.

The bar now only contains buttons for zoom in, zoom out, Keep, Make Current, Copy and Print. The menus for Information and Intensity are visible but disabled. The menu File only has the entry Close.

**To test:**

Run the script from the initial issue.

Fixes 686.


mducle (approved):
Looks good to me!

Issue #689, opened 2022-01-27
Update screenshots in documentation


SilkeSchomann:
For cutting.rst cut_panel.png and for quickstart.rst cut_q.png and en_units_combo.png are out of date.

SilkeSchomann:
Also mention that Save to Workbench is only available when MSlice is run as a Mantid interface, not in standalone mode.

SilkeSchomann:
Add comments about cut algorithm changes from #686 as well

PR #696, opened 2022-02-01
Update cut algorithm names and added combo box on cut tab

### LINKED ISSUE #687: Misleading "Cut Algorithm" names
**Describe the bug**

This is not really a bug, more of a documentation / notation issue.

MSlice has two methods to compute 1D cuts *for "non-PSD" data*, which are accessible in the "Cut Algorithms" menu, and called:

1. `Rebin` - this uses the `SofQW3` algorithm directly and effectively **averages** the counts in the integration range
2. `Integration` - this first rebins the data with `SofQW3` and then calls `Integration`, and effectively **sums** the counts in the integration range.

At the very least the menu option names should be changed to `Rebin (Averages Counts)` and `Integration (Sum Counts)`.

If there's time, it might be good also to add a new drop-down menu with these titles to the "Cut" tab of the main window so that these options are more visible to users (but only when the workspace is of a "non-PSD" type, since for "PSD" type data, only the `BinMD` algorithm is used which always averages the counts).

**To Reproduce**
N/A

SilkeSchomann:
The cut algorithm names were misleading and easily overlooked as only available as a global setting in the main menu.
The displayed names are now updated and the cut tab has a combo box to select the cut algorithm individually (using the previously unused slot for the Smooth input ). The logic behind this is the same as for the energy units. The global setting changes the current workspace and every new workspace, previous workspaces remember their setting.

**To test:**

Check both global and local cut algorithm settings.

The documentation is not yet updated to reflect this and will be updated as part of #689.

Fixes #687.

### DIFF
--- a/mslice/presenters/cut_widget_presenter.py
+++ b/mslice/presenters/cut_widget_presenter.py
@@ -202,7 +202,5 @@ class CutWidgetPresenter(PresenterUtility):
         self._cut_view.set_energy_units(en_default)
 
     def set_cut_algorithm_default(self, algo_default):
-        print("cut default")
-        print(algo_default)
         self._cut_view.set_cut_alg_default(algo_default)
         self._cut_view.set_cut_algorithm(algo_default)
--- a/mslice/presenters/cut_widget_presenter.py
+++ b/mslice/presenters/cut_widget_presenter.py
@@ -95,7 +95,7 @@ class CutWidgetPresenter(PresenterUtility):
 
         norm_to_one = bool(self._cut_view.get_intensity_is_norm_to_one())
         width = self._cut_view.get_integration_width()
-        algo = self._cut_view.get_cut_algorithm()
+        algo = ['Rebin', 'Integration'][self._cut_view.get_cut_algorithm()]
         return cut_axis, integration_axis, intensity_start, intensity_end, norm_to_one, width, algo
 
     def _set_minimum_step(self, workspace, axis):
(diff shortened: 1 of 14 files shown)


mslice/presenters/cut_widget_presenter.py
```diff
@@ -95,7 +95,7 @@ def _parse_input(self):
 
         norm_to_one = bool(self._cut_view.get_intensity_is_norm_to_one())
         width = self._cut_view.get_integration_width()
-        algo = self._main_presenter.get_cut_algorithm() if self._main_presenter is not None else 'Rebin'
+        algo = self._cut_view.get_cut_algorithm()
```
mducle, line 98:
```suggestion
        algo = ['Rebin', 'Integration'][self._cut_view.get_cut_algorithm()]
```

This is called by `_cut` to get parameters for the `Cut` algorithm, which expects a string in the `Algorithm` keyword argument. Whilst `_main_presenter.get_cut_algorithm` returned a string, `_cut_view.get_cut_algorithm` returns an index... and this causes the `Cut` algorithm to fail.

There is a `_name_to_index` but it is defined in the `CutWidget` class only so can't be used here...

mducle (changes requested):
Looks great and works as described.

Just one small change - could you change the text in the [mainwindow menu code](https://github.com/mantidproject/mslice/blob/7022f7cac2f3a848d231254e594d8c60d823cd06/mslice/app/mainwindow.ui#L429) from `Cut algorithm` to `Cut algorithm default`, please?

Also, in the merge commit text could you note that the previously unused slot for the `Smooth` input was used for this option, please? (this is just to allow easier future search of the history)

mducle (approved):

SilkeSchomann:
Sure, I've also updated the text of this PR to include information on replacing Smooth on the UI.

PR #700, opened 2022-02-01
Reverted changes from integration scale bug fix

SilkeSchomann:


mducle (approved):
Looks good to me.

As noted in https://github.com/mantidproject/mslice/pull/683 the scaling is no longer needed as the underlying issue with the `Integration` algorithm is fixed in [Mantid#33408](https://github.com/mantidproject/mantid/pull/33408/).

Issue #693, opened 2022-01-29
Update pyplot to MPL3 needed.


mducle:
**Describe the bug**

While testing #692 on Windows, where Mantid ships with Matplotlib 3.1.2, running the example script in the issue that PR fixes (#686) gives this error:

```python
~\src\mslice\mslice\plotting\pyplot.py in plot(*args, **kwargs)
   3200     # Deprecated: allow callers to override the hold state
   3201     # by passing hold=True|False
-> 3202     washold = ax._hold
   3203     hold = kwargs.pop('hold', None)
   3204     if hold is not None:

AttributeError: 'AxesSubplot' object has no attribute '_hold'
```

This is because the hold on/off and overplot functionality of Matplotlib was [deprecated](https://github.com/matplotlib/matplotlib/blob/f6e0ee49c598f59c6e6cf4eefe473e4dc634a58a/doc/api/prev_api_changes/api_changes_2.0.0.rst#hold-functionality-deprecated) in Matplotlib 2 and [removed](https://github.com/matplotlib/matplotlib/blob/7443d8842a0e613b36f97d945ecaad18aadad15f/doc/api/prev_api_changes/api_changes_3.0.0.rst#hold-machinery) in Matplotlib 3. On RHEL7 / SL7 used on DAaaS the system Python (3.6) still uses Matplotlib 2.2.3 and the `pyplot.py` used by MSlice was copied from Matplotlib 2 so still had the hold on/off code.

We should amend the `pyplot.py` code in MSlice so it will work with both Matplotlib 2 and 3 (we need a `try` block).

Edit: This bug was fixed in #701 with an `if` block but a longer term fix is to refactor `pyplot.py` so that we don't have such a large almost-verbatim copy of Matplotlib code.

**To Reproduce**

Run the script in #686 on a system with Matplotlib 3.

**Expected behavior**

MSlice should not give an exception and the plot curve should appear.

**MSlice Version (please complete the following information):**
 - MSlice Version: 2.3.2 with Mantid Nightly (6.2.20220127.1718) on Windows 10.

mducle:
It might also be worthwhile checking if instead of including the large `pyplot.py` file in MSlice we can monkey patch the `pyplot.py` directly from the running Matplotlib implementation with just the changes needed for MSlice's keep/make current mechanism...

mducle:
PR #701 fixes the problem for the release, so this issue is repurposed as a longer term maintenance task to refactor `pyplot.py`.

[...]

PR #701, opened 2022-02-01
Pyplot update

### LINKED ISSUE #686: Opening settings causes a crash if mslice pyplot used to plot arrays
**Describe the bug**

MSlice scripts generate a header with `import mslice.plotting.pyplot as plt` which overrides the usual alias (`plt`) that is used for `matplotlib.pyplot`. This can lead users to mistakenly plot plain lists / numpy arrays as they would do with plain `matplotlib.pyplot` but these plots then end up in an "MSlice window" with MSlice controls such as the settings button. However, the generated plots do not go through the MSlice projection and so there is not the required data which MSlice needs to populate the dialog, generating a crash.

**To Reproduce**

Run this script:

```python
import numpy as np
x = np.linspace(-3, 3, 100)
y = np.sin(x)
plt.plot(x,y)
plt.show()
```

Then click the "Settings" (cogwheel) button. The following error then occurs:

```python
   File "/opt/mantidnightly/scripts/ExternalInterfaces/mslice/plotting/plot_window/plot_figure_manager.py", line 176, in _plot_options    
       self.plot_handler.plot_options()
AttributeError: 'NoneType' object has no attribute 'plot_options'
```

**Expected behavior**
Settings dialog to appear or settings button to be hidden if the data is not from an MSlice workspace.

 - 6.2.20220126.1758 on DAaaS (SL7).

mducle:
MSlice has a version of `pyplot.py` from Matplotlib in order to provide a similar command line interface - this was copied almost verbatim from Matplotlib 1.5 and was updated to Matplotlib 2 several years ago. Since then Matplotlib 3 has been released which has removed the deprecated `hold` option and this now causes non-MSlice plotting to fail.

This change is just to check if the hold feature is present or not and if not to not crash due to an `AttributeError`. 

A more long lasting fix is to refactor the `pyplot.py` file to import from the underlying Matplotlib `pyplot` instead of copying its contents as described in issue #693

**To test:**

Run the script in issue #686 with a Matplotlib 3 installation and check that it does not error and a sine curve is plotted.

Fixes #686

### DIFF
--- a/mslice/plotting/pyplot.py
+++ b/mslice/plotting/pyplot.py
@@ -3199,6 +3199,8 @@ def plot(*args, **kwargs):
     ax = gca()
     # Deprecated: allow callers to override the hold state
     # by passing hold=True|False
+    if not hasattr(ax, '_hold'):
+        return ax.plot(*args, **kwargs)
     washold = ax._hold
     hold = kwargs.pop('hold', None)
     if hold is not None:


SilkeSchomann (approved):
Works as described :-)

PR #709, opened 2022-02-10
Updated screenshots and added cut algorithm info in documentation

### LINKED ISSUE #689: Update screenshots in documentation

SilkeSchomann:
All screenshots for the Cut tab were outdated after adding the Cut algorithm combo box and changing the 'Save to Workspace' button into 'Save to Workbench'. These screenshots have been replaced with current ones.
In addition, there is now information on the two methods for computing a cut.

**To test:**

Please check spelling and grammar of the new text and whether these new instructions make sense.

Fixes #689.


docs/source/cutting.rst
```diff
@@ -30,6 +30,10 @@ integrate over [0,3], [3,6], [6,9] and [9,10] respectively.
 
 Cuts with the same range from multiple datasets can be plotted by first selecting multiple workspaces in the left panel.
 
+There are two different methods to compute cuts for **non-PSD** data. ``Rebin`` uses the ``SofQW3`` algorithm directly and
+effectively averages the counts in the integration range. ``Integration`` rebins the data first with ``SofQW3`` and then
+calls ``Integration``, and effectively sums the counts in the integration range.
+
```
mducle, line 36: Actually... after PR #703 the two methods now for **PSD** data as well...

I suggest to keep it simple:

```
There are two different methods to compute cuts. ``Rebin`` uses the basic rebinning algorithms directly and effectively averages
the counts in the integration range, whilst ``Integration`` sums the counts in the integration range.
```

Or for more details:

```
There are two different methods to compute cuts. ``Rebin`` uses the ``SofQW3`` (``BinMD``) algorithm for **non-PSD** (**PSD**)
data directly and effectively averages the counts in the integration range. ``Integration`` rebins the data first with
``SofQW3`` or ``BinMD`` and then calls either ``Integration`` or ``IntegrateMDHistoWorkspace``, and effectively sums
the counts in the integration range.
```

instead.

mducle (changes requested):

docs/source/cutting.rst
SilkeSchomann: I think I prefer the simple version in this case.

mducle (approved):

PR #716, opened 2022-02-28
Update cut plot to allow for new font size quick options

### LINKED ISSUE #715: Quick Options doesn't work on Cut Plot
**Describe the bug**
Quick Options on Cut Plot doesn't work following changes made to enable change to font size on Slice Plots

**To Reproduce**
-In the Workspace Manager tab select the workspace MAR21335_Ei60meV
-Navigate to the Cut tab
-In the row labelled along, set the from value to 0 and the to value to 10
-In the row labelled over, set the from value to -5 and the to value to 5
-Click Plot. A new window with a cut plot should open.
-Right click on axis label to open quick options

MialLewis:
**Description of work:**

PR#710 extended the quick option interface to allow for the changing of font size. The slice plot model was updated to provided the properties necessary to enable this feature, but the cut plot model was not. 

This PR updates the cut plot model to enable this.

**To test:**
-In the Workspace Manager tab select the workspace MAR21335_Ei60meV
-Navigate to the Cut tab
-In the row labelled along, set the from value to 0 and the to value to 10
-In the row labelled over, set the from value to -5 and the to value to 5
-Click Plot. A new window with a cut plot should open.
-Right click on axis label to open quick options
-Check that changing the font size works as per slice plots.

Fixes #715.

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -4,6 +4,7 @@ from matplotlib.collections import LineCollection
 from matplotlib.container import ErrorbarContainer
 from matplotlib.legend import Legend
 from matplotlib.lines import Line2D
+from matplotlib.text import Text
 
 import warnings
 import numpy as np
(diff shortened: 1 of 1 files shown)


SilkeSchomann (approved):
Confirmed that font sizes can be changed on cut plots now.

PR #717, opened 2022-02-28
Update default cut algorithm to Integration

### LINKED ISSUE #705: Feature request: Default Cut method change to "Integration (summed counts)"
Historically MSlice used a rebinning algorithm (either `SofQW3` or `Rebin2D`) to generate 1D cuts. A subsequent user request added another option which summed the counts within the integration range. Recent investigations found that there was some errors with the normalisation used in this summed/integration option which has been fixed upstream in Mantid.

This option is particularly important for absolute units measurements, as the "Rebin (averaged counts)" method will give incorrect and misleading values, especially for different integration ranges. For non-absolute units measurements it doesn't really matter as there will be a scaling factor in any case. 

Thus, it would make sense to make the "Integration (summed counts)" Cut method the default, as it is required for absolute units measurements and it will not be incorrect for non-absolute units measurement. It will result in different values from that which users would have calculated historically though, so it may need a warning and it will need documentation.

MialLewis:
**Description of work**
The default cutting method has been updated from `rebin` to `integration`.

This has been done on both the GUI, and on related functions i.e. default algorithm argument changed from `rebin` to `integration`.

A print warning has been added to mslice on start up via the main window. The docs have been updated to detail this change.

**To test:**

Resolves #705.


SilkeSchomann:
There seems to be something going wrong when generating a script from a cut using the Rebin algorithm now. On the left hand side is the original plot, on the right hand side the plot created by running the script generated from the original plot. It does work when using the Integration algorithm instead though.

image

MialLewis:
> There seems to be something going wrong when generating a script from a cut using the Rebin algorithm now. On the left hand side is the original plot, on the right hand side the plot created by running the script generated from the original plot. It does work when using the Integration algorithm instead though.
> 
> image
> 
> I've replaced the MSlice interface in Mantid 6.3 with the branch mslice_705 merged into MSlice main to test this.

Interesting.. I'll take a look to see what could be causing this!

MialLewis:
> There seems to be something going wrong when generating a script from a cut using the Rebin algorithm now. On the left hand side is the original plot, on the right hand side the plot created by running the script generated from the original plot. It does work when using the Integration algorithm instead though.
> 
> image
> 
> I've replaced the MSlice interface in Mantid 6.3 with the branch mslice_705 merged into MSlice main to test this.

This should be fixed now, rebin was expected as default in a scripting helper function.

MialLewis:
> > There seems to be something going wrong when generating a script from a cut using the Rebin algorithm now. On the left hand side is the original plot, on the right hand side the plot created by running the script generated from the original plot. It does work when using the Integration algorithm instead though.
> > image
> > I've replaced the MSlice interface in Mantid 6.3 with the branch mslice_705 merged into MSlice main to test this.
> 
> This should be fixed now, rebin was expected as default in a scripting helper function.

Will this go out in a minor release or 6.4?

SilkeSchomann:
It should go into 6.4

PR #721, opened 2022-03-22
Update to matplotlib 3.5

### LINKED ISSUE #720: Investigate deprecated matplotlib functionality
Aim is to update matplotlib to 3.5.* in conda. On windows MSlice is able to run with matplotlib 3.2.1 (the version used by Mantid). Some functionality was deprecated since then.

Currently known issue is set_offset_position (deprecated in matplotlib 3.3): https://github.com/mantidproject/mslice/blob/6ffbdf4a1a1cdeadf2bbf519086408bb358b596c/mslice/plotting/plot_window/cut_plot.py#L401

This Mantid issue might be useful as some of the issues found there could be present in MSlice as well: https://github.com/mantidproject/mantid/issues/33529

Linked to MSlice issue https://github.com/mantidproject/mslice/issues/693

SilkeSchomann:
Added fixes for matplotlib functionality that is deprecated since 3.2.1 so that MSlice can be used with matplotlib 3.5 but still works with older versions as well (3.2.1 on Windows and 2.2.3 on Linux). 

**To test:**

1. Open MSlice
2. Load data, for instance MAR21335_Ei60meV.nxs from unit tests
3. Click Display on Slice tab and check that there are no error messages
4. Navigate to Cut tab
5. Enter '0' in the 'from' boxes as well as '1' in the 'to' boxes for x and y
6. Enter '0.2' in width
7. Click Plot and check that there is no error message and no warning apart from 'converting a masked element to nan.'
8. Click Waterfall and try different values for x and y offsets. There should be no warnings or error messages.

Fixes #720.

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -1,5 +1,7 @@
+from distutils.version import LooseVersion
 from functools import partial
 
+from matplotlib import __version__ as mpl_version
 from matplotlib.collections import LineCollection
 from matplotlib.container import ErrorbarContainer
 from matplotlib.legend import Legend
@@ -398,7 +400,8 @@ class CutPlot(IPlot):
                     line.set_xdata(self._waterfall_cache[line][0] + ind * x)
                     line.set_ydata(self._waterfall_cache[line][1] + ind * y)
                 elif isinstance(line, LineCollection):
-                    line.set_offset_position('data')
+                    if LooseVersion(mpl_version) < LooseVersion('3.3'):
+                        line.set_offset_position('data') # set_offset_position is deprecated since 3.3
                     line.set_offsets((ind * x, ind * y))
 
     def on_newplot(self, ax):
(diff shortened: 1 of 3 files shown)


mslice/plotting/plot_window/cut_plot.py
```diff
@@ -398,7 +400,8 @@ def _apply_offset(self, x, y):
                     line.set_xdata(self._waterfall_cache[line][0] + ind * x)
                     line.set_ydata(self._waterfall_cache[line][1] + ind * y)
                 elif isinstance(line, LineCollection):
-                    line.set_offset_position('data')
+                    if LooseVersion(mpl_version) < LooseVersion('3.3'):
+                        line.set_offset_position('data') # set_offset_position is deprecated since 3.3
```
Pasarus, line 404: Do we not need to do this on matplotlib 3.5?

Pasarus (changes requested):

mslice/plotting/plot_window/cut_plot.py
SilkeSchomann: No, set_offset_position was removed and not replaced. It seems to detect this automatically now.

Pasarus: Ah, brilliant, matplotlib has done this with a few things for 3.4/3.5

Pasarus:
When finishing the tests I get this error in the terminal:
```
The 'drawtype' parameter of __init__() was deprecated in Matplotlib 3.5 and will be removed two minor releases later. If any parameter follows 'drawtype', they should be passed as keyword, not positionally.
```
Are we concerned about this or is this something we will ignore until next time we upgrade matplotlib to 3.7?

SilkeSchomann:
I will open an issue to collect things like that for an upgrade to matplotlib 3.7.

Pasarus (approved):

PR #724, opened 2022-03-23
MSlice conda recipe and Github workflows for deployment

SilkeSchomann:
- Created noarch Python conda recipe for MSlice
- Added Github workflows for deploying MSlice nightly and main
- Cleaned up existing Github workflows for MSlice

Fixes https://github.com/mantidproject/mslice/issues/594.


setup.py
```diff
@@ -91,12 +92,12 @@ def get_data_files():
 
 # Startup scripts - these use the mantidpython wrappers so we cannot
 # go through the entry_points mechanism
-scripts = ['scripts/start_mslice.py']
-if os.name == 'nt':
-    scripts.append('scripts/mslice.bat')
-else:
-    scripts.append('scripts/mslice')
-setup_args['scripts'] = scripts
+#scripts = ['scripts/start_mslice.py']
```
Pasarus, line 95: Is this meant to be commented out, and could we delete it instead?

Pasarus (changes requested):

setup.py
SilkeSchomann: That is a good question. If we are not installing MSlice standalone outside a conda environment anymore it can be just deleted. Otherwise I will probably have to remove the comments and only run this section if setup.py is not used within a conda environment.

SilkeSchomann: I've removed the commented out lines for now.

Pasarus (approved):

PR #743, opened 2022-05-18
Changed default size for Bragg peaks on interactive plots

### LINKED ISSUE #735: Lines for Bragg peaks to small in interactive cut plots
On interactive cut plot the lines for Bragg peaks are almost not visible. It would be better if they had a similar size to the ones displayed on cut plots made via the cut tab.

SilkeSchomann:
The lines for Bragg peaks were to small on interactive cuts due to a hardcoded scale factor. This scale factor was increased to a more suitable size.

**To test:**

Add a Bragg peak to an interactive cut and a normal cut plot. They should have similar sizes.

Fixes #735

### DIFF
--- a/mslice/presenters/cut_plotter_presenter.py
+++ b/mslice/presenters/cut_plotter_presenter.py
@@ -107,7 +107,7 @@ class CutPlotterPresenter(PresenterUtility):
             scale_fac = np.nanmax(ws_handle.get_signal()) / 10
         except KeyError:
             # Workspace is interactively generated and is not in the workspace list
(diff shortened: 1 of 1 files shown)


SilkeSchomann:
Yes, that is ok. For normal cuts the line for the Bragg peak is calculated relative to the highest peak but that would be too difficult for interactive cuts so it is using a default value. In the past this default value was so small that you could hardly see the Bragg peaks. I think now they are clearly visible.

robertapplin (approved):

PR #744, opened 2022-05-19
Replaced error message for empty cut axes

### LINKED ISSUE #734: Error message for trying to plot empty cut plot not helpful
At the moment the error message that occurs when attempting to plot an empty cut plot is 'could not convert string to float:"'.
This error message should be improved.

SilkeSchomann:
Replaced the error message displayed when attempting to plot a cut without specifying cut axes.

**To test:**

Open a data set in MSlice and navigate to the Cut tab. Without entering any values, click on Plot and check the error message displayed on the bottom of the MSlice window.

Fixes #734


robertapplin (approved):
I can confirm this is working as expected

Issue #731, opened 2022-05-16
Crash when setting axes to log scale 


SilkeSchomann:
**Describe the bug**
When setting axes to log scale MSlice crashed. For instance, when changing the scale setting in the plot options for a cut plot the following error message was displayed on Windows (MSlice as interface of Mantid nightly executable):
`Traceback (most recent call last):
 File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\presenters\plot_options_presenter.py", line 97, in get_new_config
 self._model.change_axis_scale(self._xy_config)
 File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 176, in change_axis_scale
 min = get_min(ydata, absolute_minimum=0.)
 File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 29, in get_min
 mask = np.greater(data, absolute_minimum)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

When changing the scale by double-clicking on the y-axis the error message was different:
`Traceback (most recent call last):
  File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\presenters\quick_options_presenter.py", line 30, in <lambda>
    view.ok_clicked.connect(lambda: _set_axis_options(view, target, model, has_logarithmic, grid))
  File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\presenters\quick_options_presenter.py", line 51, in _set_axis_options
    setattr(model, target[:-5] + 'log', view.log_scale.isChecked())
  File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 457, in y_log
    self.change_axis_scale(config)
  File "C:\MantidNightlyInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 178, in change_axis_scale
    current_axis.set_yscale('symlog', linthreshy=pow(10, np.floor(np.log10(min))))
  File "C:\MantidNightlyInstall\bin\lib\site-packages\matplotlib\axes\_base.py", line 4117, in set_yscale
    ax.yaxis._set_scale(value, **kwargs)
  File "C:\MantidNightlyInstall\bin\lib\site-packages\matplotlib\axis.py", line 761, in _set_scale
[...]

PR #745, opened 2022-05-19
Fix Crash when setting axes to log scale

MialLewis:
#REF731

Description of work.

This fixes the crash that occurs when setting axes to log scale.

This was caused my a deprecation of the `linthreshx/y` kwarg in matplotlib 3.3.

Note that issue #731 refers to two other bugs when using the plot options and when setting the color bar to log scale with a negative min value:
    - For the bug when using plot options, I cannot replicate this exactly. I get the same error message as when changing via the    
      quick options (which this PR fixes). I have tried to replicate this using the mantid nightly build, a local mslice build using an 
      older version of matplotlib, and a conda build of mslice using an updated version of matplotlib
    - The colorscale bug occurs only when setting the max value as a negative when using a log scale. This is a seperate issue, so 
      will be addressed as such.

**To test:**
NOTE: These tests must be completed using a build with matplotlib version < 3,3 and another >= 3.3.

-In the Workspace Manager tab select the workspace MAR21335_Ei60meV
-Navigate to the Cut tab
-In the row labelled along, set the from value to 0 and the to value to 10
-In the row labelled over, set the from value to -5 and the to value to 5
-Click Plot. A new window with a cut plot should open.

-Right click on y axis to open quick options.
-Check the Logarithmic checkbox. 
-Press ok, observed no crash and correct behaviour.
-Repeat for x axis.

-Open Plot Options using the cog symbol on the toolbar.
-For X Axis, check the Logarithmic checkbox.
-Press ok, observed no crash and correct behaviour.
-Repeat for Y axis.

-Check effected unit tests pass.

Partially Fixes 731, will reference and close this issue when in a subsequent PR.

### DIFF
--- a/mslice/scripting/helperfunctions.py
+++ b/mslice/scripting/helperfunctions.py
@@ -135,7 +135,7 @@ def add_cut_plot_statements(script_lines, plot_handler, ax):
                             f"linthresh{x_axis_str}=pow(10, np.floor(np.log10({plot_handler.x_axis_min}))))\n")
 
     if plot_handler.is_changed("y_log"):
-        script_lines.append(f"ax.set_xscale('symlog', "
+        script_lines.append(f"ax.set_yscale('symlog', "
                             f"linthresh{y_axis_str}=pow(10, np.floor(np.log10({plot_handler.y_axis_min}))))\n")
 
 def add_cut_lines(script_lines, plot_handler, ax):
--- a/mslice/scripting/helperfunctions.py
+++ b/mslice/scripting/helperfunctions.py
@@ -123,12 +123,8 @@ def add_cut_plot_statements(script_lines, plot_handler, ax):
     add_cut_lines(script_lines, plot_handler, ax)
     add_plot_options(script_lines, plot_handler)
 
-    if LooseVersion(mpl_version) < LooseVersion('3.3'):
-        x_axis_str = "x"
-        y_axis_str = "y"
-    else:
-        x_axis_str = ""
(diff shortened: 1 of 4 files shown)


robertapplin:
Tested this for matplotlib > 3.3 and it works!

mslice/scripting/helperfunctions.py
```diff
@@ -122,13 +124,20 @@ def add_cut_plot_statements(script_lines, plot_handler, ax):
...
 
     if plot_handler.is_changed("y_log"):
-        script_lines.append("ax.set_yscale('symlog', linthreshy=pow(10, np.floor(np.log10({}))))\n".format(
-            plot_handler.y_axis_min))
-
+        if LooseVersion(mpl_version) < LooseVersion('3.3'):
+            script_lines.append("ax.set_yscale('symlog', linthreshy=pow(10, np.floor(np.log10({}))))\n".format(
+                plot_handler.y_axis_min))
+        else:
+            script_lines.append("ax.set_yscale('symlog', linthresh=pow(10, np.floor(np.log10({}))))\n".format(
```
robertapplin, line 139: This is picky, but it would be good to modernise and use f-strings instead of the format function. Same for in the test file as well

robertapplin (changes requested):
This works for matplotlib < 3.3 as well. I just have one minor suggestion and then I'm happy to approve

robertapplin (approved):
Thanks, looks good to go

PR #749, opened 2022-05-20
Redraw Bragg peaks when axes flipped in interactive cut

### LINKED ISSUE #740: Bragg peaks are removed when flipping axes on interactive cut plots
When flipping the axes of interactive cut plots with Bragg peaks, the Bragg peaks are removed. They should still be present though and after flipping the axes twice the plot should be identical to the original plot.

SilkeSchomann:
For interactive cuts the enabled Bragg peaks now remain checked when flipping the axes and are redrawn on the new plot.

**To test:**

Add a Bragg peak to an interactive cut and flip the axes a couple of times. The Bragg peak should remain enabled and the corresponding lines be adapted to the current plot.

Fixes #740


robertapplin (approved):
Its working as expected and the code looks good

PR #756, opened 2022-05-24
Colour scheme is now consistent within MSlice

### LINKED ISSUE #733: Colour scheme changes after modifying plot options
**Describe the bug**
The colour scheme of MSlice is not consistent. After a change in the line settings via the plot options menu the colours of a cut plot are slightly different. 

**To Reproduce**
Open a cut plot with several lines and change a line setting, for instance 'Show Legends'.

**Expected behavior**
Preferred would be a consistent colour scheme for MSlice.

**Additional information**
There used to be a bug regarding the colour scheme (https://github.com/mantidproject/mslice/issues/550).

SilkeSchomann:
Instead of using the colour name directly in set_option now the name_to_color function from mslice.models.colors is used. This ensures that a colour name is always mapped to the same hex value. The hex mapping for basic colours has been updated as well to reflect the colour scheme changes within matplotlib. The hex values are now hardcoded instead of generated dynamically on start up to speed up plotting. It is still possible to use colours that are not included in the basic colour mapping though.

**To test:**

Create a cut plot with several lines and change a line setting, for instance 'Show Legends'. After clicking on Ok the colours of the lines should not change.

Also, try to re-test https://github.com/mantidproject/mslice/issues/550 as this should still be fixed.

Fixes #733

### DIFF
--- a/mslice/models/colors.py
+++ b/mslice/models/colors.py
@@ -19,6 +19,7 @@ the string cyan
 from __future__ import (absolute_import, division)
 
 from matplotlib import rcParams
+
 from six import iteritems
 try:
     from matplotlib.colors import to_hex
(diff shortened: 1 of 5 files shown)


MialLewis (approved):
Tested, all working including hex mapped and non-hex mapped colors. Previous fix for #550 still works.

Issue #644, opened 2021-12-01
ValueError caused by function get_min


SilkeSchomann:
Calling get_min in mslice\plotting\plot_window\cut_plot.py can cause the following _ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()_

This is the complete stack trace:

```
D:\programs\MantidInstall\scripts\ExternalInterfaces\mslice\presenters\quick_options_presenter.py", line 27, in <lambda>
   view.ok_clicked.connect(lambda: _set_axis_options(view, target, model, has_logarithmic, grid))
 File "D:\programs\MantidInstall\scripts\ExternalInterfaces\mslice\presenters\quick_options_presenter.py", line 47, in _set_axis_options
   setattr(model, target[:-5] + 'log', view.log_scale.isChecked())
 File "D:\programs\MantidInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 394, in y_log
   self.change_axis_scale(config)
 File "D:\programs\MantidInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 144, in change_axis_scale
   min = get_min(ydata, absolute_minimum=0.)
 File "D:\programs\MantidInstall\scripts\ExternalInterfaces\mslice\plotting\plot_window\cut_plot.py", line 19, in get_min
   mask = np.greater(data, absolute_minimum)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

PR #761, opened 2022-05-26
Fix ValueError caused by function get_min

MialLewis:
Description of work.

Previously, if the axis of a cut plot with bragg peaks was changed to logarithmic the `get_min` function throws a value error.

This is because the bragg preak data has a differing number of elements to the spectra data - something the `np.greater` function used can't handle.

The function now performs the operation row by row to alleviate this issue.

**Note:** As part of this PR I've removed the suppression of warnings in the `get_min` function, as I can't see any reason for it.

**Note 2:** This PR does not change the way the coordinates of Bragg peaks are specified. As such, when the y axis is set to log the peaks look like this:
image

I suspect we would rather they look like this?
image

If so I will create a subsequent PR/Issue.

**To test:**
1) Load any data
2) Plot a cut
3) Add Bragg peaks
4) Change x or y scale to logarithmic
5) Observe no error

<!-- Replace #644

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -24,10 +24,8 @@ from mslice.util.compat import legend_set_draggable
 
 def get_min(data, absolute_minimum=-np.inf):
     """Determines the minimum value in a set of numpy arrays (ignoring values below absolute_minimum)"""
-    with warnings.catch_warnings():
-        warnings.simplefilter('ignore')
-        mask = np.greater(data, absolute_minimum)
-    return np.min(np.extract(mask, data))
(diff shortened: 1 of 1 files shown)


MialLewis:
@SilkeSchomann What do you think re. note 2?

SilkeSchomann:
@MialLewis Yes, I think the lines for the Bragg peaks should be modified to keep it consistent with other plots.

MialLewis:
> @MialLewis Yes, I think the lines for the Bragg peaks should be modified to keep it consistent with other plots.

I've created a separate issue for this, so will update on a subsequent PR

robertapplin (approved):
There is no longer and error and code looks good

PR #762, opened 2022-05-26
Change formatting of start up warning re change to cut by integration

MialLewis:
Description of work.

Previously  the warning upon start-up  regarding the change of default cut to cut by integration was formatted to look nice in the console in pycharm. When printed to the mantid console, it does not look right at all.

I've updated the formatting to correct this.

**To test:**
Open mslice via mantid, observe formatting of output warning looks ok.


SilkeSchomann (approved):
Looks fine both in Mantid and standalone 👍

PR #773, opened 2022-05-31
Enable anaconda upload

SilkeSchomann:
Replaced local artefacts with upload to anaconda for both main and nightly MSlice conda packages.

This is a maintenance task and not related to the release so it should go directly into main.


martyngigg (approved):
Thanks for adding the common action. It's good to consolidate that code and avoid duplication

PR #775, opened 2022-06-06
Fix scaling of Bragg peaks for log scales

### LINKED ISSUE #766: Bragg peaks should adjust when y axis is changed to log.
**Describe the bug**
Currently when bragg peaks are plotted on graph for which the y scale is log, the coordinates of the peaks do not adjust. They therefore appear stretched as per the photo below.

**To Reproduce**
1) Plot cut plot
2) Add Bragg Peaks
2) Change the y axis to logarithmic

**Expected behavior**
The brag peaks should scale to overlay the spectra aesthetically (so as effectively still being plotted on a linear axis).

**Screenshots**
image

MialLewis:
**Description of work:**
Previously, the y-coords of Bragg peak lines (which have no physical significance) would not scale when the y axis is set to logarithmic. This means the lines become stretched, not appearing unobtrusive markers as intended (see image on issue).

Through this PR, the Bragg peak lines are now scaled to take up a proportion of the axis (currently hardcoded as 15%) either side of a provided datum (set as the median value of the spectra). As such, they also are refreshed (removed and replotted) every time a change is made to the y axis.

**To test:**

1. Plot cut plot
2. Add Bragg Peaks
3. Change the y axis to logarithmic
4. Change the y axis to reflect the following scenarios:
   - Negative y values
   - Negative y minimum
   - 0 minimum
   - 0 maximum
5. Confirm behaviour of recoil lines and Bragg peak lines on slice plot unaffected.
6. Confirm when waterfall toggle is on, the Bragg peaks are plotted over the first line regardless of offset.
7. Confirm Bragg peaks refresh upon waterfall toggle, plotting of new lines, and removal of existing lines (actions which effect the datum).

Fixes #766

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -323,13 +323,13 @@ class CutPlot(IPlot):
 
     def update_bragg_peaks(self):
         if self.plot_window.action_aluminium.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Aluminium', False, cif=None)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Aluminium', False, None, self.y_log)
         if self.plot_window.action_copper.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Copper', False, cif=None)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Copper', False, None, self.y_log)
         if self.plot_window.action_niobium.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Niobium', False, cif=None)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Niobium', False, None, self.y_log)
         if self.plot_window.action_tantalum.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Tantalum', False, cif=None)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Tantalum', False, None, self.y_log)
         self.update_legend()
 
     def save_icut(self):
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -185,7 +185,6 @@ class CutPlot(IPlot):
 
             if ymin > 0:
                 xy_config['y_range'] = (ymin, xy_config['y_range'][1])
-
         else:
             current_axis.set_yscale('linear')
         self.x_range = xy_config['x_range']
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -321,15 +321,23 @@ class CutPlot(IPlot):
     def is_icut(self):
         return self._is_icut
 
+    def _get_overplot_datum(self):
+        return np.nanmedian([line.get_ydata() for line in self._canvas.figure.gca().get_lines()
+                             if not self._cut_plotter_presenter.is_overplot(line)])
+
     def update_bragg_peaks(self):
         if self.plot_window.action_aluminium.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Aluminium', False, None, self.y_log)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Aluminium', False, None, self.y_log,
+                                                          self._get_overplot_datum())
         if self.plot_window.action_copper.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Copper', False, None, self.y_log)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Copper', False, None, self.y_log,
+                                                          self._get_overplot_datum())
         if self.plot_window.action_niobium.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Niobium', False, None, self.y_log)
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Niobium', False, None, self.y_log,
+                                                          self._get_overplot_datum())
         if self.plot_window.action_tantalum.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Tantalum', False, None, self.y_log)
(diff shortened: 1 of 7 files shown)


mslice/plotting/plot_window/cut_plot.py
```diff
@@ -321,15 +325,28 @@ def set_is_icut(self, is_icut):
     def is_icut(self):
         return self._is_icut
 
-    def update_bragg_peaks(self):
+    def _get_overplot_datum(self):
+        return np.nanmedian([line.get_ydata() for line in self._canvas.figure.gca().get_lines()
+                             if not self._cut_plotter_presenter.is_overplot(line)])
+
+    def update_bragg_peaks(self, refresh=None):
+        refresh = False if refresh is None else True
```
robertapplin, line 333: Why not just set `refresh=False` by default?

mslice/presenters/cut_plotter_presenter.py
```diff
@@ -115,9 +135,13 @@ def add_overplot_line(self, workspace_name, key, recoil, cif=None):
             q_axis = cache.cut_axis
         x, y = compute_powder_line(workspace_name, q_axis, key, cif_file=cif)
         try:
-            y = np.array(y) * (scale_fac / np.nanmax(y))
+            if not y_has_logarithmic:
+                y = np.array(y) * scale_fac / np.nanmax(y) + datum
+            else:
+                y = self._get_log_bragg_y_coords(len(y), 0.15, datum)
```
robertapplin, line 141: Can the `0.15` be moved out of this function and become a 'named variable' at the top of the file. For instance something like `BRAGG_SIZE_ON_AXES = 0.15`?

mslice/presenters/cut_plotter_presenter.py
```diff
@@ -97,8 +97,28 @@ def hide_overplot_line(self, workspace, key):
...
+    def _get_log_bragg_y_coords(size, portion_of_axes, datum):
+        datum = 0.001 if datum == 0 else datum
+        y1, y2 = plt.gca().get_ylim()
+        if (y2 > 0 and y1 > 0) or (y2 < 0 and y1 < 0):
+            total_steps = np.log10(y2 / y1)
+        elif y1 < 0:
+            y1_int = -1
+            y2_int = 1
+            total_steps_up = np.log10(y2 / y2_int) + 1 if abs(y2) >= 1 else abs(y2)
+            total_steps_down = np.log10(y1 / y1_int) + 1 if abs(y1) >= 1 else abs(y1)
```
robertapplin, line 110:
```suggestion
            total_steps_up = np.log10(y2) + 1 if abs(y2) >= 1 else abs(y2)
            total_steps_down = np.log10(-y1) + 1 if abs(y1) >= 1 else abs(y1)
```

mslice/plotting/plot_window/cut_plot.py
```diff
@@ -321,15 +325,28 @@ def set_is_icut(self, is_icut):
...
         if self.plot_window.action_niobium.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Niobium', False, cif=None)
+            refresh and self._cut_plotter_presenter.hide_overplot_line(None, 'Niobium')
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Niobium', False, None, self.y_log,
+                                                          self._get_overplot_datum())
         if self.plot_window.action_tantalum.isChecked():
-            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Tantalum', False, cif=None)
+            refresh and self._cut_plotter_presenter.hide_overplot_line(None, 'Tantalum')
+            self._cut_plotter_presenter.add_overplot_line(self.ws_name, 'Tantalum', False, None, self.y_log,
+                                                          self._get_overplot_datum())
```
robertapplin, line 349: `self._get_overplot_datum()` looks like it could be expensive if its called for each of the bragg peaks. Do you think doing this calculation once, above the if statements, would be worth doing (even when none of the bragg peaks are checked)?

robertapplin (changes requested):
Its working as described so nice work. I just had a few code suggestions but feel free to push back

mslice/presenters/cut_plotter_presenter.py
MialLewis: Yep, good suggestion

mslice/presenters/cut_plotter_presenter.py
MialLewis: Good point, though out of interest would you still follow what this bit of code is trying to do after the change?

mslice/plotting/plot_window/cut_plot.py
robertapplin: Sounds good, lets just leave it as is then 👍

mslice/plotting/plot_window/cut_plot.py
robertapplin: This sounds like good advice for mutable objects and functions as it says in the article, but I would say that it has no benefit for things like booleans

mslice/plotting/plot_window/cut_plot.py
MialLewis: We could cache the median, and recalculate only upon line removal/addition. I initially thought this would be too much of a pain, but thinking of this makes we realise we probably need to refresh the bragg peaks at this point anyway as the datum will change.

mslice/presenters/cut_plotter_presenter.py
robertapplin: For me, I would probably find it easier to understand without the `_int` variables

mslice/plotting/plot_window/cut_plot.py
MialLewis: Point taken, perhaps I should have read more than the first two paragraphs of the article!

robertapplin:
Looks like there is a merge conflict @MialLewis

robertapplin:
A found a bug and a crash, (prior to #784 being merged)
1. Load data and do a cut plot between 0 and 10, -5 and 5
2. Change the y axis to logarithmic, and y min to 0
3. Turn on Copper Bragg peak
4. Do an Overplot
5. Turn on waterfall, and set y offset to 100
6. **You can see the first bug which is that the error bars are not offset for the second line**
7. Tick copper Bragg peak again (which shouldn't be allowed but is fixed by #784)
8. **Tick Aluminium Bragg peak, crash!**

I then fetched release-next and rebased it on this branch after merging #784 . I then did this test again, and its all fixed! So I am happy this is fixed

PR #779, opened 2022-06-07
Add unit test for comparing generated python files to clipboard

### LINKED ISSUE #653: Add unit test to compare scripts generated to file vs to clipboard
This is to compare scripts generated to file with scripts generated to clipboard. They should be identical. 

If this is covered by a unit test we can focus on one method of generating a script and its functionality during manual testing instead of comparing both as well.

robertapplin:
**Description of work:**
This PR adds a test which compares the results of generating a python script to file, and generating a python script to clipboard. See the reasoning in the issue.

**To test:**
Code review and make sure tests pass

Fixes #653

### DIFF
--- a/mslice/scripting/__init__.py
+++ b/mslice/scripting/__init__.py
@@ -12,9 +12,9 @@ def generate_script(ws_name, filename=None, plot_handler=None, window=None, clip
             path = path[0]
         if not path:
             return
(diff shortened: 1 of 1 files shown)


SilkeSchomann (approved):
Code looks good and test works fine

PR #781, opened 2022-06-08
Remove redundant code

robertapplin:
**Description of work:**
This removes some redundant code which is no longer used and should have been removed in a previous PR

**To test:**
Check for Mial's hard crash


MialLewis (approved):
Related functionality still working after the code removal

PR #786, opened 2022-06-08
Update README.md with conda pycharm instructions

mducle:
**Description of work:**

Adds some instructions to the README to work with the new Mantid conda packages in PyCharm.

**To test:**

Check that the instructions work for you (I've only tested on Windows).

No associated issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ First install Mantid using `conda create -n mantidnightly -c conda-forge -c mant
 then add this interpreter by going to the `File->Settings` in PyCharm, then `Project: mslice -> Python Interpreter`,
 click the cog on the right side to add an existing interpreter and select `Conda` and `Python 3.8 (mantidnightly)`.
 Then go to `Run -> Edit Configurations` and create new configuration with this interpreter.
-Specify `start_mslice.py` as the startup script as above, but you do not need to set any environment variable.
+Specify `start_mslice.py` as the startup script set the same `QT_QPA_PLATFORM_PLUGIN_PATH` environment variable as above.
 To run tests, create a `Nosetests` configuration and specify the `Target` as `Custom` with `mslice.tests`
 with the working directory being the mslice package folder (e.g. `<mslice_root>/mslice`).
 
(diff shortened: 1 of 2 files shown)


SilkeSchomann (changes requested):
I've followed the instructions but had to make two small modifications to run or debug the tests:
In the `Nosetests` configuration I had to set the environment variable QT_QPA_PLATFORM_PLUGIN_PATH to the qt5 plugin of my environment (C:\Users\eoc57742\Miniconda3\envs\mantidnightly\Library\plugins\qt5).
In mslice/models/workspacemanager/workspace_algorithms.py I had to add import mslice.util.mantid.init_mantid. Should the second modification just get changed in general for MSlice?

mducle:
> In the Nosetests configuration I had to set the environment variable QT_QPA_PLATFORM_PLUGIN_PATH to the qt5 plugin of my environment (C:\Users\eoc57742\Miniconda3\envs\mantidnightly\Library\plugins\qt5).

Ah right... it's weird I don't have to do that. I'm using mamba-forge but that shouldn't make a difference. Oh well, I'll put this change in.

> In mslice/models/workspacemanager/workspace_algorithms.py I had to add `import mslice.util.mantid.init_mantid`. Should the second modification just get changed in general for MSlice?

Oh yeah, I had to do that on Windows too (but strangely not on Linux... :shrug: 

I think it's safe enough to add it in general.

SilkeSchomann:
Maybe we should add a comment to check for `import mslice.util.mantid.init_mantid` in mslice/models/workspacemanager/workspace_algorithms.py for users with an older version of MSlice.

mducle:
So I think in normal usage this gets imported in either `mslice.app` or `mslice.cli` so users won't get this error unless they import `workspace_algorithms` directly (rather than through the documented `mslice.cli`). It's just that for some reason nosetest on Windows does a direct import whereas on Linux it doesn't... [sometimes coding feels like magic...]

SilkeSchomann (approved):
This works for me

PR #794, opened 2022-06-17
Fix bug that causes continual calls to "toggle_waterfall"

### LINKED ISSUE #793: Error caused when closing cut plot with bragg peaks and waterfall plot
**Describe the bug**
Plotting a cut plot with a waterfall and bragg peaks causes error upon close.

**To Reproduce**
1. Plot a Cut
2. Overplot another cut
3. Set an offset
4. Add a bragg peak
5. Close the plot
6. Observe crash

**Expected behavior**

**Screenshots**
If applicable/possible, add screenshots to help explain your problem.

MialLewis:
**Description of work:**
A bug was caused by the waterfall x and y edit boxes never losing focus on a cut plot. Once either box was selected every subsequent action caused by clicking on a menu/plot takes focus away from the box, but then seems to gives it back again as  the clicked object has `qt focus policy` set to disabled by default. This causes an unnecessary signal and subsequent call to `toggle_waterfall` upon each action.

Upon closing a cut plot, the erroneous signal is emitted in this way, but the `toggle_waterfall` function called after the plot has been closed.

This conflicts with a recent change which requires the `toggle_waterfall` function to update the bragg peak lines when called (this may be unnecessary, considering removal in subsequent PR),

This PR ensures that the waterfall edit boxes lose focus upon finish editing, and closing the cut plot window.

**To test:**
Follow instructions on the issue #793 
Observe how many times `toggle_waterfall` is called when waterfall plots are active, and navigating around making other changes to the cut plot (this is best done with a print statement in `toggle_waterfall` as a break point there prevents a lot of menu functionality).

Fixes #793

### DIFF
--- a/mslice/plotting/plot_window/cut_plot.py
+++ b/mslice/plotting/plot_window/cut_plot.py
@@ -83,8 +83,8 @@ class CutPlot(IPlot):
         plot_window.action_gen_script.triggered.connect(self.generate_script)
         plot_window.action_gen_script_clipboard.triggered.connect(lambda: self.generate_script(clipboard=True))
         plot_window.action_waterfall.triggered.connect(self.toggle_waterfall)
-        plot_window.waterfall_x_edt.editingFinished.connect(self.toggle_waterfall)
-        plot_window.waterfall_y_edt.editingFinished.connect(self.toggle_waterfall)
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
Confirmed that it works as expected.

PR #796, opened 2022-06-17
Fix bragg peak qops to work with new colour mapping and line width

MialLewis:
**Description of work:**

**To test:**
Line quick options (specifically the functions that deal with Bragg peaks) had not been updated for the new colour mapping or line width changes.

This PR updates them to not expect an int, and adds default Bragg peak colours to the colourmap,

1. Plot a cut
2. Add Copper Bragg peaks
3. Right Click on Bragg peak lines
4. Observe correct colour and line width in boxes
5. Click ok, no error

No Issue


SilkeSchomann (approved):
👍

PR #805, opened 2022-07-08
disable failing cut algo tests

MialLewis:
**Description of work:**

**To test:**

Fixes #xxxx.

### DIFF
--- a/mslice/tests/cut_algorithm_test.py
+++ b/mslice/tests/cut_algorithm_test.py
@@ -29,7 +29,7 @@ class CutAlgorithmTest(TestCase):
     def tearDown(self) -> None:
         AnalysisDataService.clear()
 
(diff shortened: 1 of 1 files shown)


sf1919:
Quick question - if the cut algorithm tests have been added to the omitted list of tests (assuming thats what the change to ```.coveragerc``` does), why do you need to add x to the start of some of the test names?

MialLewis:
> Quick question - if the cut algorithm tests have been added to the omitted list of tests (assuming thats what the change to `.coveragerc` does), why do you need to add x to the start of some of the test names?

That omitted list just stops `coverage` from checking that the files are covered by unit tests, as opposed to stopping the tests from running themselves (the `x` does that).

sf1919 (approved):
Now that I understand how the tests are being disabled and have been able to check that the number of tests have dropped from 470 to 466 then I'll approve this.

PR #813, opened 2022-08-05
Fix second slice plot opening when changing intensity with icut open

### LINKED ISSUE #807: Second slice plot erroneously opens whilst changing intensities during icut
**Describe the bug**
When changing intensities during an icut, a second slice window can be erroneously opened.

**To Reproduce**
1. Open mslice dataset `MAR21335_Ei60meV`
2. Display Slice
3. Toggle Interactive Cuts
4. Draw region using rectangle selector
5. Change intensity to `chi"(Q,E)` using the `Intensity` dropdown
6. Enter sample temperature of `100`, press ok
7. Observe second slice opened with adjusted intensity, functionality of first slice opened broken

**Expected behaviour**
Should not be possible to open a second slice - second slice should overrite first.

 - MSlice from Mantid 6.4.0

MialLewis:
**Description of work:**
Previously, if changing intensity with `interactive cut` toggled on and an `icut ` open, a second slice plot would appear and the rectangle selector would disappear. 

Also, when changing intensity prior to the plotting of an `icut`, the rectangle selector and CrossCursor will appear upon interacting with the plot.

This PR resolves these issues.

NOTE: It is a minor annoyance that the rectangle selector flickers upon a change of intensity, not sure how to get rid of this so let me know if you have a solution.

**To test:**

_Second Plot Window_
1.  Plot any slice
2.  Toggle `interactive cut` on the `slice plot` window
3.  Use the rectangle selector to plot an `icut`
4.  Change the intensity type use the intensity drop down menu of the `slice plot` window.
5. Observe the current slice plot intensity is corrected and that the rectangle selector remains in place.

_Disappearing Slice_
1. Plot any slice
2. Toggle` interactive cut` on the `slice plot`.
3. Change the intensity type use the intensity drop down menu of the `slice plot` window.
4. Use the rectangle selector to take an `icut`
5. Observe CrossCursor remains, and that drawn rectangle is visible

Fixes #807 and #812.

### DIFF
--- a/mslice/plotting/globalfiguremanager.py
+++ b/mslice/plotting/globalfiguremanager.py
@@ -485,6 +485,10 @@ class GlobalFigureManager(object):
         cls._disable_make_current = False
 
     @classmethod
+    def make_current_disabled(cls):
+        return cls._disable_make_current
+
+    @classmethod
     def all_figure_numbers(cls):
         """An iterator over all figure numbers"""
         return list(cls._figures.keys())
--- a/mslice/plotting/plot_window/interactive_cut.py
+++ b/mslice/plotting/plot_window/interactive_cut.py
@@ -116,7 +116,11 @@ class InteractiveCut(object):
         self.slice_plot.plot_window.action_interactive_cuts.setChecked(False)
 
     def refresh_rect_selector(self, ax):
+        extents = self.rect.extents
         self.rect = RectangleSelector(ax, self.plot_from_mouse_event,
                                       drawtype='box', useblit=True,
                                       button=[1, 3], spancoords='pixels', interactive=True)
+        if self._rect_pos_cache:
+            self.rect.to_draw.set_visible(True)
+            self.rect.extents = extents
         self.slice_plot.set_cross_cursor()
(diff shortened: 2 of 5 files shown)


SilkeSchomann:
There are a few QT messages appearing when the intensity is changed:

```
QWidget::repaint: Recursive repaint detected
QWidget::paintEngine: Should no longer be called
QPainter::begin: Paint device returned engine == 0, type: 1
QPainter::end: Painter not active, aborted
```
Not sure if this causes the rectangle selector to flicker. I've only noticed this behaviour as it was explained in the PR, otherwise I would probably not even have spotted it!

SilkeSchomann (approved):
Looks good and works as described.

MialLewis:
> There are a few QT messages appearing when the intensity is changed:
> 
> ```
> QWidget::repaint: Recursive repaint detected
> QWidget::paintEngine: Should no longer be called
> QPainter::begin: Paint device returned engine == 0, type: 1
> QPainter::end: Painter not active, aborted
> ```
> 
> Not sure if this causes the rectangle selector to flicker. I've only noticed this behaviour as it was explained in the PR, otherwise I would probably not even have spotted it!

Thanks, will take a look at this.

PR #811, opened 2022-07-28
Add ability to correct intensity of cuts

MialLewis:
**Description of work:**
This change adds the ability to correct the intensity of cuts. This provides the following functionality:
- Intensity corrections made to slice's now follow through to interactive cuts taken from the slice.
- On cut plots the `intensity` menu can now be used to correct intensity as per previous slice functionality.

For consistency, the code pattern used to correct intensity for slice plots has largely been reused for cut plots.

**To test:**
Being wary of existing issues which will be fixed in subsequent PRs (#809, #810), follow steps 1-10 below for the `HYS_63373_4pixel`, `MAR25087_Ei40.00meV` and `MAR21335_Ei60meV` datasets.

**SLICE/ICUT TEST**
1. Load dataset.
2. Calculate projection if pixel workspace, using `DeltaE` as `u1`.
3. Display slice.
4. Correct the intensity of the slice using the `Intensity` drop down. Check each intensity, noting that GDOS is currently disabled and displays an explanatory tooltip. (You can type 100 for use a valid field for the temperature)
5. With the intensity corrected to `chi`, use the `Interactive Cut` feature to plot a cut over a region of interest around 0meV. Ensure the cut is taken along `|Q|`, swap the axis if necessary.
6. Observe that the 'dip' around 0meV apparent on the slice, is also apparent on the `icut`.
8. On the slice, toggle `Interactive Cuts`. This will convert the previously plotted cut from an `icut` to a regular cut plot.
9. On this cut plot, observe that the intensity checked on the `Intensity` drop down menu is the same as that previously selected on the parent slice.
10. Using the `Intensity` drop down menu, cycle through intensity correction options and observe behaviour.
11. Close slice and cut plot

**MULTI CUT TEST**
12. With the `MAR25087_Ei60.00meV` dataset selected, plot a cut along `DeltaE` from -5 to 5, over `|Q|` from 0 to 10, with a width of 2.5.
13. Repeat this for the `MAR25087_Ei40.00meV` dataset, though ensure to `plot over`
14. Correct the intensity to `chi`.
15. Observe that for each dataset loaded, you are prompted to provide a sample temperature.
16. Observe that all cuts on the plot are corrected for intensity.
17. Using the `HYS_63373_4pixel` dataset, plot over a set of cuts along `DeltaE` from -0.5 to 0.5, from 1.7 to 2.
18. Observe that you are prompted to enter a sample temperature for the new dataset.
[...]

PR #823, opened 2022-09-23
Change intensity type logic to use to enums

### LINKED ISSUE #818: Change intensity correction type identifiers to Enums
At the moment in mslice each intensity correction type and associated functions/actions/objects are referred to in multiple ways, eg. `chi`, `dynamical_susceptibility`, `show_dynamical_susceptibility`.

There are multiple functions to switch between these different identifiers and to get the associated functions/actions/objects from them.

It would be good to improve this, using Enums for consistency and providing the ability to get the associated functions/actions/objects from a single place (thinking a static class).

MialLewis:
**Description of work:**
Intensity Correction `Enums` have been added to prevent the need for having "conversation methods" in the current code, and in any extensions going forwards. Respective unit tests have been added,

A few unlisted issues have been fixed:
- Fixed a bug that caused a crash when trying to change intensity having toggled `interactive cut` on, but not yet having selected a rectangle region.
- Fixed a bug on slices generated by scripts, that would cause a crash when trying to change intensity.

**To test:**
Follow testing instructions on this previous PR:
https://github.com/mantidproject/mslice/pull/811

Resolves #818


robertapplin (approved):
The code looks great, happy to approve this. There was a slight issue when trying to open an interactive plot after selected `Chi` in the Intensity menu, but we have confirmed this is an unrelated issue and only seems to happen for Ubuntu 22

PR #825, opened 2022-09-27
Propagate font size in generated scripts from plots

### LINKED ISSUE #795: Font size change on plot is not propagated to generated script
**Describe the bug**
When changing the font size on a plot, for instance the axis title, and then generating a script, the script does not use the new font size.

robertapplin:
**Description of work:**
This PR ensures the title, axis label and colorbar font sizes get propagated when a python script is generated from a plot.

**To test:**
Load data into Mslice
Do a slice plot and cut plot
Change the sizes of the labels
Save the scripts to files
Load these scripts in and run them in Mantid
The font sizes should be propagated

Fixes #795


SilkeSchomann (changes requested):
It works well for labels and the plot title. Font size changes for the tick labels are still ignored though.

SilkeSchomann (approved):
Works fine now :-)

PR #828, opened 2022-10-03
Fix temperature not cached when set from a log/field

### LINKED ISSUE #827: Temperature not cached when set from a log/field
**Describe the bug**
When setting a workspace temperature from a log from a slice, the temperature isn't cached for access from derived cuts.

**To Reproduce**
1) Plot a slice
2) Change intensity to `GDOS`
3) When promoted to select a temperature, choose a valid log from the drop down.
4) Take an interactive cut
5) Observe crash

**Expected behaviour**
No crash

MialLewis:
**Description of work:**

**To test:**

1) Load any `.nxs` mslice dataset (I used `MAR21335_Ei60meV.nxs`)
2) `Display` slice
3) Change intensity to `GDOS` via the `intensity` drop down.
4) When promoted to select a temperature, choose `T_Head` from the drop down.
5) Take an interactive cut
6) Observe no crash

Fixes #827

### DIFF
--- a/mslice/plotting/plot_window/slice_plot.py
+++ b/mslice/plotting/plot_window/slice_plot.py
@@ -338,7 +338,7 @@ class SlicePlot(IPlot):
                 return False
             if field:
                 self._slice_plotter_presenter.add_sample_temperature_field(temp_value_raw)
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
Works fine :-)

PR #838, opened 2022-10-18
Updated Ubuntu version to 20.04 in GitHub workflows

### LINKED ISSUE #837: Replace Ubuntu-18.04 in GitHub workflows
The Ubuntu-18.04 is deprecated and has to be replaced in our workflows.

SilkeSchomann:
**Description of work:**

Changed all GitHub workflows from Ubuntu-18.04 to Ubuntu-20.04.

**To test:**

Ensure that workflows are still running. For most workflows that will be only possible after merging into main though.

Fixes #837

### DIFF
--- a/.github/workflows/deploy_conda.yml
+++ b/.github/workflows/deploy_conda.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build_conda_and_upload:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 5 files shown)


MialLewis:
Is there no danger that we end up with tests etc, passing on `20.04`, then failing on `18.04` when build on `mantid`? I mainly ask as I have assumed this for the `vesuvio` repo.

MialLewis (approved):
Unit tests work, so I am inferring that the other workflows will work too.

SilkeSchomann:
I would be surprised if they fail as they are running fine on my local Ubuntu 20.04.

PR #846, opened 2022-12-16
MPL 3.6.2 updates

SilkeSchomann:
In Mantid matplotlib has been updated to version 3.6.2. The pyplot version modified for MSlice does not work with matplotlib 3.6.2 so it was replaced by a pyplot based on matplotlib 3.6.2. In addition, the boilerplate script was replaced by the version from matplotlib 3.6.2 and the number of required modifications for MSlice reduced.

There are also minor modifications to the quick options tests and the script generation for waterfall plots so that MSlice can both work with matplotlib 3.6.2 as well as older versions of matplotlib.

**To test:**

- Ensure unit tests are successful
- Code review


robertapplin (approved):
Looks good to me 👍

PR #848, opened 2023-01-06
Allow custom hex color when opening plot options

### LINKED ISSUE #847: MSlice crashes upon opening plot options for a non-supported colour
**Describe the bug**
If using a script (probably old) to generate a cut plot that uses a colour that we don't "support", when opening plot options `mslice` crashes.

**To Reproduce**
Steps to reproduce the behaviour:
1.  Plot any cut
2.  Generate a script to clipboard, paste it back into the `mslice` interface
3.  Change the colour specified in the script to `#9933ff` 
4. Run the script
5. Open plot options, observe error.

**Expected behaviour**
This error should be handled properly.

robertapplin:
**Description of work:**
This PR ensures the `color_to_name` and `name_to_color` will return the passed in string if the string is a valid hex code provided by a user. This avoids raising a value error when the color isn't one of the basic types, but is still a color.

**To test:**
Follow the instructions in the issue to test.
Click `Ok` in the plot options to ensure no error when the users custom color is still selected.

Fixes #847

### DIFF
--- a/mslice/models/colors.py
+++ b/mslice/models/colors.py
@@ -18,6 +18,8 @@ the string cyan
 """
 from __future__ import (absolute_import, division)
 
+import re
+
 from matplotlib import rcParams
 
 from six import iteritems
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
this works quite well and enables the user to select from a much wider range of colours 👍

PR #849, opened 2023-01-10
Update for latest mock version 

SilkeSchomann:
**Description of work:**

We are using the latest version of mock for the unit tests. Mock was upgraded from 5.0.0. to 5.0.1 recently and one of the changes was that calls to `called_once_with` now result in an AttributeError to enforce usage of `assert_called_once_with` instead. 

There are unrelated issues with the mantid and mantidqt nightly at the moment. That is the reason why the unit test workflow uses mantid and mantidqt 6.5.0 at the moment.

**To test:**

Check that the unit tests are running successfully.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install Mantid
         run: |
-          conda install mantid mantidqt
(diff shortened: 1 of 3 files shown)


robertapplin (approved):
Code looks good and the unit tests are passing

PR #852, opened 2023-01-20
Fix for segmentation fault in unit test workflows

SilkeSchomann:
**Description of work:**

Since the mantid/mantidqt upgrade to Qt 5.15 both MSlice unit test workflows failed with segmentation faults. After running the nosetests and coverage successfully the clean up
routine in mantidqt.utils.qt.testing tried to access a QApplication instance that was already closed. This was fixed by removing the start_qapplication decorator that is not 
required in MSlice tests as a QApplication instance is created separately for MSlice tests.

In addition, several components of the GitHub workflows have been upgraded:
- Ubuntu from 20.04 to latest
- actions/checkout from v2 to v3
- conda-incubator/setup-miniconda from v2.1.1 to v2.2.0

To remove deprecation warnings setuptools is now pinned to 47.0.0 and setup.py is now opening files with the 'r' flag instead of 'rU'.

**To test:**

Make sure that the unit tests are passing without a segmentation fault.

### DIFF
--- a/.github/workflows/deploy_conda.yml
+++ b/.github/workflows/deploy_conda.yml
@@ -6,19 +6,19 @@ on:
 
 jobs:
   build_conda_and_upload:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     defaults:
       run:
         shell: bash -l {0}
 
     steps:
       - name: Checkout MSlice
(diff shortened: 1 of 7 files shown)


MialLewis (approved):
Looks good - fixes workflow issue

PR #873, opened 2023-02-27
Improved packaging and configuration

### LINKED ISSUE #857: Improve packaging for MSlice
Packaging with help of Setuptools has changed over the time so that the current approach in MSlice is nowadays outdated. This applies to the project structure and the configuration among others. In particular, pyproject.toml is now the standard way of specifying project metadata and even settings for some of the tools like coverage and flake8.

Aim is to modernize the project structure and configuration of MSlice and clean up unused files.

SilkeSchomann:
**Description of work:**

- Modernized the project structure by moving mslice code into a src folder and moving tests to the same level as src. 
- Moved most of the configuration into a pyproject.toml file (setup.cfg is only used for nosetests from now on). 
- Cleaned up unused files

**To test:**

Check that unit tests are passing. Check that mslicedevel scripts are still working. Once merged into main check the nightly conda package from the next day on and also the online documentation (it should show the version as 2.6 instead of 2.6.0.

Fixes #857


README.md
```diff
@@ -55,8 +55,8 @@ then add this interpreter by going to the `File->Settings` in PyCharm, then `Pro
 click the cog on the right side to add an existing interpreter and select `Conda` and `Python 3.8 (mantidnightly)`.
 Then go to `Run -> Edit Configurations` and create new configuration with this interpreter.
 Specify `start_mslice.py` as the startup script.
```
MialLewis, line 57: Might be worth adding that `src` now needs to be added as a `sources root` or have the working directory at `<mslice root>\src` rather than `<mslice root>` as per the image.

MialLewis (approved):
This change is really good, the package is much improved from where it was ~1 year ago.

I have replicated all of the tests described on Linux and windows to success. 

I still can't get the unit tests to run on the command line in windows, but they all pass using pycharm regardless.

One suggestion for an addition to the readme, though this is pretty obvious so not essential.

README.md
SilkeSchomann: Good point, I will just replace the image.

MialLewis (approved):

PR #874, opened 2023-02-27
Fix IndexError caused by saving a slice as a matlab file

### LINKED ISSUE #830: Saving a slice plot as a matlab file causes an IndexError
**Describe the bug**
An IndexError is thrown when trying to save a slice as a matlab file.

**To Reproduce**
1. Load any data in MSlice
2. Plot a slice 
3. Try saving the slice as a matlab file

MialLewis:
**Description of work:**

**To test:**

Follow instructions in https://github.com/mantidproject/mslice/issues/830.

This PR ensures that properties of a workspace are only propagated if they are not `None`.

Fixes #830


SilkeSchomann (approved):
Code looks good and solves the problem 👍

PR #875, opened 2023-03-01
Check that the sample log provided is valid before plotting

### LINKED ISSUE #809: Unhandled exception when selecting invalid field for sample temperature
**Describe the bug**
When an invalid (from a physics perspective, probably) field is selected, an unhandled error can be thrown. This error should be handled.

**To Reproduce**
1. Load `MAR21335_Ei60meV` dataset (other's likely to reproduce issue)
2. Display slice
3. On the intensity dropdown select `Chi''(Q,E)`
4. Press ok, with the default `DirectInelasticReductionNormalisedBy` in the combo box.
5. Observe unhandled error.

**Expected behavior**
Resultant error should be handled.

 - MSlice Version 6.4.0

robertapplin:
**Description of work:**
This PR fixes an unhandled exception caused when selecting an invalid sample log for the sample temperature. The solution is to check that the sample log returns a valid temperature before attempting to continue with plotting. If it is not valid, and error message is displayed.

**To test:**
Follow the testing instructions in the issue. There should be no exception, but instead a message displayed.

Fixes #809

### DIFF
--- a/src/mslice/plotting/plot_window/slice_plot.py
+++ b/src/mslice/plotting/plot_window/slice_plot.py
@@ -332,7 +332,6 @@ class SlicePlot(IPlot):
 
     def _run_temp_dependent(self, slice_plotter_method, previous):
         temp_value_raw = None
-        temp_value = None
         try:
             slice_plotter_method(self.ws_name)
         except ValueError:  # sample temperature not yet set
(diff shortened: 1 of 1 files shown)


MialLewis:
Bug has already been resolved

PR #876, opened 2023-03-02
Refactor run temp dependant function

MialLewis:
**Description of work:**

Refactors a large function that was getting hard to work with.

Also, I've removed the caching of sample log fields on the `slice_plotter_presenter`. This used to mean that you only had to select a sample log for temperature once, then it would be used for all other subsequent slices. I accidently bypassed this functionality a while ago when adding intensity change for cuts, taking the view that the temperature should be specified for every new dataset introduced.

If users take issue with this, I can reimplement.

**To test:**

### DIFF
--- a/src/mslice/plotting/plot_window/slice_plot.py
+++ b/src/mslice/plotting/plot_window/slice_plot.py
@@ -21,6 +21,7 @@ from mslice.plotting.pyplot import GlobalFigureManager
 from mslice.scripting import generate_script
 from mslice.util.compat import legend_set_draggable
 from mslice.util.intensity_correction import IntensityType, IntensityCache
+from mslice.models.intensity_correction_algs import sample_temperature
 
 DEFAULT_LABEL_SIZE = 10
 DEFAULT_TITLE_SIZE = 12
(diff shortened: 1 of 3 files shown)


src/mslice/presenters/slice_plotter_presenter.py
```diff
@@ -20,18 +19,16 @@ class SlicePlotterPresenter(PresenterUtility):
...
         workspace = get_workspace_handle(selected_ws)
         self.create_slice(workspace, x_axis, y_axis, intensity_start, intensity_end, norm_to_one, colourmap)
         self.plot_from_cache(workspace)
 
     def create_slice(self, selected_ws, x_axis, y_axis, intensity_start, intensity_end, norm_to_one, colourmap):
-        sample_temp = sample_temperature(selected_ws, self._sample_temp_fields)
         norm = Normalize(vmin=intensity_start, vmax=intensity_end)
         slice = compute_slice(selected_ws, x_axis, y_axis, norm_to_one)
-        self._cache_slice(slice, colourmap, norm, sample_temp, x_axis, y_axis)
+        self._cache_slice(slice, colourmap, norm, None, x_axis, y_axis)
```
robertapplin, line 31: Would it be possible to check with the users that caching the sample temperature is no longer wanted before we remove it?

src/mslice/plotting/plot_window/slice_plot.py
```diff
@@ -331,40 +332,43 @@ def update_canvas(self, cbar_range, cbar_log, x_range, y_range, title):
...
+
+        if temp_value is None or temp_value < 0:
+            self.plot_window.display_error("Invalid value entered for sample temperature. Enter a value in Kelvin \
+                                            or a sample log field.")
+            return False
+        else:
+            self.default_options['temp'] = temp_value
+            self.temp = temp_value
+            self._slice_plotter_presenter.set_sample_temperature(self.ws_name, temp_value)
+            return True
```
robertapplin, line 368:
```suggestion
        self.default_options['temp'] = temp_value
        self.temp = temp_value
        self._slice_plotter_presenter.set_sample_temperature(self.ws_name, temp_value)
        return True
```

src/mslice/plotting/plot_window/slice_plot.py
```diff
@@ -331,40 +332,43 @@ def update_canvas(self, cbar_range, cbar_log, x_range, y_range, title):
...
+        try:
+            temp_value_raw, field = self.ask_sample_temperature_field(str(self.ws_name))
+            temperature_found = self._handle_temperature_input(temp_value_raw, field)
+        except RuntimeError:  # if cancel is clicked, go back to previous selection
+            temperature_found = False
+        if not temperature_found:
+            self.set_intensity(previous)
+        return temperature_found
+
+    def _handle_temperature_input(self, temp_value_raw, field):
```
robertapplin, line 354: While we are doing a refactor in this area, could we add type hinting to these new functions?

src/mslice/plotting/plot_window/slice_plot.py
```diff
@@ -331,40 +332,43 @@ def update_canvas(self, cbar_range, cbar_log, x_range, y_range, title):
...
-                        temp_value = float(temp_value)
-                    except ValueError:
-                        temp_value = None
-            if temp_value is None or temp_value < 0:
-                self.plot_window.display_error("Invalid value entered for sample temperature. Enter a value in Kelvin \
-                                                or a sample log field.")
-                self.set_intensity(previous)
+        except ValueError:  # sample temperature not yet set, get it and reattempt method
+            if self._set_sample_temperature(previous):
+                slice_plotter_method(self.ws_name)
```
robertapplin, line 339: I think it is possible for `previous` to be None, and so there could still be an unhandled exception on line 339?

robertapplin (changes requested):
This refactor looks good. I just had a few suggestions and questions. In particular, I think it would be worth checking with a user that the sample temperature doesn't need to be cached before removing this functionality

src/mslice/presenters/slice_plotter_presenter.py
MialLewis: This feature was accidently removed a couple of months ago, so I think we're safe to remove. I'll raise an issue to investigate whether this should be reimplemented. At this point it's more effort to fix than to remove.

src/mslice/plotting/plot_window/slice_plot.py
MialLewis: This cannot be none, its actually the `QAction` related to the previously selected intensity type in the drop down menu. As a menu item is always selected, there is always a previous.

src/mslice/plotting/plot_window/slice_plot.py
MialLewis: With pleasure

src/mslice/presenters/slice_plotter_presenter.py
mducle: Yes, I think it's fine to remove it. In practice different slices often have different temperatures so the caching is not always useful. 

Is `_cache_slice` referenced elsewhere where it needs the temperature? If not, maybe it can be refactored to remove this argument, instead of passing `None`?

Ideally what we could have in future is to cache the name of the log variable which holds the temperature and use that to determine the temperature of the slice. However (1) this was never implemented, and more importantly (2) the `.nxspe` file format we often use discards all the log information :( the reduced/processed `.nxs` files retains them but it's hard to get users to change to using it...

MialLewis: It is not used elsewhere. I think I will leave it as is now, and implement the feature as you outline shortly.

robertapplin (approved):
Thanks for the changes, happy to approve

PR #879, opened 2023-03-15
Replace conda install with mamba install

SilkeSchomann:
**Description of work:**

To speed up the unit test workflows and the deploy action `conda install` was replaced with `mamba install`. Also, with the new packaging is it not necessary anymore to install setuptools in the conda environment for tests.

**To test:**

Check that unit test are running successfully.


MialLewis (approved):
Good change, nice to be able to remove `setuptools` also

PR #880, opened 2023-03-15
Pre commit use `autofix_prs` option

robertapplin:
**Description of work:**
This PR adds the `autofix_prs` option to our pre-commit setup. This allows us to remove `flake8` from our other unit test workflows as the `autofix_prs` option makes sure pre-commit is run, and commits any fixes it makes.

**To do**
- [ ] Add instruction to install `pre-commit` locally (optionally but would be good practise)

**To test:**

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -25,10 +25,6 @@ jobs:
           python-version: 3.8
           auto-activate-base: false
 
-      - name: Flake8
-        run: |
-          python -m flake8
-
       - name: Install Mantid
         run: |
           mamba install -c mantid/label/nightly mantid mantidqt
--- a/.github/workflows/unit_tests_nightly.yml
+++ b/.github/workflows/unit_tests_nightly.yml
@@ -25,10 +25,6 @@ jobs:
           python-version: 3.8
           auto-activate-base: false
 
-      - name: Flake8
-        run: |
-          python -m flake8
-
       - name: Install Mantid
         run: |
           mamba install -c mantid/label/nightly mantid mantidqt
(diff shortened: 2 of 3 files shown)


robertapplin:
It looks like a large number of error codes are excluded from our current flake8. Here the `E2` and `E3` codes will exclude all codes beginning with `E2` and `E3`
https://flake8.pycqa.org/en/latest/user/violations.html#extending-the-default-ignore-list

@SilkeSchomann  Do you know if it is intentional that we exclude all of these? If not, do we want a maintenance issue to look at which of these `E2` and `E3` error codes we might want to include in the future?

SilkeSchomann:
> It looks like a large number of error codes are excluded from our current flake8. Here the `E2` and `E3` codes will exclude all codes beginning with `E2` and `E3` https://flake8.pycqa.org/en/latest/user/violations.html#extending-the-default-ignore-list
> 
> @SilkeSchomann Do you know if it is intentional that we exclude all of these? If not, do we want a maintenance issue to look at which of these `E2` and `E3` error codes we might want to include in the future?

A maintenance issue to check if these should still be included is a good idea. It is probably possible to exclude fewer than these and just fix the code instead.

robertapplin:
> > It looks like a large number of error codes are excluded from our current flake8. Here the `E2` and `E3` codes will exclude all codes beginning with `E2` and `E3` https://flake8.pycqa.org/en/latest/user/violations.html#extending-the-default-ignore-list
> > @SilkeSchomann Do you know if it is intentional that we exclude all of these? If not, do we want a maintenance issue to look at which of these `E2` and `E3` error codes we might want to include in the future?
> 
> A maintenance issue to check if these should still be included is a good idea. It is probably possible to exclude fewer than these and just fix the code instead.

👍 I have opened a maintenance issue here #881

PR #883, opened 2023-03-21
Replace nose with pytest

### LINKED ISSUE #858: Investigate replacement of nosetests by pytest
Nose has not been updated for several years now and is widely seen as outdated. Pytest is actively maintained, has a large user community and is supposed to be easier to use than unittest. 

It would be good to know if the advantages of switching to pytest outweigh the effort of porting our tests from nose to pytest.

robertapplin:
**Description of work:**
This PR replaces the usage of nose with pytest for running the Mslice tests. 

**To do**
- [x] Add instructions for running pytest from command line
- [x] Add instructions for running pytest from pycharm
- [x] Check coverage reporting still works

**To test:**
Make sure tests are run successfully, and the coverage is reported
Grep for references to nosetest

Fixes #858


robertapplin:
I have opened an issue to fix the pytest warnings #884

SilkeSchomann:
Would it be possible to move the settings from pytest.ini and .coveragerc into pyproject.toml? Then all configuration is done with just one file.

robertapplin:
> Would it be possible to move the settings from pytest.ini and .coveragerc into pyproject.toml? Then all configuration is done with just one file.

Thanks, its ready for another look

PR #885, opened 2023-03-22
Eliminate pytest warnings

### LINKED ISSUE #884: Eliminate pytest warnings and enforce zero warning policy
**Describe the bug**
Pytest has uncovered a large amount of warnings when run in our CI, and locally. We should look into fixing these warnings, and then making sure pytest with fail if a new warning appears, so we have a zero warning policy going forward.

**To Reproduce**
Look at the test output for a PR such as this one https://github.com/mantidproject/mslice/actions/runs/4479637121/jobs/7873841486?pr=883

**Screenshots**
image

robertapplin:
**Description of work:**
This PR eliminates all the pytest warnings seen when running pytest. It also ensures that pytest will fail whenever a new warning appears, so that we have a zero warning policy.

**To test:**
Code review
Make sure the tests are passing. Read the output to make sure there are no warnings

Fixes #884


SilkeSchomann (approved):
It is good to see the code getting modernised by this!

PR #892, opened 2023-03-30
Add warning for small step Cut Plots

### LINKED ISSUE #728: Lines missing on cut plots
**Describe the bug**
Occasionally lines are missing in cut plots, for example for the largest peak in the following plot
image

**To Reproduce**

2. Navigate to the Cut tab

The problem does only occur occasionally and usually disappears when plotting for a second time.

**Expected behavior**
The plot above should look like this:
image

MSlice Version 2.3.1

### LINKED ISSUE #842: Lines missing on cut plot with reduced step sizes on pixel workspaces
**Describe the bug**
When plotting a cut from a pixel workspace, when the cut is taken using a step smaller than that of the pixel workspace a number of the lines are missing. The amount of missing lines increases proportional to the difference between the step sizes.

Perhaps we should prevent cuts and slices from being taken that have a higher resolution than the parent dataset - I don't see any reason why you would want to do this?

I think this is possibly related to #728. When you take a slice from a pixel workspace with a higher resolution than that workspace, you get the following:
image

The gaps (N/As) in the data are likely the direct cause of the missing lines - prehaps the missing lines in #728 are also caused by this.

**To Reproduce**
Run the following script, after adjusting the file path to the `HYS_63373_4pixel` dataset.

# Python Script Generated by Mslice on 2022-11-04 10:06:01
```
import mslice.cli as mc

HYS_63373_4pixel = mc.Load(Filename='C:/Users/mlc47243/Documents/mantid-data/HYS_63373_4pixel.nxspe', OutputWorkspace='HYS_63373_4pixel')
ws_HYS_63373_4pixel_QE = mc.MakeProjection(InputWorkspace=HYS_63373_4pixel, Axis1='|Q|', Axis2='DeltaE')

fig = plt.gcf()
fig.clf()
ax = fig.add_subplot(111, projection="mslice")
cut_ws_0 = mc.Cut(ws_HYS_63373_4pixel_QE, CutAxis="DeltaE,-0.5,0.5,0.05", IntegrationAxis="|Q|,0.2,0.3,0.0", NormToOne=False, IntensityCorrection=False, SampleTemperature=None)
ax.errorbar(cut_ws_0, label="HYS_63373_4pixel_QE 0.20<|Q|<0.30", color="#1f77b4", marker="o", ls="-", lw=1.5)

cut_ws_1 = mc.Cut(ws_HYS_63373_4pixel_QE, CutAxis="DeltaE,-0.5,0.5,0.035", IntegrationAxis="|Q|,0.2,0.3,0.0", NormToOne=False, IntensityCorrection=False, SampleTemperature=None)
ax.errorbar(cut_ws_1, label="HYS_63373_4pixel_QE 0.20<|Q|<0.30", color="#ff7f0e", marker="o", ls="-", lw=1.5)

cut_ws_2 = mc.Cut(ws_HYS_63373_4pixel_QE, CutAxis="DeltaE,-0.5,0.5,0.02", IntegrationAxis="|Q|,0.2,0.3,0.0", NormToOne=False, IntensityCorrection=False, SampleTemperature=None)
ax.errorbar(cut_ws_2, label="HYS_63373_4pixel_QE 0.20<|Q|<0.30", color="#2ca02c", marker="o", ls="-", lw=1.5)

from mslice.cli.helperfunctions import hide_a_line_and_errorbars, append_visible_handle_and_label
from mslice.util.compat import legend_set_draggable

# hide lines, errorbars, and legends
handles, labels = ax.get_legend_handles_labels()
visible_handles = []
visible_labels = []

append_visible_handle_and_label(visible_handles, handles, visible_labels, labels, 0)

append_visible_handle_and_label(visible_handles, handles, visible_labels, labels, 1)

append_visible_handle_and_label(visible_handles, handles, visible_labels, labels, 2)

legend_set_draggable(ax.legend(visible_handles, visible_labels, fontsize='medium'), True)

ax.set_title('')
ax.set_ylabel(r'Signal/#Events')
ax.set_xlabel(r'Energy Transfer (meV)')
ax.set_ylim(bottom=-0.0010965015655078642, top=0.14673213392107215)
ax.set_xlim(left=-0.5310344827586206, right=0.5310344827586206)
mc.Show()
```

**Expected behavior**
No missing lines.

robertapplin:
**Description of work:**
This PR adds a warning when a step for a cut plot is smaller than the steps in the x axis data of a workspace. This can cause plots to appear to having missing lines. It has been implemented as follows:

- From the cli, a warning message will be output to the log
- From the interface, a warning message will be displayed in the status bar

**To test:**
Follow the instructions in the testing issue
Make sure two warnings are output after running the script, and that they make sense. (two warnings because two of the lines have cut axis steps which are too small)

Repeat the same steps but via the interface. See that a warning is displayed in the status bar.

Fixes #842 
Fixes #728

### DIFF
--- a/src/mslice/cli/helperfunctions.py
+++ b/src/mslice/cli/helperfunctions.py
@@ -80,8 +80,8 @@ def _process_axis(axis, fallback_index, input_workspace, string_function=_string
             return axis
         x_step = get_axis_range(input_workspace, axis.units)[-1]
         if axis.step < x_step:
-            logging.warning(f"The {axis.units} step provided ({axis.step}) is smaller than the data step in "
-                            f"the workspace ({x_step}). Please provide a larger {axis.units} step.")
+            logging.warning(f"The {axis.units} step provided ({axis.step:.4f}) is smaller than the data step in "
+                            f"the workspace ({x_step:.4f}). Please provide a larger {axis.units} step.")
     elif axis in available_axes:
         range = get_axis_range(input_workspace, axis)
         range = list(map(float, range))
--- a/src/mslice/cli/helperfunctions.py
+++ b/src/mslice/cli/helperfunctions.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 
 from mslice.workspace.histogram_workspace import HistogramWorkspace
 from mslice.workspace.base import WorkspaceBase as Workspace
@@ -75,6 +76,12 @@ def _process_axis(axis, fallback_index, input_workspace, string_function=_string
     # check to see if axis is just a name e.g 'DeltaE' or a full binning spec e.g. 'DeltaE,0,1,100'
     if ',' in axis:
         axis = string_function(axis)
+        if fallback_index != 0:
+            return axis
+        x_step = get_axis_range(input_workspace, axis.units)[-1]
+        if axis.step < x_step:
+            logging.warning(f"The {axis.units} step provided ({axis.step}) is smaller than the data step in "
+                            f"the workspace ({x_step}). Please provide a larger {axis.units} step.")
     elif axis in available_axes:
         range = get_axis_range(input_workspace, axis)
         range = list(map(float, range))
--- a/src/mslice/cli/helperfunctions.py
+++ b/src/mslice/cli/helperfunctions.py
@@ -4,7 +4,7 @@ import logging
 from mslice.workspace.histogram_workspace import HistogramWorkspace
 from mslice.workspace.base import WorkspaceBase as Workspace
 from mslice.workspace.workspace import Workspace as MatrixWorkspace
-from mslice.models.alg_workspace_ops import get_axis_range, get_available_axes
+from mslice.models.alg_workspace_ops import get_axis_range, get_axis_step, get_available_axes
 from mslice.models.axis import Axis
 from mslice.models.workspacemanager.workspace_provider import workspace_exists
 from mslice.models.intensity_correction_algs import (compute_chi, compute_d2sigma,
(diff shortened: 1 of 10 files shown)


src/mslice/cli/helperfunctions.py
```diff
@@ -75,6 +76,12 @@ def _process_axis(axis, fallback_index, input_workspace, string_function=_string
     # check to see if axis is just a name e.g 'DeltaE' or a full binning spec e.g. 'DeltaE,0,1,100'
     if ',' in axis:
         axis = string_function(axis)
+        if fallback_index != 0:
+            return axis
+        x_step = get_axis_range(input_workspace, axis.units)[-1]
```
SilkeSchomann, line 81: Instead of using the last element of the tuple you could try using the step element directly ([2]) and catch the error if this is not possible. Just in case get_axis_range does not return the expected tuple...

src/mslice/presenters/cut_plotter_presenter.py
```diff
@@ -92,6 +93,13 @@ def _plot_with_width(self, workspace, cut, plot_over):
             plot_over = True
         cut.reset_integration_axis(cut.start, cut.end)
 
+        cut_axis = cut.cut_axis
+
+        x_step = get_axis_range(workspace, cut_axis.units)[-1]
```
SilkeSchomann, line 98: Same comment as for helperfunctions.py.

SilkeSchomann (changes requested):

SilkeSchomann (approved):
Cuts with too small step sizes now show warnings, both from the interface and as a script.

PR #894, opened 2023-04-06
Prevent axis restriction after intensity change

### LINKED ISSUE #841: Zoom stuck after slice intensity change
**Describe the bug**
If you zoom in on a slice then change intensity, you can not use the tool bar to zoom back out again, having to adjust the axis limits manually.

**To Reproduce**
1. Plot Slice
2. Zoom in via `+`
3. Change intensity via `Intensity` drop down.
4. Attempt to zoom out via `-`

**Expected behavior**
Should be able to zoom back out to the original view. The

**Additional context**

robertapplin:
**Description of work:**
This PR ensures that the axis limits for a slice plot get reset to their original values when the selected intensity is changed. The toolbar history gets cleared inside the matplotlib code when a new intensity is plotted, and so we believe this might be the best option for us currently.

**To test:**
Follow the instructions in the attached issue

Fixes #841

### DIFF
--- a/src/mslice/plotting/plot_window/slice_plot.py
+++ b/src/mslice/plotting/plot_window/slice_plot.py
@@ -308,8 +308,6 @@ class SlicePlot(IPlot):
             self.set_intensity(action)
             cbar_log = self.colorbar_log
             cbar_range = self.colorbar_range
-            x_range = self.x_range
-            y_range = self.y_range
             title = self.title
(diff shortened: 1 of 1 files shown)


SilkeSchomann (approved):
Zooming out automatically when changing the intensity of a plot is more convenient for the user since manually zooming out is not possible anymore after the intensity change.

PR #895, opened 2023-04-11
Fix Bragg peak too large on interactive cuts

### LINKED ISSUE #853: Bragg peaks on interactive cut are too long
**Describe the bug**
The lines for Bragg peaks on interactive cuts are too long. Sometimes the plot looses its zoom capability as well, even if the peaks are removed.

**To Reproduce**
Create interactive cut and add Bragg peaks.

robertapplin:
**Description of work:**
This PR fixes a bug where the Bragg Peak lines displayed on an interactive cut are too large. They default to having 90% the length of the y axis. Instead, we want the bragg peaks to be 10% of the maximum signal - as is done for a direct cut.

Follow the testing instructions in the issue. You should get bragg peaks similar in size to this:
image

Fixes #853

### DIFF
--- a/src/mslice/presenters/cut_plotter_presenter.py
+++ b/src/mslice/presenters/cut_plotter_presenter.py
@@ -192,11 +192,12 @@ class CutPlotterPresenter(PresenterUtility):
         try:
             ws_handle = get_workspace_handle(workspace_name)
             workspace_name = ws_handle.parent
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
The Bragg peaks now have a similar size for interactive cut plots as for direct cuts 👍

PR #896, opened 2023-04-12
Support saving of NXSPE files 

### LINKED ISSUE #540: Add nxspe save format option
At the moment workspaces can be saved as `.nxs`, `.txt` or `.mat` formats.

In order to facilitate communications with other software it would be good if the legacy `.nxspe` format could also be a save option. This format is used by Horace, (Matlab) MSlice and DAVE and is a de-facto standard.

The Mantid `SaveNXSPE` algorithm can be used to write such files.

### LINKED ISSUE #834: Saved nxs from scaled or subtracted workspaces not reloadable
**Describe the bug**
MSlice has the ability to compute scaled or subtracted workspaces using buttons in the "Workspace Manager" tab (or via the CLI). There is also a "Save" button to save these two Nexus / `nxs` format. However, MSlice gives an error if these workspaces are loaded again.

**To Reproduce**
1. Load a datafile
2. Apply a scaling to the loaded workspace
3. Save the scaled workspace as "Nexus"
4. Load the saved `.nxs` file - MSlice will give an error

**Expected behavior**
MSlice should be able to reload the saved workspace and treat it as it would as if that workspace was the result of the processing. This likely means that additional functionality needs to be added to save the required MSlice workspace metadata in order to construct the [`Workspace`](https://github.com/mantidproject/mslice/blob/main/mslice/workspace/workspace.py) wrapper as logs in the underlying Mantid workspace so that they can be saved by `SaveNexus` and reloaded later.

 - MSlice Version 2.4.0 (Mantid nightly 221013.2104)

robertapplin:
**Description of work:**
This PR fixes an issue where the metadata for a workspace based off the NXSPE file format was not being saved correctly. Previously, we were using the `SaveNexus` file to save the NXSPE workspace, however we should have been using `SaveNXSPE` to ensure all the correct metadata gets saved.

It was decided that we would not add an explicit option to save NXSPE files via the interface. Instead, I added a check to see if the algorithm history says the file was loaded in from a .nxspe file or .nxs file. If it finds the file was loaded using the `LoadNXSPE` algorithm, then the corresponding save algorithm would be used to save it. Otherwise, the `SaveNexus` algorithm gets used.

This ensures the saved nxspe workspace can be loaded back into the MSlice interface without any problems.

**To test:**
Load `MARI28540_80meV.nxspe`
Scale the workspace
Save the file as Nexus to a different name such as `MARI28540_80meV_scaled`
There should be an error message displayed at the bottom suggesting that metadata will not be saved.
Save the file as NXSPE to a different name such as `MARI28540_80meV_scaled`
Load the `MARI28540_80meV_scaled` workspace back into MSlice
There should be no error

Fixes #834
Fixes #540

### DIFF
--- a/src/mslice/app/mainwindow.py
+++ b/src/mslice/app/mainwindow.py
@@ -90,6 +90,7 @@ class MainWindow(MainView, QMainWindow):
         self.ws_tab_changed(0)
 
         self.wgtCut.warning_occurred.connect(self.show_warning)
+        self.wgtWorkspacemanager.warning_occurred.connect(self.show_warning)
 
         self.wgtCut.error_occurred.connect(self.show_error)
         self.wgtSlice.error_occurred.connect(self.show_error)
@@ -121,6 +122,7 @@ class MainWindow(MainView, QMainWindow):
     def setup_save(self):
         menu = QMenu()
         menu.addAction("Nexus (*.nxs)", lambda: self.button_save('Nexus'))
+        menu.addAction("NXSPE (*.nxspe)", lambda: self.button_save('NXSPE'))
         menu.addAction("ASCII (*.txt)", lambda: self.button_save('Ascii'))
         menu.addAction("Matlab (*.mat)", lambda: self.button_save('Matlab'))
         self.btnSave.setMenu(menu)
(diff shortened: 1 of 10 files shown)


MialLewis (approved):
Code looks good and works well.

Issue #360, opened 2018-08-01
Incorrect number of bins calculated for some workspaces


MatthewBowles:
Some non-PSD workspaces (.e.g. [this one](https://github.com/mantidproject/mslice/files/2249440/MAR18120_Ei39.53meV.nxspe.zip)) are treated as having the wrong number of bins.

The Energy axis is saved in workspace.limits when the workspace is loaded as `-70, 40, 0.20992` (min, max, step size). Later, these values are used to calculate the number of steps:
```
 (40 - - 70) / 0.20092 = 524.009
```
Which is then rounded up to 525. For most workspaces, this is correct behaviour, but in this case the number of bins is 524. This can cause errors e.g. when changing the intensity, as the number of bins in the workspace is different to what is expected.

srikanthravipati:
@SilkeSchomann I am not sure if this is still an issue. For the attached non-PSD workspace, calculated number of bins is 524 in Workbench!

mducle:
@srikanthravipati and @SilkeSchomann 

Sorry the bug report is very sparse - it made sense at the time but that we a long while ago. I think this bug still exists, but it's not where you think it is.

The issue is in the `ceil` here: https://github.com/mantidproject/mslice/blob/0abeace37a91cdd82621f97bf8f6311f936acbcb/mslice/models/alg_workspace_ops.py#L5

This is really only called by the PSD-algorithms (because we use `BinMD` in thoses cases which explicitly requires the number of steps instead of the step sizes; for non-PSD workspaces we call `SofQWNormalisePolygons` which requires step sizes and not number of steps... ¯\_(ツ)_/¯

Also, it's called by the "intensity types" slice menu options which is where this bug comes in. If you load that data, then plot a slice with the default parameters and then click `Intensity` in the menu bar and `Chi''(Q,E)` and put in a number in the edit box (say `300`), then click ok, you get an error and the traceback:

```
Traceback (most recent call last):
  File "/home/mdl27/src/mslice/mslice/plotting/plot_window/slice_plot.py", line 313, in show_intensity_plot
    if not self._run_temp_dependent(slice_plotter_method, previous):
  File "/home/mdl27/src/mslice/mslice/plotting/plot_window/slice_plot.py", line 363, in _run_temp_dependent
    slice_plotter_method(self.ws_name)
  File "/home/mdl27/src/mslice/mslice/presenters/slice_plotter_presenter.py", line 69, in show_dynamical_susceptibility
[...]

PR #900, opened 2023-04-17
Use closest whole step number if within a small tolerance

### LINKED ISSUE #899: Crash when changing Intensity type
**Describe the bug**
There is a crash on MSlice after plotting a slice and then selecting the intensity type to `Chi''` and selecting the Ei sample log.

**To Reproduce**
1. Open MSlice and load `MARI28540_80meV.nxspe`
2. Plot slice
3. Change Intensity to `Chi''`
4. Select Ei sample log and click ok
5. Crash

**Expected behavior**
There should be no crash, and a warning message if necessary

robertapplin:
**Description of work:**
This PR ensures that the number of steps in an axis gets rounded to the closest whole number if `(axis.end - axis.start)/axis.step` is close to a whole number. We might want to adjust the absolute tolerance in the future if 0.05 is too small.

**To test:**
Follow the instructions in the attached issue

Fixes #899

### DIFF
--- a/src/mslice/models/alg_workspace_ops.py
+++ b/src/mslice/models/alg_workspace_ops.py
@@ -5,14 +5,13 @@ from mslice.models.workspacemanager.workspace_provider import get_workspace_hand
 def get_number_of_steps(axis):
     if axis.step == 0:
         return 1
-    else:
-        step_num = max(1, (axis.end - axis.start) / axis.step)
-        step_num_round = np.around(step_num)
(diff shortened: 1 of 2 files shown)


MialLewis (approved):
Code looks good, and bug is fixed.
To note, this code was originally introduced as a bit of a hacky fix to this bug: https://github.com/mantidproject/mslice/issues/360

We may need to keep an eye on that tolerance value - ideally it should be as low as possible.

Issue #854, opened 2023-01-24
Improve exception handling


SilkeSchomann:
**Describe the bug**
Some exceptions could be handled better, preferably by catching the exception and providing a meaningful error message.

**To Reproduce**
Some scenarios with unhandled exceptions:

- [x] Clear ADS, navigate to MD Histo tab and click on delete button without selecting a workspace.
- [x] Create a cut plot, change intensity to Chi"(Q,E) but click on cancel. https://github.com/mantidproject/mslice/pull/903
- [x] Create a waterfall plot and enter a string in x box, then navigate to y box and hit enter. https://github.com/mantidproject/mslice/pull/904
- [x] https://github.com/mantidproject/mslice/pull/906

PR #903, opened 2023-04-28
Set intensity from action type

robertapplin:
**Description of work:**
This PR fixes a crash caused by caching a string rather than an action object. This causes a crash when calling `setChecked` on the string rather than on the action object.

**To test:**
1. Create a cut plot
2. change intensity to Chi"(Q,E) but click on cancel
3. There should be no crash. The intensity should remain the same

Part of #854

### DIFF
--- a/src/mslice/plotting/plot_window/cut_plot.py
+++ b/src/mslice/plotting/plot_window/cut_plot.py
@@ -640,8 +640,7 @@ class CutPlot(IPlot):
 
     def _set_intensity_to_previous(self, previous_type):
         self._intensity_type = previous_type
-        previous_action = IntensityCache.get_action(previous_type)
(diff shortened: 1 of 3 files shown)


MialLewis (changes requested):

MialLewis:
The fix as is solves the exception in question, but causes an error to be thrown in a few other places.

I've pushed a change which I suggest as an alternative fix.

The code I've changed here is a hangover from when action objects themselves used to be cached that slipped through the net.

robertapplin (commented):
Looks good, your fix is much better. I can't approve this PR unfortunately as I am the original Author

SilkeSchomann (approved):
The fix works well 👍

PR #904, opened 2023-04-28
Use LineEditDoubleValidator to prevent broken e notation

robertapplin:
**Description of work:**
This PR ensures we use a `LineEditDoubleValidator` for the waterfall x and y options so that providing a number in broken e notation will not cause a crash. Instead, the value will get reset to the original value.

**To test:**
Create a waterfall plot
Enter '2e' into the x box, and then click on the y box
The x box should switch back to its original value of 0
Try entering valid values and make sure they still work for x and y

Part of  #854


MialLewis (approved):
Code looks good, fix works.

PR #908, opened 2023-05-04
Fix SaveAscii index out of range error

### LINKED ISSUE #907: Error when saving ascii from mslice command line
**Describe the bug**
When attempting to save a workspace to an ascii file from the mslice terminal, there is an index out of range error.

**To Reproduce**
1. Load `MARI28540_80meV` into Mslice
2. Attempt to save this workspace by typing the following into the terminal:
```
SaveData('MARI28540_80meV', 'test.txt')
```
3. There is an error:
```
    147     y = workspace.raw_ws.extractY()
    148     e = workspace.raw_ws.extractE()
--> 149     ix = [i for i, ax in enumerate(workspace.axes) if 'DeltaE' in ax.units][0]
    150     iy = 0 if ix == 1 else 1
    151 dim_sz = [workspace.raw_ws.getDimension(i).getNBins() for i in range(workspace.raw_ws.getNumDims())]

IndexError: list index out of range
```

**Expected behavior**
There should be no error and the workspace should be saved successfully as an ascii file.

robertapplin:
**Description of work:**
This PR adds a flag `is_slice` to the common workspace properties class. This will provide us with a unified way for determining whether a workspace is a slice or not. Previously, the GUI and CLI used a different method for determining whether a workspace was a slice. I believe the CLI implementation was incorrect and therefore caused the error in the attached issue.

**To test:**
Follow the steps in the attached issue. There should be no error. The file should be saved as expected.

Fixes #907

### DIFF
--- a/src/mslice/cli/__init__.py
+++ b/src/mslice/cli/__init__.py
@@ -3,7 +3,7 @@ from __future__ import (absolute_import, division, print_function)
 from mslice.plotting.pyplot import *  # noqa: F401, F403
 from matplotlib.axes import Axes
 from matplotlib.projections import register_projection
-from mslice.cli.helperfunctions import is_slice, is_cut, is_hs_workspace
+from mslice.cli.helperfunctions import is_cut, is_hs_workspace
 from ._mslice_commands import *  # noqa: F401, F403
 from mslice.app import is_gui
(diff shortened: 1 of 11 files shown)


SilkeSchomann (approved):
It is good to have this fixed and the code cleaned up at the same time!

PR #917, opened 2023-05-26
Fix crash when changing intensity

### LINKED ISSUE #916: Crash when changing slice intensity
**Describe the bug**
Currently mslice crashes when changing the intensity of a slice.

**To Reproduce**
1. Display a slice
2. Change slice intensity to `chi`.

**Expected behavior**
No crash, intensity change.

MialLewis:
**Description of work:**

#908 adds a property `is_slice` to workspaces to differentiate between slices and cuts. When you perform a binary operation on an `mslice` workspace, this property was not passed onto the new resultant workspace and therefore lost.

This ultimately caused a crash when setting the intensity of a slice.

**To test:**
1) Plot a slice.
2) Change intensity to `chi`
3) Observe no crash

Fixes #916


SilkeSchomann (approved):
Confirmed that this fixes the bug while the original problem does not reappear.

PR #920, opened 2023-06-20
Fix incorrect integration of GDOS cuts

### LINKED ISSUE #919: Incorrect intensity for GDOS option in cuts
**Describe the bug**
The computed GDOS cuts in MSlice should agree with the output of the `ComputeIncoherentDOS` algorithm in Mantid but currently does not.

**To Reproduce**
1. Load a data file (e.g. `MAR21335_Ei60.00meV.nxs` in the test dataset)
2. Make a cut, then click on the "Intensity" menu option and select "GDOS". Type in a value for the temperature (e.g. 5).
3. Make a slice using the default parameters and set the settings to be able to see invisible workspaces; find out which of these workspaces are of `RebinnedOutput` type. Run the `ComputeIncoherentDOS` algorithm on this workspace with the same limits as for the cuts, the same temperature and with `MeanSquareDisplacement` set to zero.
4. Check that the two cuts (with MSlice and with `ComputeIncoherentDOS`) are the same - on nightly `6.6.2023.0602.1033` they are not (see figure below).

This script should do the same action as above but currently errors in the `Cut(..., IntensityCorrection='gdos', ...)` part, with a `TypeError: cut_compute_gdos() missing 1 required positional argument: 'is_icut'`.

```python
import mslice.cli as mc
import matplotlib.pyplot as plt
from mantid.simpleapi import ComputeIncoherentDOS

ws_MAR21335_Ei60_00meV = mc.Load(Filename='MAR21335_Ei60.00meV.nxs', OutputWorkspace='MAR21335_Ei60.00meV')
slice_ws = mc.Slice(ws_MAR21335_Ei60_00meV, Axis1="|Q|,0.27356,11.0351,0.04671", 
                    Axis2="DeltaE,-30.0,58.2,0.3", NormToOne=False)
ws_dos = ComputeIncoherentDOS(mc.Slice(ws_MAR21335_Ei60_00meV).raw_ws, Temperature=5, MeanSquareDisplacement=0,
                              QSumRange='6,11', EnergyBinning='5,0.5,55', StoreInADS=False)

cut_ws_0 = mc.Cut(ws_MAR21335_Ei60_00meV, CutAxis="DeltaE,5.0,55.0,0.5", IntegrationAxis="|Q|,6.0,11.0,0.0", 
                  NormToOne=False, IntensityCorrection='gdos', SampleTemperature=5.0)
                  
ax = plt.subplot(111, projection='mantid')
ax.errorbar(ws_dos, 'or', label='Mantid')
ax.errorbar(cut_ws_0.raw_ws, '.k', label='MSlice')
ax.legend()
```

With the manual cuts, I found that the MSlice cuts was about 30x the intensity of the `ComputeIncoherentDOS` and also had much large errorbars and a different shape:

image

i.e. the two peaks (at 20 and 35meV) should have about equal heights, but in the MSlice cut, the low energy (20meV) peak is clearly bigger. Check with the literature: https://authors.library.caltech.edu/9459/1/KREprb08.pdf at low energies the two peaks should be about equal (although the lower energy peak is slightly higher).

**Expected behavior**
Cuts using MSlice with the "GDOS" intensity correction option should be the same as with the `ComputeIncoherentDOS` in Mantid.

**Screenshots**
See above

 - MSlice from Mantid nightly `6.6.2023.0602.1033` on DAaaS (Scientific Linux 7).

mducle:
**Description of work:**

As noted in the issue #919, the GDOS cuts computed by MSlice does not agree with that computed by the `ComputeIncoherentDOS` algorithm although they use in principle the same equations method. In particular there are two steps to the calculation:

1. Compute a 2D slice `GDOS(q,w) = S(q,w) * (E/Q^2) * bose` as noted in the [docs](https://mantidproject.github.io/mslice/slicing.html#converting-intensity)
2. Integrate this 2D slice in either Q (only mode supported by `ComputeIncoherentDOS`) or in E.

The issue is with this second step. This is because the first step produces a `RebinnedOutput` type workspace which has a (hidden-from-Python) extra weight for each bin. To correctly compute the integration/rebin this weight must be taken into account, but the previous MSlice code simply sums the counts in each bin. Instead it should use the Mantid `Rebin2D` or `Integration` algorithms which take into account the fractional weight rather than just a simplistic sum.

**To test:**

Run the script in the issue and checks that the MSlice and `ComputeIncoherentDOS` outputs agree.
Plot the slice and change intensity to `GDOS` and check that interactive plotting works as expected.

Fixes #919.


SilkeSchomann (approved):
The GDOS cuts now match the output of the ComputeIncoherentDOS in Mantid.

src/mslice/models/cut/cut_algorithm.py
```diff
@@ -126,27 +126,30 @@ def _compute_cut_nonPSD(selected_workspace, cut_axis, integration_axis, emode, a
...
             ws_out = Transpose(InputWorkspace=ws_out, EnableLogging=False)
             ws_out = Rebin(InputWorkspace=ws_out, Params=ax2_binning, EnableLogging=False)
         else:
-            ws_out = Rebin(InputWorkspace=converted_nonpsd, Params=ax1_binning, EnableLogging=False)
+            ws_out = Rebin(InputWorkspace=selected_workspace, Params=ax1_binning, EnableLogging=False)
             ws_out = Transpose(InputWorkspace=ws_out, EnableLogging=False)
             ws_out = Integration(InputWorkspace=ws_out, RangeLower=ax2[0], RangeUpper=ax2[2], EnableLogging=False)
     else:
-        ws_out = Rebin2D(InputWorkspace=converted_nonpsd, Axis1Binning=ax1_binning, Axis2Binning=ax2_binning,
+        ws_out = Rebin2D(InputWorkspace=selected_workspace, Axis1Binning=ax1_binning, Axis2Binning=ax2_binning,
```
MialLewis, line 142: When testing on a slice taken from a PSD workspace, this alg threw an error.

There appears to be two issues:
1) We cannot use the algorithm whilst the workspace does not have a name. The name is stripped by `CloneWorkspace` whilst performing a binary op during GDOS correction as the workspace is not stored in the ADS. We can fix this temporarily by cloning the workspace here, storing it in the ADS and giving it a name.
2) Following the above fix a histogram workspace is output rather than a rebinned workspace, which can not be used with `Rebin2D`.

I now remember that this issue was the reason I was summing the slice bins in the first place, rather than using `Rebin` or `cut`.

Do you have any idea on a solution for this? If there is nothing simple, we may have to disable `GDOS` correction on slices taken from `PSD` workspaces for this release.

mducle: Right, sorry I didn't check this but the original routine with `np.nansum` should work for these workspaces (because they don't have the extra hidden-from-python weights), so I'll restore that bit of the code for PSD workspaces.

src/mslice/models/intensity_correction_algs.py
```diff
@@ -220,6 +219,31 @@ def _reduce_bins_along_int_axis(slice_gdos, algorithm, cut_axis, int_axis, cut_a
...
+        ws_out = _cut_nonPSD_general(ax2, ax1, slice_gdos.raw_ws, algorithm)
+    signal = ws_out.extractY()
+    error = ws_out.extractE()
+    return signal, error
+
+
+def _reduce_bins_and_return_signal_error_PSD(slice_gdos, algorithm, int_axis, cut_axis_id, cut_slice_alignment):
+    # PSD type workspace, just sum all the bins
+    signal = np.nansum(slice_gdos.get_signal(), axis=cut_axis_id, keepdims=True)
+    error = np.nansum(np.sqrt(slice_gdos.get_variance()), cut_axis_id, keepdims=True)
```
MialLewis, line 237: Note that I changed this from `error = np.sqrt(np.nansum(slice_gdos.get_variance()), cut_axis_id, keepdims=True))`

mducle: @MialLewis sorry I didn't get a chance to check this as I was still at the workshop yesterday but I think this is incorrect. `np.sqrt(slice_gdos.get_variance())` is the same as `slice_gdos.get_error()` but what we need to calculate the error in the sum is to use the quadrature rule:

```
sqrt(e1^2 + e2^2 + e3^3 + ...)
```

(see [wiki](https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae)). e1^2 etc are the variances

So it really has to be:

```
error = np.sqrt(np.nansum(slice_gdos.get_variance()), cut_axis_id, keepdims=True))
```

This is the reason the errorbars calculated in the code appears to be very large.

mducle: No probs... I think the issue was that the code in the test was not changed and was the old code so you reverted the code to make the test pass... Anyway, I think it's ok that this fix goes into nightly.

Thanks for looking into this!

PR #921, opened 2023-06-21
Fix crash when correcting GDOS intensity of pixel ws

MialLewis:
**Description of work:**

As per a similar bug fix, this PR ensures that `is_slice` propagates on binary operations involving histogram workspaces.

This prevents a crash when correcting the intensity of a slice taken from a pixel workspace to `GDOS`.

This bug was introduced through this pr: https://github.com/mantidproject/mslice/pull/908

**To test:**
1) Load `HYS_63373_4pixel.nxspe`
2) Project, and plot as a slice
3) Take an `icut`.
4) Using the slice window, correct the intensity to `GDOS`.
5) Observe no crash.


SilkeSchomann (approved):
Confirmed that changing the intensity to GDOS does not cause a crash anymore.

PR #922, opened 2023-06-23
Correct error propagation in GDOS intensity correction

MialLewis:
**Description of work:**

This corrects a bug where errors were simply being summed as if they were absolute. This is not the case, and they need to be added in quadrature.

https://docs.mantidproject.org/v3.7.2/concepts/ErrorPropagation.html

**To test:**

1). Plot a slice, take a cut where appropriate
2) Correct to GDOS
3) See smaller error bars.


SilkeSchomann (approved):
Error bars are not elongated anymore.

PR #926, opened 2023-07-19
Updated Python from 3.8 to 3.10

SilkeSchomann:
**Description of work:**

**To test:**

Fixes #xxxx.


robertapplin:
I can confirm the QMenu's appear to be formatted as expected with no weird behaviour

.github/workflows/unit_tests_nightly.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           miniforge-variant: Mambaforge
           activate-environment: mslice-env
           environment-file: environment.yml
-          python-version: 3.8
+          python-version: 3.10
```
robertapplin, line 25: Do we no longer need this as `python = 3.10` is in the environment.yml file?

robertapplin: Same for the `deploy_*` workflows

conda/conda_build_config.yaml
```diff
@@ -1,5 +1,5 @@
 python:
-  - 3.8
+  - 3.10
 
 setuptools:
-  - 62.0
+  - 65.0
```
robertapplin, line 5: Out of interest, do you know what problem is tying us to setuptools v65 ?

robertapplin (changes requested):
Looks good, I have a few questions

.github/workflows/unit_tests_nightly.yml
SilkeSchomann: Good spot, it should be redundant by now.

conda/conda_build_config.yaml
SilkeSchomann: There used to be a problem with newer setuptools versions but it can be unpinned now.

SilkeSchomann:
I've addressed requested changes and also removed code that was only necessary for testing.

robertapplin (approved):
Looks good! I did a grep for mentions of python 3.8 and didn't find anything. Also did a very quick test of opening mslice, loading data, and doing a cut/slice. Seems to be working same as before :shipit:

PR #935, opened 2023-09-14
Propagate recoil and bragg line colors to script

### LINKED ISSUE #931: Colour change for recoil line/Bragg peak on plot is not propagated to generated script
**Describe the bug**
When changing the colour of a recoil line or Bragg peak on a plot and then generating a script, the script does not use the new colour.

We used to have a similar problem with the font size: https://github.com/mantidproject/mslice/issues/795

robertapplin:
**Description of work:**
This PR ensures the line colors of the bragg and recoil lines on a Cut and Slice plot get propagated to a generated plotting script.

**To test:**
Load data into mslice
Do a slice
Add a bragg peak and recoil line, and change the color
Generate a script to clipboard
Run the script in mantid
The colors should be propagated

Fixes #931

### DIFF
--- a/src/mslice/cli/__init__.py
+++ b/src/mslice/cli/__init__.py
@@ -53,7 +53,7 @@ class MSliceAxes(Axes):
         _update_overplot_checklist(key)
         _update_legend()
 
-    def bragg(self, workspace, element=None, cif=None):
+    def bragg(self, workspace, element=None, cif=None, **kwargs):
         from mslice.app.presenters import get_cut_plotter_presenter, get_slice_plotter_presenter
         _check_workspace_name(workspace)
         workspace = get_workspace_handle(workspace)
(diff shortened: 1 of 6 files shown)


src/mslice/cli/__init__.py
```diff
@@ -36,7 +36,7 @@ def pcolormesh(self, *args, **kwargs):
         else:
             return Axes.pcolormesh(self, *args, **kwargs)
 
-    def recoil(self, workspace, element=None, rmm=None):
+    def recoil(self, workspace, element=None, rmm=None, **kwargs):
```
MialLewis, line 39: Can I ask why you decided to use `kwargs` here, and not just add an argument for color?

robertapplin: Elsewhere in Mantid kwargs are used quite often to pass a large number of plotting options through the mantid interface and onto matplotlib. I decided that using kwargs here would keep consistency, and allows us to easily propagate more plotting arguments in the future (e.g. linewidth)

MialLewis (approved):

PR #937, opened 2023-09-14
Escape special characters in workspace name when renaming

### LINKED ISSUE #934: Workspace name containing a ')' causes problems
**Describe the bug**
Renaming a workspace to a name containing a closing bracket (')') causes this workspace to function. Renaming this workspace again does not solve the problem.

This could be a Mantid-wide issue.

robertapplin:
**Description of work:**
This PR ensures we escape special characters when renaming workspaces. Previously, renaming a workspace containing a special character would cause an exception.

**To test:**
Load data
Rename the workspace to contain a `)` character
Attempt to rename the workspace a second time
There should be no exception and renaming the workspace should work.

Fixes #934


MialLewis (approved):

robertapplin:
I think we merged this to main by accident

SilkeSchomann:
Yes, the code is now also present in release-next.

PR #939, opened 2023-09-15
Fix save plot saving underlying ws rather than derived ws

### LINKED ISSUE #927: Not saving slice as text file
**Describe the bug**
If I plot a slice and the click the "floppy disk" icon to save the slice, and select the output type as "ASCII", MSlice now saves the underlying (loaded) workspace and not the slice workspace.

**To Reproduce**
1. Load data and plot a slice.
2. Click on the floppy disk icon and select "Ascii" option.
3. In Mantid workbench load the data again and check it is the loaded workspace and not the slice.

**Expected behavior**
The callback for the save icon in the slice window should call `mc.save_ascii(slice_ws)`.

**Screenshots**
N/A

 - MSlice Version from Mantid 6.7 on DAaaS.

**Additional context**
Reported by users.

MialLewis:
**Description of work:**

This rolls back some changes in #908.

The change of the if conditional to `is_slice` misses the fact that the workspace provided is actually the underlying workspace, which then needs to be converted to a derived workspace before saving.

I've removed the `slice_nonpsd` argument from `save_workspaces` and  `_save_single_ws` as in the only use in the current code it is fed a hardcoded argument as `True` (even though the default was `False`). It was also confusing for two reasons:
-  The workspace being saved could also be a `cut`
- In the `_save_single_ws` function there is the statement `non_psd_slice = slice_nonpsd and isinstance(workspace, Workspace) and not workspace.is_PSD`, which seems to imply inequivalence between `non_psd_slice` and `slice_nonpsd`.

I don't understand what this if statement is trying to check. So I've just rolled it back to what it was previously minus the redundant `slice_nonpsd` variable. Prehaps @mducle can advise so we can change it to something clearer: https://github.com/mantidproject/mslice/commit/b87a65235d98d1f765ca4c2ca78f2c05c95154e9

**Additionally**: I've snuck in a fix for a unit test that was failing on windows, but not our linux github runners.

**To test:**

- Load an initial dataset
- Plot any slice
- Using the save button on the slice plot, save the plot as ACII.
- Using `workbench`, load the initial dataset again, save this dataset as ASCII using the `SaveASCII` alg.
- Compare the two files to see that they are not the same anymore.

Fixes #927

### DIFF
--- a/src/mslice/models/workspacemanager/workspace_algorithms.py
+++ b/src/mslice/models/workspacemanager/workspace_algorithms.py
@@ -236,13 +236,12 @@ def scale_workspaces(workspaces, scale_factor=None, from_temp=None, to_temp=None
             propagate_properties(ws, result)
 
 
-def save_workspaces(workspaces, path, save_name, extension, slice_nonpsd=False):
+def save_workspaces(workspaces, path, save_name, extension):
     """
     :param workspaces: list of workspaces to save
     :param path: directory to save to
     :param save_name: name to save the file as (plus file extension). Pass none to use workspace name
     :param extension: file extension (such as .txt)
-    :param slice_nonpsd: whether the selection is in non_psd mode
     """
     if extension == '.nxs':
         save_method = save_nexus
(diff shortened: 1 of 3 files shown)


SilkeSchomann (approved):
The slice plot is now saved as an asci txt file and not the underlying workspace.

It is also good to have the file IO test corrected!

Issue #913, opened 2023-05-17
Error when using a script generated by mslice


MialLewis:
**Describe the bug**
When using scripts generated by mslice they often error (I've seen a few different messages).

**To Reproduce**
Haven't had chance to replicate yet, but this was one script generated:

```
# Python Script Generated by Mslice on 2023-05-17 11:07:51
import mslice.cli as mc
import mslice.plotting.pyplot as plt
ws_MAR21335_Ei60meV = mc.Load(Filename='C:/Users/joy22959/Development/mantid/build/ExternalData/Testing/Data/UnitTest\MAR21335_Ei60meV.nxs', OutputWorkspace='MAR21335_Ei60meV')
fig = plt.gcf()
fig.clf()
ax = fig.add_subplot(111, projection="mslice")
cut_ws_0 = mc.Cut(ws_MAR21335_Ei60meV, CutAxis="|Q|,0.0,10.0,0.04671", IntegrationAxis="DeltaE,-5.0,5.0,0.0", NormToOne=False, IntensityCorrection=False, SampleTemperature=None)
ax.errorbar(cut_ws_0, label="MAR21335_Ei60meV -5.00<E<5.00", color="#1f77b4", marker="o", ls="-", lw=1.5)
from mslice.cli.helperfunctions import hide_a_line_and_errorbars, append_visible_handle_and_label
from mslice.util.compat import legend_set_draggable
# hide lines, errorbars, and legends
handles, labels = ax.get_legend_handles_labels()
visible_handles = []
visible_labels = []
append_visible_handle_and_label(visible_handles, handles, visible_labels, labels, 0)
legend_set_draggable(ax.legend(visible_handles, visible_labels, fontsize='medium'), True)
ax.set_title('', fontsize=12.0)
ax.set_ylabel(r'Signal/#Events', fontsize=10.0)
ax.set_ylim(bottom=-17.772422415630686, top=127.87112555418936)
ax.set_xlim(left=-0.17990654205607487, right=10.460280373831777)
ax.bragg(workspace='MAR21335_Ei60meV_cut(32.455,50.345)', element='Aluminium')
mc.Show()
```

This line clearly has wrong cut limits `ax.bragg(workspace='MAR21335_Ei60meV_cut(32.455,50.345)', element='Aluminium')`

mducle:
I can reproduce this error on Linux (CentOS 7) nightly `6.6.20230413.1231` (sorry a bit old, but the latest on IDAaaS):

[...]

PR #940, opened 2023-09-19
Fix mslice scripting issue when overplotting bragg peaks

MialLewis:
**Description of work:**

This PR fixes various scripting issues:
- Generated scripts default to `plot_over=True`. In this case, if there is a current active figure with Bragg peaks, it will try to reset them though they will no longer exist as the figure will have been wiped previously in the script. This PR sets `plot_over=False` on the first line plotted, and `plot_over=True` on subsequent lines.
- When plotting a cut with `plot_over=False`, the workspace name cached on the `CutPlot` was not updated. This filtered through to any script generated from the plot. This has been fixed, though the script issue has also been resolved by...
- When adding overplot features to a cut plot, generated scripts now use the workspace variable rather than the workspace name. Due to the appending of `_(1)` etc, on duplicate cuts, the name of a workspace cannot be assumed based upon the workspace it is taken from.

Test coverage has been added for any new functionality.

I have updated `.format` to f strings in the relevant files.

**To test:**

Follow instructions left by @mducle in the linked issue, see all resolved.

Fixes issues seen in #913.


src/mslice/scripting/helperfunctions.py
```diff
@@ -107,23 +107,23 @@ def add_overplot_statements(script_lines, plot_handler):
         recoil = True if rmm is not None or key in [1, 2, 4] else False
         cif = None  # Does not yet account for CIF files
 
+        ws_var = ws_vars.pop(0) if ws_vars else f"'{plot_handler.ws_name}'"
```
robertapplin, line 110: Does the pop bit need single quotes around it?

```suggestion
        ws_var = f"'{ws_vars.pop(0)}'" if ws_vars else f"'{plot_handler.ws_name}'"
```

robertapplin (changes requested):

src/mslice/scripting/helperfunctions.py
MialLewis: I don't _think_ so, as ws_vars.pop(0) returns a string as it is.

robertapplin: Yes I think it does return a string, however when it is input into the python code in the `script_lines` variable, it will not have quotes around the variable name and so it doesn't think its a string inside the python script we create. e.g. the color string variable needs quotes
```
ax.recoil(workspace={ws_var}, rmm={rmm}, color='{color}')\n
```

MialLewis: Ah I see what you mean now, this is intentional in this case, the `ws_var` refers to a python variable that we assign to earlier in the script, not the name of a workspace in the form of a string.

robertapplin (approved):
Code looks good and is working as expected

PR #946, opened 2023-09-27
Fix legend remaining turned off when Waterfall is clicked

### LINKED ISSUE #863: Legend is displayed when clicking on Waterfall despite being toggled off
**Describe the bug**
When on a plot with several lines legends are toggled off and Waterfall is clicked, legends appear again but the Legend button seems to be off.

**To Reproduce**

1. Load MAR21335_Ei60meV
2. On Cut tab, plot a cut along DeltaE from 0 to 5, from 0 to 5, width 1
3. Click on Plot
4. Toggle legends off
5. Click on Waterfall

**Expected behavior**
Legends should remain toggled off.

GuiMacielPereira:
### Summary of changes
- Introduced an if statement that checks a boolean flag on whether the legend is supposed to be shown or not.
- Edited the unit test of the changed function to include testing of both options, i.e. when legend is not shown and when legend is shown

#### Finding the problem
- When the legends box is clicked, a `_toggle_legend` function is activated which sets the boolean attribute `_legends_shown` to its opposite
- The attribute `_legends_shown` of `CutPlot` is thus supposed to control whether or not the legend is visible in the plot.
- The `update_legend` function does not make use of the attribute `_legends_shown` in its scope, which leads to legends being shown even when `_legends_shown` is set to False.
- This behaviour is present in 3 scenarios on cut viewer, as all these instances call the function `update_legend`:
  - When Waterfall is clicked
  - When the Intensity is changed
  - When any line in the plot is clicked

#### Fixing the problem
- Introduced an `if` statement inside `update_legend`
- Same `if` statement is used in the analogous slice viewer

#### Edited unit test
- Introduced the new scenario of making sure `axes.legend()` is not called when `_legends_shown` is set to `False`.

## To test
- Open cut viewer in mslice and run over the 3 scenarios that were previously making the legend visible:
   - Click **Legend** so legends become invisible
   - Click **Waterfall** and check legends remain invisible
   - Click any plot line > cancel, and check legends remain invisible
   - Click **Intensity** > any intensity that works, and check legends remain invisible
 
- Comment out `if`` statement and run test `cut_plot_test.py`. It should fail the test `test_update_legend`. Restore `if` statement and check the test passes.

Fixes #863 .

### DIFF
--- a/src/mslice/plotting/plot_window/cut_plot.py
+++ b/src/mslice/plotting/plot_window/cut_plot.py
@@ -191,8 +191,10 @@ class CutPlot(IPlot):
                     handles_to_show.append(handles[i])
                     labels_to_show.append(line_data[i]['label'])
                 self._legends_visible[i] = line_data[i]['legend']
-        legend = axes.legend(handles_to_show, labels_to_show, fontsize='medium')  # add new legends
-        legend_set_draggable(legend, True)
+
+        if self._legends_shown:
+            legend = axes.legend(handles_to_show, labels_to_show, fontsize='medium')  # add new legends
+            legend_set_draggable(legend, True)
 
     def change_axis_scale(self, xy_config):
         current_axis = self._canvas.figure.gca()
--- a/src/mslice/plotting/plot_window/cut_plot.py
+++ b/src/mslice/plotting/plot_window/cut_plot.py
@@ -178,19 +178,16 @@ class CutPlot(IPlot):
         handles_to_show = []
         handles, labels = axes.get_legend_handles_labels()
         if line_data is None:
-            i = 0
-            for handle, label in zip(handles, labels):
+            for i, (handle, label) in enumerate(zip(handles, labels)):
                 if self.legend_visible(i):
                     labels_to_show.append(label)
                     handles_to_show.append(handle)
-                i += 1
         else:
-            containers = axes.containers
-            for i in range(len(containers)):
-                if line_data[i]['legend']:
-                    handles_to_show.append(handles[i])
-                    labels_to_show.append(line_data[i]['label'])
-                self._legends_visible[i] = line_data[i]['legend']
+            for line, handle in zip(line_data, handles):
+                if line['legend']:
+                    handles_to_show.append(handle)
+                    labels_to_show.append(line['label'])
+            self._legends_visible = [line['legend'] for line in line_data]
 
         if self._legends_shown:
             legend = axes.legend(handles_to_show, labels_to_show, fontsize='medium')  # add new legends
--- a/tests/cut_plot_test.py
+++ b/tests/cut_plot_test.py
@@ -85,7 +85,7 @@ class CutPlotTest(unittest.TestCase):
         self.cut_plot.update_legend.assert_called_once()
         self.cut_plot._canvas.draw.assert_called_once()
 
-    def test_update_legend(self):
+    def test_update_legend_legends_not_shown(self):
         line = Line2D([], [])
         self.axes.get_legend_handles_labels = MagicMock(return_value=([line], ['some_label']))
         self.cut_plot._legends_shown = False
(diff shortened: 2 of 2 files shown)


tests/cut_plot_test.py
```diff
@@ -88,6 +88,12 @@ def test_object_clicked(self, quick_options_mock):
     def test_update_legend(self):
         line = Line2D([], [])
         self.axes.get_legend_handles_labels = MagicMock(return_value=([line], ['some_label']))
+        self.cut_plot._legends_shown = False
```
MialLewis, line 91: Good unit test, but typically we want to separate out each test case. For example `def test_update_legend_legends_shown` and `def test_update_legend_legends_not_shown`.

The title of the failure should point you directly to the failed case. This is not always the case within mantid and mslice, but for new tests we try to follow this.

MialLewis (changes requested):

tests/cut_plot_test.py
GuiMacielPereira: Ah I see! Will do that then.

GuiMacielPereira: How does it look now?

MialLewis: Thanks for separating them out.

Do you know what we are checking for with  `self.assertTrue(self.cut_plot._legends_visible[0])`? It is asserting `True` for both, intuitively it feels like it should be asserting `False` if the legends are not shown.

GuiMacielPereira: Actually, I am not sure. There are 3 atributes of CutPlot that control the legend: `._lines_visible = {}`, .`_legends_visible = [] `and `._legends_shown = True`. My best guess thus far is that `_legends_visible` is a list of booleans that controls which lines are displayed in the legend box itself, rather than the whether the legend box is visible or not. 

Should I look into it further to make sure?

MialLewis: That makes sense, it's probably work a check to confirm that, and maybe updating the variable name to something that is clearer!

src/mslice/plotting/plot_window/cut_plot.py
```diff
@@ -178,19 +178,16 @@ def update_legend(self, line_data=None):
         handles_to_show = []
         handles, labels = axes.get_legend_handles_labels()
         if line_data is None:
-            i = 0
-            for handle, label in zip(handles, labels):
+            for i, (handle, label) in enumerate(zip(handles, labels)):
                 if self.legend_visible(i):
                     labels_to_show.append(label)
                     handles_to_show.append(handle)
-                i += 1
```
GuiMacielPereira, line 186: Cleaned this loop with enumerate

src/mslice/plotting/plot_window/cut_plot.py
```diff
@@ -178,19 +178,16 @@ def update_legend(self, line_data=None):
...
-            for i in range(len(containers)):
-                if line_data[i]['legend']:
-                    handles_to_show.append(handles[i])
-                    labels_to_show.append(line_data[i]['label'])
-                self._legends_visible[i] = line_data[i]['legend']
+            for line, handle in zip(line_data, handles):
+                if line['legend']:
+                    handles_to_show.append(handle)
+                    labels_to_show.append(line['label'])
+            self._legends_visible = [line['legend'] for line in line_data]
```
GuiMacielPereira, line 191: The `axes.containers `variable was just being used to loop through `handles` and` line_data`, so I removed it and used `zip` instead. Actually, inside this `else` statement, `handles` **will be the same as** `axes.containers` (I checked this in the source code of matplotlib), so it is best removed anyway.

tests/cut_plot_test.py
```diff
@@ -101,6 +101,19 @@ def test_update_legend_legends_shown(self):
...
+            {'shown': True, 'color': 'cyan', 'style': ':', 'width': 1.5, 'marker': 'o', 'legend': 0, 'label': 'non_visible_line_data_label', 'error_bar': True}
+        ]
+        mock_line = Line2D([], [])
+        another_mock_line = Line2D([], [])
+        self.axes.get_legend_handles_labels = MagicMock(return_value=([mock_line, another_mock_line], ['mock_label', 'another_mock_label']))
+
+        self.cut_plot.update_legend(line_data)
+        self.assertEqual(self.cut_plot._legends_visible, [2, 0])
+        self.axes.legend.assert_called_with([mock_line], ['visible_line_data_label'], fontsize=ANY)
+
```
GuiMacielPereira, line 118: Added this unit test function to test labels that should be visible within the legends box

PR #949, opened 2023-10-09
942 increase all fonts in plot

### LINKED ISSUE #942: Increase font size simultaneously
Currently when setting font size, you have to do it individually for each element. This is laborious.

Preferably, there should be an option to scale all the fonts on a plot at once, or specify a fixed size.

Requested by Adroja.

GuiMacielPereira:
**Summary of Changes**
- Added widget to set all fonts in plot to a given value, present when Plot Options is opened.
- Added another widget to control scaling of all fonts

**Description of work:**
- Three main components were altered: the GUI window (also called view) class, the cut and slice plots classes, and the PlotOptionsPresenter class (which functions and the link between the GUI and the internal class for controlling plot options)
- The GUI window: Added four widgets in total: two checkboxes and two line edits:
image
- Added functions in Plot Options class that controls how the checkboxes interact with the line edits and between themselves (can't check the two checkboxes simultaneously, and can only edit line edits if checkbox is checked)
- Added signals in Plot Options class that are caught by PlotOptionsPresenter in `plot_options_presenter.py` to signal whether the two new fields (for setting all font sizes and scaling all font sizes) are part of the modified values. Added a new function also in this class that removes elements from the dictionary of modified values. This function is triggered when user unticks either of the checkboxes, so that any changes previously made are internally discarded from the dict of modified values. Similarly, when a checkbox is ticked, the default values are loaded onto`modified values` and any sunbsequent edit by the user in the Gui apdatest these values.
- In `cut_plot.py`, the implementation of setting and scaling all fonts is contained whithin the funcitons `CutPlot.all_fonts_size` and `CutPlot.scale_all_fonts` respectively. These two functions are implemented by getter and setter decorators.
- Added tests for cut and slice plots, on the implementation of setting and scaling all fonts. Added another test in `cut_options_presenter_test.py` to confirm that data transfer between GUI view and internal class is happening successfully.
- Manually confirmed that all functunality is preserved whether the plot options window is kept open or not.

**To test:**

- Open plot options window and play around with the new widgets. 
    - Change fonts of plot one by one to different sizes and check that setting all fonts to the same size works.
    - Change fonts to different sizes again and check that scaling all fonts preserves proportions.
    - Enable and disable checkboxes and confirm that plot fonts remain unaltered.
    - Check that it is not possible to simultaneously setect both checkboxes.
    - Enable one checkbox but don't edit the corresponding line, changes should still be applied to plot
 - Change one of the values in the tests for slice and cut plots and check that test fails.
 - In `cut_options_presenter_test.py`, comment out any line that calls `self.presenter._value_modified` and check that test fails (this is the equivalent of no checkboxes being ticked)

Fixes #942 .

### DIFF
--- a/src/mslice/plotting/plot_window/slice_plot.py
+++ b/src/mslice/plotting/plot_window/slice_plot.py
@@ -469,6 +469,10 @@ class SlicePlot(IPlot):
     def colorbar_label_size(self):
         return self._canvas.figure.get_axes()[1].yaxis.label.get_size()
 
+    @colorbar_label_size.setter
+    def colorbar_label_size(self, value):
+        self._canvas.figure.get_axes()[1].yaxis.label.set_size(value)
+
     @property
     def colorbar_range(self):
         return self._canvas.figure.gca().collections[0].get_clim()
@@ -508,6 +512,10 @@ class SlicePlot(IPlot):
     def title_size(self):
         return self.manager.title_size
 
+    @title_size.setter
+    def title_size(self, value):
+        self.manager.title_size = value
+
     @property
     def x_label(self):
         return self.manager.x_label
@@ -523,6 +531,10 @@ class SlicePlot(IPlot):
     def x_label_size(self):
         return self.manager.x_label_size
 
+    @x_label_size.setter
+    def x_label_size(self, value):
+        self.manager.x_label_size = value
+
     @property
     def y_label(self):
         return self.manager.y_label
@@ -537,6 +549,10 @@ class SlicePlot(IPlot):
     @property
     def y_label_size(self):
         return self.manager.y_label_size
+
+    @y_label_size.setter
+    def y_label_size(self, value):
+        self.manager.y_label_size = value
 
     @property
     def x_range(self):
@@ -609,3 +625,17 @@ class SlicePlot(IPlot):
 
     def set_cross_cursor(self):
         self._canvas.setCursor(Qt.CrossCursor)
+
+    @property
+    def all_fonts_size(self):
+        return DEFAULT_LABEL_SIZE
+
+    @all_fonts_size.setter
+    def all_fonts_size(self, value):
+        self.title_size = value
+        self.x_range_font_size = value
+        self.y_range_font_size = value
+        self.x_label_size = value
+        self.y_label_size = value
+        self.colorbar_label_size = value
+        self.colorbar_range_font_size = value
--- a/src/mslice/plotting/plot_window/slice_plot.py
+++ b/src/mslice/plotting/plot_window/slice_plot.py
@@ -58,10 +58,10 @@ class SlicePlot(IPlot):
         self.plot_fonts_properties = [
                 'title_size',
                 'x_range_font_size',
-                'y_range_font_size', 
+                'y_range_font_size',
                 'x_label_size',
                 'y_label_size',
-                'colorbar_label_size', 
+                'colorbar_label_size',
                 'colorbar_range_font_size'
                 ]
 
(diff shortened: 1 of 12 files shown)


robertapplin (changes requested):
Just had a quick first look, the code looks great. Very neat and tests have been added

I have a few questions including
- I was surprised how the "scale font" option worked, as I was expecting it to work more like a scale factor where if it was 1 then the font would stay the same size, if it was 2 it would double etc. . Do you know which interpretation the original issue was aiming for?

- When I tick the "Set all fonts size" option and set it to `20`. Then if I change it again and set it to an invalid value such as `bad`, the font sizes will change back to `10`. I think it shouldn't change the font size at all when an invalid size is provided?
- The last thing I would suggest is to rename the `Scale all fonts` option to match the other option so it reads `Scale all font sizes`

Overall, looks really great and works very well!

GuiMacielPereira:
- Changed 'scale font' in variable names to 'increment font', as it describes more accurately what it does.
- Added a double validator, so string input like 'bad' is not allowed.
- Handled very big or very small font sizes by introducing `if` statements that set font size to reasonable limits: 1 or 1000.
     - I tried imposing a range of (1, 1000) directly on the validator but it caused a worse user experience (for example, no longer able to type numbers with decimal places, and text field would get automatically set to `0.0` after an invalid input - which, in itself, is also an invalid input)
- Changed GUI description to `Increment all font sizes`

robertapplin:
The tooltip change looks good 
image

I think the next step is:
- Make the font box empty if there is more than one font size. In this is the case, if `Apply` is clicked, then the font sizes to use should be the ones displayed in the tooltip
- Populate the font box if there is only one font size. Use this font size for everything when `Apply` is clicked

GuiMacielPereira:
Overall Changes:
- UI is much simpler and more intuitive:
- 
image
- Real-time updating of fonts, not dependent of clicking 'OK' button

src/mslice/plotting/plot_window/slice_plot.py
```diff
@@ -609,3 +634,24 @@ def _get_overplot_datum():  # needed for interface consistency with cut plot
...
+        return font_sizes_config
+
+    @all_fonts_size.setter
+    def all_fonts_size(self, values: dict):
+        for key in values:
+            setattr(self, key, values[key])
+
+    def increase_all_fonts(self):
+        for p in self.plot_fonts_properties:
+            setattr(self, p, getattr(self, p) + 0.5)
```
robertapplin, line 653: Incrementing and decrementing font sizes works really well, so good job with this. The only thing I would change is I think 0.5 is quite small, perhaps we want it to be 1 to keep the font sizes at whole numbers?

It would also be good to have this number as a variable at the top of the file in capitals to signify it is a constant

src/mslice/presenters/plot_options_presenter.py
```diff
@@ -27,6 +34,31 @@ def _xy_config_modified(self, key):
...
+        if fonts_size is not None:
+            new_config_dict = {key: fonts_size for key in new_config_dict}
+
+        self._model.all_fonts_size = new_config_dict
+
+    def _set_font_sizes_tooltip(self):
+        tip = self._convert_font_config_to_tooltip()
+        self._view.allFntSz.setToolTip(tip)
+        self._view.sclUpFntSz.setToolTip(tip)
+        self._view.sclDownFntSz.setToolTip(tip)
```
robertapplin, line 53: At the top of the scale tooltips, can you put something like "Increment font sizes\n\n" for the up arrow, and "Decrement font sizes\n\n" for the down arrow?

robertapplin (changes requested):
This new feature works really well and looks very professional - good job! The code is also written well and has tests. I just have two minor suggestions and then I'm happy to approve

Also I notice the all font size text box is empty when you open the settings, even when all font sizes are the same, but I think this works quite nicely so happy to keep it as is

robertapplin (approved):
Thanks for all your work on this - very happy to approve

PR #951, opened 2023-10-10
941 append scaling to ws name

### LINKED ISSUE #941: Prevent workspace overwriting when scaling during workspace arithmetic
Currently when performing workspace arithmetic, the resultant workspace has the same name regardless of the scaling option chosen.

It would be ideally to append the scaling option chosen to the workspace if it is not default to prevent workspaces with different scaling factors from overwriting each other. 

image

Requested by Adroja.

GuiMacielPereira:
**Summary**
- Created new class that returns name of workspaces for the operations of add, subtract, rebose and scale
- Updated tests with new class, so any changes to naming conventions will not have tests failing

**Description of work:**
- The main observable change is the new format of naming workspaces after subtraction
- Instead of the suffix `_subtracted`, I changed it to include the scaling factor of the background workspace to `_minus_ssf_0.95`, as seen in:
- 
image
- Also altered the scaling suffix to match new naming convention

**Question**
- I altered the tests so that the new class I introduced gets imported into the test and the correct name of the workspace is always used.
- This means that if this class gets changed to include a different naming convention, the tests will continue to pass.
- Is this good practice or should I hard-code the workspace name string into the tests?

**To test:**
- In the workspace manager, load a workspace and try the arithmetic operations:
  - Scale
  - Subtract
  - Rebose
  - Sum
- Give your opinion on whether the naming convention I implemented makes sense
- Look at the new class `WorkspaceNameAppender` and tell me if it's fine (maybe needs an initializer method?)
- Look at the location of the new class (under `workspaces.helperfunctions`), is it the best location or should I move somewhere else?

Fixes #941 .

### DIFF
--- a/src/mslice/models/workspacemanager/workspace_algorithms.py
+++ b/src/mslice/models/workspacemanager/workspace_algorithms.py
@@ -25,6 +25,7 @@ from mslice.util.mantid.mantid_algorithms import Load, MergeMD, MergeRuns, Scale
 from mslice.workspace.pixel_workspace import PixelWorkspace
 from mslice.workspace.histogram_workspace import HistogramWorkspace
 from mslice.workspace.workspace import Workspace
+from mslice.workspace.helperfunctions import WorkspaceNameAppender
 
 from .file_io import save_ascii, save_matlab, save_nexus, save_nxspe
 
@@ -198,18 +199,18 @@ def combine_workspace(selected_workspaces, new_name):
 
 
 def add_workspace_runs(selected_ws):
-    out_ws_name = selected_ws[0] + '_sum'
-    sum_ws = MergeRuns(OutputWorkspace=out_ws_name, InputWorkspaces=selected_ws)
+    sum_ws = MergeRuns(OutputWorkspace=WorkspaceNameAppender.sum(selected_ws[0]),
+                       InputWorkspaces=selected_ws)
     propagate_properties(get_workspace_handle(selected_ws[0]), sum_ws)
 
 
 def subtract(workspaces, background_ws, ssf):
-    scaled_bg_ws = Scale(OutputWorkspace=str(background_ws) + '_scaled',
+    scaled_bg_ws = Scale(OutputWorkspace=WorkspaceNameAppender.scale(str(background_ws), ssf),
                          InputWorkspace=str(background_ws), Factor=ssf, store=False)
     try:
         for ws_name in workspaces:
-            result = Minus(OutputWorkspace=ws_name + f'_subtracted_by_{ssf:.2f}', LHSWorkspace=ws_name,
-                           RHSWorkspace=scaled_bg_ws)
+            result = Minus(OutputWorkspace=WorkspaceNameAppender.subtract(ws_name, ssf),
+                           LHSWorkspace=ws_name, RHSWorkspace=scaled_bg_ws)
             propagate_properties(get_workspace_handle(ws_name), result)
     except ValueError as e:
         raise ValueError(e)
@@ -217,8 +218,8 @@ def subtract(workspaces, background_ws, ssf):
 
 def rebose_single(ws, from_temp, to_temp):
     ws = get_workspace_handle(ws)
-    results = Rebose(InputWorkspace=ws, CurrentTemperature=from_temp, TargetTemperature=to_temp,
-                     OutputWorkspace=ws.name+'_bosed')
+    results = Rebose(OutputWorkspace=WorkspaceNameAppender.rebose(ws.name),
+                     InputWorkspace=ws, CurrentTemperature=from_temp, TargetTemperature=to_temp)
     propagate_properties(ws, results)
     return results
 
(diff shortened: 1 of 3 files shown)


src/mslice/models/workspacemanager/workspace_algorithms.py
```diff
@@ -25,6 +25,7 @@
 from mslice.workspace.pixel_workspace import PixelWorkspace
 from mslice.workspace.histogram_workspace import HistogramWorkspace
 from mslice.workspace.workspace import Workspace
+from mslice.workspace.helperfunctions import WorkspaceNameAppender
```
MialLewis, line 28: Could we rename this `WorkspaceNameModifier` or something, as we will have to prepend, not just append?

MialLewis (changes requested):
This looks good. I can think of a few other instances where this would be useful to use, could you apply them here:
https://github.com/mantidproject/mslice/blob/5af891e298559faef096b834992c4cafc00d8bd4/src/mslice/util/mantid/algorithm_wrapper.py#L55&#L59

Also, we could do with getting rid of checks like this https://github.com/mantidproject/mslice/blob/5af891e298559faef096b834992c4cafc00d8bd4/src/mslice/workspace/helperfunctions.py#L88
Something like WorkspaceNameHander.isHidden(wsName).

If you search for `__MSL` `__HIDDEN` etc. will you will find other instances too.

GuiMacielPereira:
- Put all workspace name changes into the same class
- The workspace names should not have the character '.' and so I replaced it by '_'
    - (If using the character '.' and produce a script, the script will include a `RenameWorkspace` algorithm and change the name of the workspace to something weird) 
- The prefixes `__MSL` and `__MSLTMP` and suffix `_HIDDEN` are not covered by the tests, so I did a visual test:
    - Followed the following procedure:
        - Load ws `MAR21335_Ei60meV`
        - Subtract this workspace with itself with some scaling factor
        - Select the subtracted workspace and plot a Slice plot (using default parameters)
     - Check names of workspace in Mantid:
         - These are the same workspaces that appeared before making changes to the code
         - `__MSL` and `_HIDDEN` are correctly added to ws names
image
- The remaining prefixes and suffixes are covered by the tests
     - I checked that altering one of these in the code will cause some tests to fail
     - My new implementation of previous names did not break old tests
- My changes touched files that produce mslice scritps, so I also did a check that all is good with the scripting
     - I followed the procedure described above for the Slice plot and clicked to produce a script
     - The script matches with the script from before the changes (excepting workspace names that are meant to change)
     - Oddly enough, I detected one very small difference that I can't quite understand:
         - The new script uses the Load() function with the argument for `Filename` as a normal string, whilst the original script before any changes passes this argument as a raw string. I don't think this makes any significant difference.
     - Finally, I tried to run the script I produced on mantid, but this is not possible, as the `mslice.cli` module does not support the `Minus()` algorithm (But this is out of scope for this PR)

src/mslice/workspace/helperfunctions.py
```diff
@@ -116,3 +115,60 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         if self.workspace:
             self.workspace.remove_saved_attributes()
         return True
+
+
+class WorkspaceNameHandler:
```
MialLewis, line 120: I'm surprised there are so many options! I don't think it's ideal to have a separate function for each combination of functions. It's plausible in the future we have to add more, and therefore the number of combination functions will increase exponentially.

I wonder if it's a good idea to instead have a `WorkspaceNameOptions` class, that has a constructor that default sets everything to false via named arguments.

We could then do a combination of functions by doing:
`WorkspaceNameHandler.set_workspace_options(workspace.name, WorkSpaceOptions(hidden_from_mslice=True, hidden_from_ADS=True))`

You would then only need individual functions in the `WorkspaceNameHandler` class, and these would not have to be exposed.

Perhaps when checking we could then do:
`WorkspaceNameHandler.assert_workspace_options(workspace, WorkspaceOptions(hidden_from_mslice=True, hidden_from_ADS=True)`

src/mslice/workspace/helperfunctions.py
```diff
@@ -116,3 +115,60 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def rebosed(self) -> str:
+        return self.add_sufix('_bosed')
+
+    def combined(self) -> str:
+        return self.add_sufix('_combined')
+
+    def merged(self) -> str:
+        return self.add_sufix('_merged')
+
+    def isHiddenFromMsl(self) -> bool:
```
MialLewis, line 149: You've swapped from snake case to camel case here, worth having a flick through these if you haven't already: https://developer.mantidproject.org/Standards/PythonStandards.html

MialLewis (changes requested):
Good work, have done an initial review of implementation. Let me know what you think.

GuiMacielPereira:
Altered class to follow Mial's suggestions.
As Mial is on his sabbatical, here is the full breakdown of the manual test to perform for someone unfamiliar with this PR:

- Load `MAR21335_Ei60meV` into mslice
- On the Slice interface, click `Display`
- These hidden workspaces should appear on the mantid ADS:
image

- Go to the plot window, and click to generate script to clipboard
- Open a new tab on the mantid editor, paste the script and run it to check it produces the same plot
- Now go back to the mslice interface and click to subtract the ws with itself by some scaling factor:
image

- A new hidden workspace should appear on the mantid ADS with the suffix `_HIDDEN`
- Try the other workspace arithmetic operations and check the new naming convention
- Try deleting a workspace in mslice and check it is successfully deleted in the mantid ADS

src/mslice/workspace/helperfunctions.py
GuiMacielPereira: I have now implemented these suggestions, but instead of creating and entirely separate `WorkspaceOptions` class, I preferred to opt for two methods `get_name` and `assert_name` that mimicks this. So in this case one would do `WorkspaceNameHandler(workspace.name).get_name(hidden_from_mslice=True, hidden_from_ADS=True` and `WorkspaceNameHandler(workspace.name).assert_name(hidden_from_mslice=True, hiddden_from_ADS=True)`.

PR #959, opened 2023-11-16
Upgrade to matplotlib 3.7

### LINKED ISSUE #725: Upgrade to matplotlib 3.7
Known issues so far:

`The 'drawtype' parameter of __init__() was deprecated in Matplotlib 3.5 and will be removed two minor releases later. If any parameter follows 'drawtype', they should be passed as keyword, not positionally.`

There are potentially more though.

SilkeSchomann:
**Description of work:**

Upgrade from matplotlib `3.6.*` to `3.7.*`.

The 'drawtype' parameter of `__init__()` was deprecated in matplotlib 3.5 and  removed in 3.7. 

The default parameter for 'drawtype' is `box`, so it was possible to remove this keyword entirely. 

The boilerplate script was updated to the version in matplotlibd 3.7 and a new version of pyplot generated.

**To test:**

The standalone MSlice package can be installed with `mamba install -c mantid/label/matplotlib3_7 mslice`.

Just try to use MSlice, especially different types of plotting.

Fixes #725 .


src/mslice/plotting/pyplot.py
```diff
@@ -2453,8 +2456,8 @@ def hlines(
 @_copy_docstring_and_deprecators(Axes.imshow)
 @set_category(CATEGORY_SLICE)
 def imshow(
-        X, cmap=None, norm=None, aspect=None, interpolation=None,
-        alpha=None, vmin=None, vmax=None, origin=None, extent=None, *,
+        X, cmap=None, norm=None, *, aspect=None, interpolation=None,
```
robertapplin, line 2459: Out of interest, what does the `*` parameter do/mean?

SilkeSchomann: From the [Python 3 documentation](https://docs.python.org/3/reference/compound_stmts.html#function-definitions):

Parameters after “*” or “*identifier” are keyword-only parameters and may only be passed by keyword arguments.

robertapplin (approved):
Plotting seems to be working as expected :shipit:

PR #970, opened 2024-02-05
Update tests with True flag for ADS.clear()

SilkeSchomann:
**Description of work:**

ADS.clear() now has a silent flag to suppress the warning message (https://github.com/mantidproject/mantid/issues/36690) thrown when running this function. It is useful to set this flag when clearing the ADS after unit tests, as it simplifies the clean-up process.

**To test:**

Check that the unit tests were running without problems for this branch: https://github.com/mantidproject/mslice/actions/runs/7757957015


GuiMacielPereira (approved):
- Checked that all tests passed without issues

Issue #975, opened 2024-02-26
Fix deprecated conversion


SilkeSchomann:
The test `test_sample_temperature_list` in `IntensityCorrectionAlgsTest` is now failing. This is caused by the following deprecation warning:  `DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)`

To fix this, `sample_temperature` in `intensity_correction_algs` needs to be updated when trying to convert a list of temperatures to a float.

PR #976, opened 2024-02-26
Fix for deprecation warning

SilkeSchomann:
**Description of work:**

Recently a unit test failed, caused by a deprecation warning. This problem is described in https://github.com/mantidproject/mslice/issues/975 and will be fixed soon. Since testing for the upcoming MSlice release has already finished, the fix needs to be added after the release. Therefore, this deprecation warning is temporarily suppressed in the project configuration.

**To test:**

Check if the corresponding workflow was running successfully: https://github.com/mantidproject/mslice/actions/runs/8046107930

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ where = ["src"]
 [tool.pytest.ini_options]
 pythonpath = ["src"]
 testpaths = ["tests"]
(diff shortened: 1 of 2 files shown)


MialLewis (approved):
Test no longer fails with depreciation warning so this temporary fix clearly works!

PR #978, opened 2024-02-26
Permanent fix for deprecation warning

### LINKED ISSUE #975: Fix deprecated conversion

SilkeSchomann:
**Description of work:**

This is a permanent fix for a deprecation warning that occurred in the unit tests. Conversion of an array to a scalar is deprecated in NumPy 1.25. In the function sample_temperature in the intensity correction algorithms, we attempted to convert a one-dimensional array into a float. This triggers now the deprecation warning. The solution is to first check for lists and arrays before attempting to convert a single value to a float.

**To test:**

1. Load `MAR21335_Ei60meV` dataset 
2. Display slice
3. On the intensity dropdown select `Chi''(Q,E)`
4. Enter `80` in combo box and click ok
5. On the intensity dropdown select `GDOS`
6. There should be no pop up window asking for the sample temperature

Also, check that the unit tests were running without failure for this branch: https://github.com/mantidproject/mslice/actions/runs/8047878035

Fixes #975.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ where = ["src"]
 [tool.pytest.ini_options]
 pythonpath = ["src"]
 testpaths = ["tests"]
-filterwarnings = ["error", "ignore::DeprecationWarning"]
+filterwarnings = ["error"]
 
 [tool.coverage.report]
 include = [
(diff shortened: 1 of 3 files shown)


MohamedAlmaki (approved):
LGTM and works as expected

PR #982, opened 2024-03-26
Fix for failing unit test workflows

SilkeSchomann:
**Description of work:**

All unit test workflows fell over for MSlice, although there were no changes made to the MSlice main branch (which is used for the nightly unit test workflow). Switching from `mamba install` to `conda install` can prevent this. 

I have also upgraded the conda-incubator from version 2.2.0 to version 3.

**To test:**

Check that the GitHub workflow for the unit tests does not fail for this branch: https://github.com/mantidproject/mslice/actions/runs/8433683620

### DIFF
--- a/.github/workflows/deploy_conda.yml
+++ b/.github/workflows/deploy_conda.yml
@@ -18,7 +18,7 @@ jobs:
           fetch-depth: 0
 
       - name: Setup Miniconda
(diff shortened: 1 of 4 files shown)


robertapplin (approved):
Looks good to me!

PR #989, opened 2024-06-12
Added None as a line option

### LINKED ISSUE #981: Feature request: Add "none" option to line style
Original reporter: ISIS / M. Aouane

Sometimes users do not necessarily want to plot a line spectra - just the points or errorbars. MSlice allows users to customize the plot using the "Cogs" (Settings) icon which brings up another window. In this window, on the right hand side is a box which contains all traces in the plot with options to change the `Color`, `Style` etc - for line `Style` a user requests that there be an addition option `None` which will just plot the markers and not the line (the `Show Line` check box governs whether the whole trace (line and marker) is shown or not).

SilkeSchomann:
**Description of work:**

Added a new line style, "None". Here, only the error bars are plotted and not the line. 

**To test:**

2. Do a cut
3. Click on the settings icon and change the line style to 'None'
4. Observe that the cut only shows error bars and no line
5. Change the line style to any of the other options
6. Check that the line on the plot is shown with the corresponding line style

Fixes #981

### DIFF
--- a/src/mslice/plotting/plot_window/plot_options.py
+++ b/src/mslice/plotting/plot_window/plot_options.py
@@ -314,7 +314,7 @@ class LegendAndLineOptionsSetter(QtWidgets.QWidget):
     """This is a widget that has various legend and line controls for each line of a plot"""
 
     # dictionaries used to convert from matplotlib arguments to UI selection and vice versa
-    styles = {'-': 'Solid', '--': 'Dashed', '-.': 'Dashdot', ':': 'Dotted'}
+    styles = {'-': 'Solid', '--': 'Dashed', '-.': 'Dashdot', ':': 'Dotted', 'None': 'None'}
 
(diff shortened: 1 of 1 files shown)


adriazalvarez (approved):
Following test instructions, now there's the option to plot `None` and then it only plots the scatter plot and error bars. 
I'm happy to approve. 

Additionally, maybe I have missed it the option, but it seemed you have to change line options individually for each line represented in the plot. It would be nice future feature to have the choice to change all the line options at once.

SilkeSchomann:
> Following test instructions, now there's the option to plot `None` and then it only plots the scatter plot and error bars. I'm happy to approve.
> 
> Additionally, maybe I have missed the option, but it seemed you have to change line options individually for each line represented in the plot. It would be nice future feature to have the choice to change all the line styles at once.

@adriazalvarez This is an excellent suggestion. I will raise an issue for this.

PR #994, opened 2024-07-10
Omit spectrum ID when saving as ASCII

### LINKED ISSUE #993: Don't write the spectrum ID when saving a 2D workspace as ASCII
Currently, the spectrum ID is written below the header of an ASCII file when saving a workspace directly from the 2D tab. For single spectrum workspaces, this information is not required, so it would be better to omit it.

SilkeSchomann:
**Description of work:**

The spectrum ID is now omitted when saving a 2D workspace as an ASCII file.

**To test:**

2. Select the workpace in the 2D tab of the workspace manager
3. Click the Save button and select ASCII
4. Check that the second line remains empty

Fixes #993.


robertapplin (approved):
I can confirm the 2nd line in the saved ASCII file remains empty with this change 👍

PR #1024, opened 2024-10-25
Added section on how to save multiple files via CLI

### LINKED ISSUE #1023: Saving multiple cuts via command-line interface
When a user needs to save several cuts as ASCII files, it is often easier to use the command-line interface with save_ascii instead of the GUI. It would be good to have a section in the MSlice documentation describing this, preferably with an example script.

SilkeSchomann:
**Description of work:**
Added an example script for saving multiple cuts as ASCII files. 

Also, corrected an import error in the section "Plotting a series of cuts".

**To test:**

Read instructions and see whether they make sense. Try out the example script.

Fixes #1023.

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -177,7 +177,7 @@ Plotting a series of cuts
 
 .. code:: python
 
-    import mslice.cli as m
+    import mslice.cli as mc
 
     # Plot a series of energy cuts at different Q (similar to putting something in the width box in GUI).
     ws = mc.Load('data.nxspe')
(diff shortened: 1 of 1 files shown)


GuiMacielPereira (changes requested):
I just wanted to check that when I run the script, it successfully creates and saves the cuts, but if I try to plot the cuts in mantid, the legend is not visible and I get the following warning:
```
The label '__MSLMAR25087_Ei40.00meV_cut(1.500,2.500)' of <matplotlib.lines.Line2D object at 0x7d490aaed360> starts with '_'. It is thus excluded from the legend.
```

This is fine, right?

SilkeSchomann:
This is weird, I cannot reproduce it. Which steps are you following when plotting the cuts?

GuiMacielPereira:
I was just running the script inside Mantid and then double clicking on the workspaces to plot

SilkeSchomann:
The workspace in MSlice or in Mantid?

GuiMacielPereira:
In Mantid :thinking:

SilkeSchomann:
In that case, it is not a problem. It means you are displaying hidden workspaces in Mantid and the warning is useful in case a user does the same. These workspaces are all accessible through the MDHisto tab in MSlice after running the script. There they are not hidden, and the warning does not show up when the user clicks the `Plot` button.

GuiMacielPereira (approved):
I just confirmed that the workspaces correctly appear on mslice :+1:

PR #1034, opened 2024-11-22
Pin setuptools to 75.1.0

SilkeSchomann:
**Description of work:**
Building MSlice with setuptools > 75.1.0 fails at the moment, so this PR pins setuptools to the last verified working version.

**To test:**

Code review only.


GuiMacielPereira (approved):
Changes make sense.

PR #1041, opened 2024-12-10
Added ADSObserver to MSlice

### LINKED ISSUE #839: User issues ADS.clear() in script and breaks MSlice
**Describe the bug**
I'm not sure this is a bug in MSlice per-se, but because MSlice workspaces are just wrappers around hidden Mantid workspaces (stored in the ADS in order to allow the history functionality to work), users don't necessarily understand that the Mantid workspaces exist. This manifested in a recent bug where the user had `ADS.clear()` at the start of a script (presumably to clean up unwanted temporary workspaces). This meant, however, that all the MSlice workspaces were also deleted (but as they were hidden, the users did not realise they were there), and MSlice to stop working after the script was run.

This issue or its solution may be related to #737

**Possible solutions**

The most basic fix is to make a check when the workspace manager is activated to see if all the underlying ADS workspaces are still available. If not, to delete the corresponding entry in the workspace manager. Alternatively this check could be done on a per-workspace basis when an MSlice workspace is accessed (e.g. for a cut/slice) - in this case, it would be helpful to flash a message at the bottom of the gui saying the underlying Mantid workspace was deleted.

If we do check the full list of workspaces when the workspace manager is accessed (this might impose too much of a performance overhead if the user has a lot (~>50?) workspaces opened), and we see that all the workspaces have been deleted, we can issue a pop-up warning saying that we think that `ADS.clear()` was run which caused all the workspaces MSlice relies on to be deleted.

Finally, the most drastic solution is to go back to using non-ADS Mantid workspaces as was the design around version 1.x (?) but this would necessitate additional code (perhaps a lot of additional code) to handle the history / script generation capability.

SilkeSchomann:
**Description of work:**
So far MSlice interacted with the ADS in a one-sided way. Some of its workspaces were stored in the ADS and occasionally deleted, but changes to the ADS from the Mantid side ignored. This sometimes led to missing workspaces in MSlice when the respective workspace was deleted on the Mantid side. However, the respective workspace would still get displayed in MSlice and cause a crash when selected.

This PR introduces an ADS observer that modifies MSlice workspaces on delete, clear and rename, respectively. 
In addition, two small bugs were fixed that became obvious when testing these changes:
1. When taking an interactive cut and closing the interactive cut window without saving the cut to the workspace, a cut was showing up in the MD Histo tab. 
2. When clearing all workspace, there were still workspaces shown in the MD Histo tab that were not available anymore and caused crashes on selection. This was particularly visible after a clear() from the ADS that should cause all workspaces in all tabs to get cleared.

**To test:**

With workspaces saved to the ADS from the MSlice interface, check that the ADS commands clear, delete and rename work as expected in MSlice when run from Mantid.

Fixes #839.

### DIFF
--- a/src/mslice/models/mslice_ads_observer.py
+++ b/src/mslice/models/mslice_ads_observer.py
@@ -1,0 +1,77 @@
+from functools import wraps
+import sys
+
+from mantid.api import AnalysisDataServiceObserver
+
+
+def _catch_exceptions(func):
+    """
+    Catch all exceptions in method and print a traceback to stderr
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except Exception:
+            sys.stderr.write("Error occurred in handler:\n")
+            import traceback
+
+            traceback.print_exc()
+
+    return wrapper
+
+
+class MSliceADSObserver(AnalysisDataServiceObserver):
+    def __init__(self, delete_callback, clear_callback, replace_callback, rename_callback):
+        super(MSliceADSObserver, self).__init__()
+        self.delete_callback = delete_callback
+        self.clear_callback = clear_callback
+        self.replace_callback = replace_callback
+        self.rename_callback = rename_callback
+
+        self.observeDelete(True)
+        self.observeRename(True)
+        self.observeReplace(True)
+        self.observeClear(True)
+
+    @_catch_exceptions
+    def deleteHandle(self, workspace_name, workspace):
+        """
+        Called when the ADS deletes a workspace, removes it from the dict of tracked workspaces.
+        :param workspace_name: name of the workspace
+        :param workspace: reference to the workspace (not used)
+        """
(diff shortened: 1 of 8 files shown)


GuiMacielPereira (approved):
Code is well written and easy to understand. Unit tests were added.
I tested this change manually by renaming, clearing and deleting workspaces in the ADS and confirmed that they are correctly updated in mslice. I tested that when closing an interactive cut without saving the workspace is not accidentally saved to the Histo tab; Checked that the clear button does clear all workpaces in ADS and mslice (even the hidden ones). 

I also did some cuts and workspace operations, saved them to the ADS from mslice, renamed and deleted them and checked that they were successfully being updated back into mslice.

I also noticed that renaming or deleting hidden mslice (`__MSL`) workspaces in the ADS does not update the workspace on mslice, and might cause some issues. However, these workspaces are hidden because the user is not supposed to interact with them in the ADS anyway (and it's something that can be dealt with in another PR)

PR #1043, opened 2024-12-10
Removed obsolete __future__ imports

### LINKED ISSUE #1042: Remove obsolete __future__ imports
There are a few __future__ imports in MSlice that are not necessary anymore since we are only using Python 3 and not 2:

- division
- print_function
- absolute_import
- unicode_literals

These obsolete __future__ imports should get removed.

SilkeSchomann:
**Description of work:**
Removed __future__ imports that are obsolete since we are not using Python 2 anymore: https://docs.python.org/3/library/__future__.html

**To test:**

Check that tests are passing. Also, the only remaining __future__ import should be the following in pyplot.py:
`from __future__ import annotations`

Fixes #1042.


GuiMacielPereira (approved):
Checked that mslice opens without errors or warnings about imports :+1:

PR #1045, opened 2025-01-10
Improve deletion of workspaces that are in the ADS

### LINKED ISSUE #1044: Mantid crash when deleting MD Histo workspace
**Describe the bug**
Mantid crashes when an MD Histo workspace that was saved to Mantid is deleted.

**To Reproduce**
1. Load some data
2. Create a cut
3. Navigate to MD Histo tab
4. Save MD Histo workspace to Mantid
5. Delete MD Histo workspace in MSlice

SilkeSchomann:
**Description of work:**
Deleting a workspace within MSlice that is also in the ADS caused a crash in Mantid. The reason was that deleting the workspace in MSlice updated the ADS and, as a result, the ADS triggered a deletion request for this workspace in MSlice. This request then failed as the workspace was already deleted within MSlice. To avoid this situation, MSlice does not attempt to delete non-existing workspaces anymore. In addition, all corresponding hidden workspaces are now deleted from the ADS as well.

**To test:**

Follow the instructions on the original issue.

Fixes #1044.

### DIFF
--- a/src/mslice/models/workspacemanager/workspace_provider.py
+++ b/src/mslice/models/workspacemanager/workspace_provider.py
@@ -46,7 +46,10 @@ def get_workspace_name(workspace):
 
 
 def delete_workspace(workspace):
-    workspace = get_workspace_handle(workspace)
+    try:
+        workspace = get_workspace_handle(workspace)
+    except KeyError:
+        return None
     remove_workspace(workspace)
     del workspace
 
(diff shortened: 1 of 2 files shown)


GuiMacielPereira (approved):
Deleting a workspace in the MD Histo tab no longer crashes Mantid :+1: .

The only thing I noticed is that if I follow the same steps but click `rename` instead of `delete`, then the workspace in the ADS is not correctly renamed, but I think this can go in a different PR

PR #1047, opened 2025-01-24
Catch unhandled exceptions

SilkeSchomann:
**Description of work:**
When testing the new ADS observer, unreliable unhandled exceptions occurred when following testing instructions in https://developer.mantidproject.org/Testing/Direct/MSliceTestGuide.html#interaction-with-ads. This PR catches these exceptions to avoid crashes.

**To test:**

Replace the MSlice version in the current Mantid nightly and follow the instructions in the issue below.

Fixes #[38675](https://github.com/mantidproject/mantid/issues/38675)

(Since this is in a different repo, we will have to manually close this Mantid issue.)

### DIFF
--- a/src/mslice/models/workspacemanager/workspace_provider.py
+++ b/src/mslice/models/workspacemanager/workspace_provider.py
@@ -19,12 +19,18 @@ def add_workspace(workspace, name):
 
 
 def remove_workspace(workspace):
-    workspace = get_workspace_handle(workspace)
+    try:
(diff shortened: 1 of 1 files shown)


GuiMacielPereira (approved):
Code is straight forward and I believe this fixes the problem.
I spent quite some time (20-30 minutes) trying different combinations of renaming and deleting both from the ADS and from MSlice, and didn't encounter any crash.

PR #1059, opened 2025-03-14
Remove invisible workspace from MSlice when deleted in ADS

### LINKED ISSUE #1050: Deleted mantid backing-workspace
**Describe the bug**

MSlice workspaces are just a thin wrapper around a Mantid workspace. In order to allow workspace history to work these have to be kept in the analysis data service (ADS) registry. However, this means that a user in Mantid can delete the backing workspace without recourse to MSlice so that the MSlice workspace no longer has access to its data. This most commonly happens when they do `mtd.clear()`.

Previously we could only issue a warning to users not to do this but since the new `ADSObserver` was [added](https://github.com/mantidproject/mslice/pull/1041) to MSlice we can now handle this issue - currently deleting a workspace in MSlice will delete the underlying (hidden) workspace in Mantid, **and** an `mtd.clear()` will result in the corresponding MSlice workspaces being deleted too.

However, manually deleting the hidden Mantid workspace still leaves the MSlice workspace intact and will cause an error if the user tries to access the MSlice workspace (or plot it etc).

This is an edge case though and it might not be worth fixing (it requires users to enable `Show Invisible Workspaces` _and_ then to delete the `__MSL` prefixed workspace(s).

**To Reproduce**
1. In Mantid File->Settings->General check `Show Invisible Workspaces`
2. Open a file in MSlice. See that a hidden Mantid workspace `__MSL<filename>` is created.
3. In the Mantid window, delete the hidden `__MSL<filename>` workspace
4. In MSlice click on the loaded workspace
5. Sad Mantid appears

**Expected behavior**
When the hidden Mantid workspace is deleted the MSlice workspace is also deleted.

**Screenshots**
N/A

 - MSlice Version `2.10.1.dev29`
 - Mantid nightly `6.11.20250131.1730`

**Additional context**

SilkeSchomann:
**Description of work:**
Added code that allows the user to delete a hidden workspace in the ADS, so that the corresponding workspace is deleted in MSlice as well. This prevents crashes when this workspace is accessed again after deletion. Before, this was only possible with visible workspaces.

**To test:**

Follow the steps to reproduce this problem in the original issue.

Fixes #1050.

### DIFF
--- a/src/mslice/presenters/workspace_manager_presenter.py
+++ b/src/mslice/presenters/workspace_manager_presenter.py
@@ -271,6 +271,8 @@ class WorkspaceManagerPresenter(WorkspaceManagerPresenterInterface):
         self._workspace_manager_view.clear_displayed_error()
 
     def delete_handle(self, workspace):
+        if workspace.startswith("__MSL"):
+            workspace = workspace[5:]
         delete_workspace(workspace)
         self.update_displayed_workspaces()
 


GuiMacielPereira (approved):
Followed instructions on original issue and checked that deleting a hidden workspace (starting with '__MSL') does not cause any issues and the workspace is successfully deleted from both mantid and mslice.

PR #1061, opened 2025-03-18
Changed layout of Compose button

SilkeSchomann:
**Description of work:**
The `Compose` button now has a similar layout as the `Save` button. This should make it easier to select a compose action on some operating systems, for instance, macOS, where the selection was sometimes difficult due to a squashed layout.

**To test:**

Follow the instructions in this Mantid issue: https://github.com/mantidproject/mantid/issues/38671


GuiMacielPereira (approved):
Compose button now has the save layout as the save button, making the dropdown easier to spot and more consistent with the interface. The code is also improved by removing unecessary chunks

PR #1067, opened 2025-05-09
Fix for bug in width calculation

SilkeSchomann:
**Description of work:**
Due to rounding errors MSlice sometimes assumed a set width for producing more than one cut when only one cut was intended. This has been fixed by setting the width to None if no value was provided.

**To test:**

Fixes #[39285](https://github.com/mantidproject/mantid/issues/39285) - Might need to get closed manually as it is a Mantid issue.

### DIFF
--- a/src/mslice/models/cut/cut.py
+++ b/src/mslice/models/cut/cut.py
@@ -143,7 +143,7 @@ class Cut(object):
                 except ValueError:
                     raise ValueError("Invalid width")
         elif width_str == "":
-            self._width = self.end - self.start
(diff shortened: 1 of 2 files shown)


mducle (approved):
Looks good and I confirm it fixes the issue.

I think that with this, the code where the original rounding error occurs ([lines 145-146 of models/cut.py](https://github.com/mantidproject/mslice/blob/dc35905cf93191d3f4b4f92178bf4cdca41eb2fe/src/mslice/models/cut/cut.py#L145-L146)) is now unreachable and should probably be removed...

PR #1069, opened 2025-05-09
Removed unused code

### LINKED ISSUE #1068: Clean up code for width calculation
A recent fix (https://github.com/mantidproject/mslice/pull/1067) made some code for the width setter for plots redundant. These lines should get removed.

SilkeSchomann:
**Description of work:**

Cleaned up code that became redundant after a recent bug fix.

**To test:**

Code review only (see https://github.com/mantidproject/mslice/pull/1067)

Fixes #1068.

### DIFF
--- a/src/mslice/models/cut/cut.py
+++ b/src/mslice/models/cut/cut.py
@@ -142,8 +142,6 @@ class Cut(object):
                     self._width = float(width_str)
                 except ValueError:
                     raise ValueError("Invalid width")
-        elif width_str == "":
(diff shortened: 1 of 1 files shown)


GuiMacielPereira (approved):
This deletion makes sense because when `width_str` is `""`, the value is not used anyway (https://github.com/mantidproject/mslice/pull/1067/files#diff-6193576ecd93b653cfc20729b1a0c9c9b5f203521e109bd4c758df081f6e17d6R149-R150)

PR #1074, opened 2025-05-19
Update to Python 3.11

SilkeSchomann:
**Description of work:**
After the recent update to Python 3.11 in Mantid, MSlice is now also using Python 3.11.

**To test:**

Check that tests are passing.


GuiMacielPereira:
The tests pass and it's acceptable that only one test is disable for now :+1:

GuiMacielPereira (approved):
Tests pass and its acceptable that only one test is disabled for now :+1:

PR #1078, opened 2025-05-20
Fix for crash when changing GDOS on interactive cut

### LINKED ISSUE #1071: Exception thrown when changing intensity to GDOS on interactive cut plot
**Describe the bug**
Changing the intensity to GDOS for an interactive cut causes an exception in MSlice.

**To Reproduce**
1. Load a data set.
2. Display a slice.
3. Click on the 'Interactive Cuts' button.
4. Try to change intensity to GDOS.
5. Observe MSlice crash.

 - MSlice Version: currently nightly (not present in released versions)

SilkeSchomann:
**Description of work:**
This fix includes two changes. The intensity on an interactive cut plot can only be changed when the `Interactive Cuts` button on the original slice plot is toggled. This had the side-effect that the underlying cut workspace got deleted and was missing when attempting to change the intensity. Now this workspace remains available when the cut plot is still open. 
The second fix improves the functions that show a corrected intensity for a plot. The dictionary used to store the cut plots utilises `MSliceAxes` objects as keys. This sometimes caused `KeyErrors`, so these functions are now more robust when searching for the value for a key.

**To test:**

Fixes #1071.


src/mslice/plotting/plot_window/slice_plot.py
```diff
@@ -521,7 +521,7 @@ def _update_lines(self):
         self._canvas.draw()
 
     def toggle_interactive_cuts(self, store=True):
-        self.toggle_icut_button(store)
+        self.toggle_icut_button(True)
```
GuiMacielPereira, line 524: Maybe delete the `store` argument since it's not used by the function anymore? And update the places where this function is being called with the argument?

GuiMacielPereira (changes requested):
I followed the instructions on the original issue and now when `Interactive Cuts` is toggled the button for the `Intensity` is grey out, which means the user is not able to change intensity i.e. does not trigger the crash.

I have only a small suggestion to make the code a bit clearer.

GuiMacielPereira (approved):
The changes make the code a bit cleaner. I tested it again manually and everything is working as expected :+1:

PR #1082, opened 2025-05-29
38673 fix overplot exception

SilkeSchomann:
**Description of work:**
This PR fixes an unhandled exception that occurred when overplotting a cut with changed intensity and a Bragg peak.

**To test:**

Please note that the Bragg peak is not supposed to get displayed after plotting over.

Fixes #[38673](https://github.com/mantidproject/mantid/issues/38673).


GuiMacielPereira (approved):
Followed manual testing instructions and checked that taking a cut > change intensity > plot bragg peak > overplot does not cause a crash. Instead the plot correctly returns to its original state (without the bragg peak).

Also great that there is a new unit test :+1:

PR #1089, opened 2025-06-05
Prevent crash for invalid characters

### LINKED ISSUE #1087: Crash when entering special character
**Describe the bug**
While testing, we experienced crashes when entering special characters. On the cut tab, enter a `-` into one of the boxes for Intensity. For `Compose`, the same happens when entering a `-` into `Scale` or `Bose`. This is not a regression.

SilkeSchomann:
**Description of work:**
During manual testing, a few crashes were caused by entering invalid characters into the boxes for `Bose` and `Scale` for `Compose` as well as for `Intensity` on the `Cut` tab.

**To test:**

Enter a special character into the boxes mentioned above and observe no crash, but a message warning about the invalid input. In case for the the` Intensity`, the message differentiates between invalid entries for the start and end value.

Fixes #1087

### DIFF
--- a/src/mslice/widgets/workspacemanager/input_boxes.py
+++ b/src/mslice/widgets/workspacemanager/input_boxes.py
@@ -50,4 +50,16 @@ class ScaleInputBox(QDialog):
     def user_input(self):
         v1 = self.edit1.text()
         v2 = self.edit2.text()
-        return float(v1) if v1 else None, float(v2) if v2 else None
+        v1_value = None
+        v2_value = None
+        if v1:
+            try:
+                v1_value = float(v1)
+            except ValueError:
+                pass
+        if v2:
+            try:
+                v2_value = float(v2)
+            except ValueError:
+                pass                
+        return v1_value, v2_value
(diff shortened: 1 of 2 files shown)


src/mslice/widgets/workspacemanager/input_boxes.py
```diff
@@ -50,4 +50,16 @@ def __init__(self, is_bose=False, parent=None):
...
+            try:
+                v1_value = float(v1)
+            except ValueError:
+                pass
+        if v2:
+            try:
+                v2_value = float(v2)
+            except ValueError:
+                pass
+        return v1_value, v2_value
```
mducle, line 65: As an alternative, this has fewer lines...

```suggestion
        try:
            v1_value = float(v1) if v1 else None
        except ValueError:
            v1_value = None
        try:
            v2_value = float(v2) if v1 else None
        except ValueError:
            v2_value = None
        return v1_value, v2_value
```

mducle (approved):
Looks good and stops the crashes in the original issue.

But, for the cuts - I don't see the warnings (in lines 121 / 126) in the UI at all (either run from the workbench or independently). I think some other parts of the code is just swallowing the warnings... but that's no reason to stop this PR.

mducle:
> The warnings are only displayed in the Message window of Mantid.

Ah right... yes I see them now - I had my logging set to `Error`...

I'm not sure if it's a regression but now if the boxes are empty (the default) it will give the warnings whereas previously there was no warnings (it just plots the limits of the data - now it plots the limits of the data _and_ gives a warning if the input is either empty or just contains the characters `-`, `+`, `.` without any numbers.

PR #1096, opened 2025-07-16
Updated MSlice quickstart guide to reflect it as an optional dependancy of conda installs

warunawickramasingha:
**Description of work:**
`MSlice` is turned into an optional dependency of `mantidworkbench`. However, MSlice would still be installed explicitly in IDAaaS and inside standalone packages. The quickstart guide of `MSlice` is to be updated to reflect those changes.

**To test:**
Check that the updated quickstart guide reads well.

Fixes #39237.


SilkeSchomann (approved):
The addition to the documentation reads well and reflects the changes regarding the MSlice packaging in Mantid.

PR #1107, opened 2025-09-16
Fix for save script error

### LINKED ISSUE #1105: Save script to clipboard error
**Original reporter:** Duc (with an error report)

**Describe the bug**

Generate a script from an MSlice plot and a `TypeError` appears

**To Reproduce**
- `Interfaces` -> `Direct` -> `MSlice`
- Load file, e.g. `MAR21335_Ei60meV.nxs` from the ISIS Sample Data
- Make a cut along energy, e.g. in the `Cut` tab cut along `deltaE`, generate plot
- On the plot, `File` -> `Generate Script to Clipboard`

**Expected behavior**
Script correctly copied to clipboard

**Platform/Version (please complete the following information):**
 - OS: Windows (probably all though)
 - OS Version: 11
 - Mantid Version 6.13.1

SilkeSchomann:
**Description of work:**

Prevent addition of `None` values for width when calculating the cut end. For the issue the crash happened when creating a script from a cut plot. However, the same problem could also occur in the cut plot presenter, so I've applied the same fix there as well.

**To test:**

Please follow the instructions in the original issue. In addition, try to use float values in the `width` box.

Fixes #1105

### DIFF
--- a/src/mslice/models/alg_workspace_ops.py
+++ b/src/mslice/models/alg_workspace_ops.py
@@ -13,8 +13,10 @@ def get_number_of_steps(axis):
 
     return int(np.ceil(step_num))
 
+
 def get_range_end(start, end, width):
     return min(start + width, end) if width is not None else end
+
 
 def get_axis_range(workspace, dimension_name):
     workspace = get_workspace_handle(workspace)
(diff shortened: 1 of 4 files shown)


warunawickramasingha (approved):
Thanks for the changes and happy to approve!

PR #1110, opened 2025-09-24
Cache the Temperature Log Entered for Use in Subsequent Slices

### LINKED ISSUE #877: Log sample used to set temperature in slice and give option use for subsequent slices
**Describe the bug**
We want to be able to cache the log used to set the temperature of a slice, and use it for future slices. It should have the following behaviour:

- If selected, option to cache the log used and use this for subsequent slices.
- If subsequent slice does not have the log, ask for temperature again. Option to again overwrite the log cache.
- This should probably come with a menu option somewhere to manually be able to change the temperature of a slice.

There was previously a feature like this but it doesn't work/wasn't fully implemented.

cailafinn:
**Description of work:**

Adds a temperature log cache to the slice presenter such that subsequent slices will, by default, use that log for setting the temperature of the slice. 

A menu item to allow this option to be set manually has also been added and is available when an appropriate Intensity is selected.

**To test:**
1. Load `MAR21335_Ei60meV` from the ISIS Sample Data
2. Swap to Workspace Manager and click `Display`
3. Change the Intensity in the menu bar to `Chi''(Q,E)`
4. Check that the `Set Temperature Log` option starts greyed out.
5. Select `moderator_temp` and tick `Overwrite Cached Value`.
6. Close the plot.
7. Click `Display` again. 
8. Select `Intensity` -> `Chi''(Q,E)` again.
9. No dialog should appear, as it is now automatically using the previously selected log.
10. Check that `Intensity` -> `Set Temperature Log` is now enabled and can be used to open the log selection dialog.

Fixes #877

### DIFF
--- a/src/mslice/plotting/plot_window/slice_plot.py
+++ b/src/mslice/plotting/plot_window/slice_plot.py
@@ -69,6 +69,7 @@ class SlicePlot(IPlot):
             "colorbar_label_size",
             "colorbar_range_font_size",
         ]
+        self.plot_window.set_manual_temp_log_enabled(False)
 
     def save_default_options(self):
         self.default_options = {
@@ -426,12 +427,14 @@ class SlicePlot(IPlot):
             cbar_range = self.colorbar_range
             title = self.title
             if temp_dependent:
+                self.plot_window.set_manual_temp_log_enabled(True)
                 if not self._run_temp_dependent(slice_plotter_method, previous):
                     self.manager.reset_current_figure_as_previous(
                         last_active_figure_number, disable_make_current_after_plot
                     )
                     return
             else:
+                self.plot_window.set_manual_temp_log_enabled(False)
                 slice_plotter_method(self.ws_name)
             self.update_canvas(cbar_range, cbar_log, title)
         else:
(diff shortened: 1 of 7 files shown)


src/mslice/plotting/plot_window/slice_plot.py
```diff
@@ -29,6 +29,8 @@
 
 from typing import Callable
 
+from plotting.plot_window.cachable_input_dialog import QCacheableInputDialog
```
SilkeSchomann, line 32:
```suggestion
from mslice.plotting.plot_window.cachable_input_dialog import QCacheableInputDialog
```
On Windows, MSlice could not find QCacheableInputDialog when using `plotting.plot_window.cachable_input_dialog import QCacheableInputDialog`

SilkeSchomann (changes requested):
I have tested the functionality and reviewed the code. It works very well and the code looks good. Also, it is great to see the test coverage.

I've only seen two small issues:
- I've left a comment for the import in slice_plot.py
- The Set Temperature Log menu item also shows up for Cut plots, but does not work there. It would be better to make it invisible for Cut plots.

SilkeSchomann (approved):
Thanks for making the requested changes. I've tested again and everything works fine now.

PR #1114, opened 2025-10-13
Fix for incorrect errorbars for GDOS corrected cuts with a cm-1 energy unit

### LINKED ISSUE #1070: incorrect errors when computing GDOS with units of cm-1
**Describe the bug**
After taking a cut in cm-1, integrating in q when you convert the intensity to gdos it massively overestimates the error bars.

**To Reproduce**
Load a dataset into mslice,
take a cut which integrates over q and has energy units of cm-1
when the plot opens click the intensity tab and choose GDOS from the dropdown
do the same but choose energy units of meV. Compare the relative error bars.

**Expected behavior**
The % error in each point should be the same whether you plot in cm or meV.

mantid workbench 6.12.0

SilkeSchomann:
**Description of work:**
When the energy units for a cut are changed from meV to cm-1, for a GDOS intensity correction the underlying slice workspace has to be scaled accordingly and the errors calculated on base of the scaled workspace.

**To test:**

Follow the instructions of the original issue for testing.

Fixes #1070

### DIFF
--- a/src/mslice/models/intensity_correction_algs.py
+++ b/src/mslice/models/intensity_correction_algs.py
@@ -103,7 +103,6 @@ def generate_modification_array(multiplier, up_to_index, signal):
 
 def slice_compute_gdos(scattering_data, sample_temp, q_axis, e_axis, rotated):
     x_units = e_axis.units if rotated else q_axis.units
-
     x_dim, y_dim = _get_slice_dimensions(scattering_data, x_units)
     x_dim_shape_index = (
         0 if x_dim.name == scattering_data._raw_ws.getXDimension().name else 1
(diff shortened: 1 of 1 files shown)


mducle:
@SilkeSchomann The problem turns out to be the `ScaleX` algorithm which is used to change the energy axis values - it doesn't know how to handle the fractional areas in the `RebinnedOutput` 2D slice intermediate workspace of the GDOS calculation and doesn't scale these correctly with the new x-axis bin sizes. This also means that DOS calculations are also slightly off because the `RebinnedOutput` rescales the "display" intensities by the fractional areas (which are incorrect when we use ScaleX to convert to `cm-1`).

Edit: The long term fix is to change the C++ `ScaleX` algorithm to handle `RebinnedOutput` but for now, a pure-python work-around for MSlice is to do the GDOS computation in `meV` and at the end to reconvert the energy axis back to `cm-1`.

With this code, the GDOS output for `meV` and `cm-1` is exactly the same (y-values and e-values) - which is the same behaviour with all other "Intensity Corrections" like "chi" etc.

GuiMacielPereira (approved):
@SilkeSchomann tested it manually and changes in the code are clear. Having the contribution from @mducle makes me happy to approve.

PR #1117, opened 2025-10-16
Fix versioning for release versions

SilkeSchomann:
**Description of work:**

There were two checks for `version_number != '0'`, but after the first one` version_number` was already changed, so the second check would always end with the same result.

**To test:**

Unfortunately the only way to test this is to check the version number on Anaconda after the release, so for now it has to be code review only.


GuiMacielPereira (approved):
The code makes sense, let's hope it works

PR #1142, opened 2026-01-12
Close plt window when corresponding workspace gets deleted 

### LINKED ISSUE #988: Close plot window when corresponding MDHisto workspace was deleted
**Describe the bug**
After deleting a workspace from the MDHisto tab while the corresponding plot window is still open, a change to the intensity causes a crash as the underlying workspace is not available anymore.

**To Reproduce**

2. Do a cut
3. Select the corresponding workspace in MDHisto and click the Delete button
4. Change the intensity in the plot window
5. Mantid error

**Expected behavior**
The plot window should get closed when the workspace is deleted in the MDHisto tab.

warunawickramasingha:
**Description of work:**
This PR fixes a bug in mslice that was occuring when some operation is performed on a plot windows after deleting the underlying workspace. This issue has been identifyed after deleting the workspace from the MDHisto tab. However the same issue can be seen from other tabs as well. 

**To test:**
Follow either of below steps

-  Original recreation steps
2. Do a cut or simply display the loaded workspace
3. Select the corresponding workspace in MDHisto and click the Delete button
4. The plot window should be closed elegantly

-  Alternative recreation steps
2. Display the loaded workspace
3. Select the corresponding workspace in 2D tab and click the Delete button
4. The plot window should be closed elegantly

Fixes #988.

### DIFF
--- a/src/mslice/plotting/globalfiguremanager.py
+++ b/src/mslice/plotting/globalfiguremanager.py
@@ -26,7 +26,9 @@ be added to the category of the command
 
 Currently there are only two categories ('1d' and '2d') hard coded into the manager.
 """
+
 from collections import defaultdict
+
 # system imports
 from functools import wraps
 
@@ -523,7 +525,9 @@ class GlobalFigureManager(object):
         plotted_windows = defaultdict(list)
         for plot_fig_man in cls.all_figures():
             if plot_fig_man.has_plot_handler():
-                plotted_windows[plot_fig_man.plot_handler.ws_name].append(plot_fig_man.plot_handler.plot_window)
+                plotted_windows[plot_fig_man.plot_handler.ws_name].append(
+                    plot_fig_man.plot_handler.plot_window
+                )
         return plotted_windows
 
     @classmethod
--- a/src/mslice/plotting/plot_window/plot_figure_manager.py
+++ b/src/mslice/plotting/plot_window/plot_figure_manager.py
@@ -425,6 +425,7 @@ class PlotFigureManagerQT(QtCore.QObject):
         if disable_make_current:
             self.disable_make_current()
 
+
 def new_plot_figure_manager(num, global_manager):
     def _new_plot_figure_manager(num, global_manager):
         """Create a new figure manager instance for the given figure.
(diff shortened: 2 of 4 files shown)


cailafinn (changes requested):
I'm running into something that I _think_ is a similar but potentially different issue:
1. Load a file.
2. Display the workspace
3. Create an interactive cut
4. Save cut to workspace
5. Switch to MDHisto and delete the workspace
6. Switch back to Data Loading and load the same file again
7. Back on MDHisto and you'll see the "deleted" workspace in the list
8. Click it - crash. 

Otherwise, the code looks good. If this turns out not to be related I'm happy to approve

warunawickramasingha:
> I'm running into something that I _think_ is a similar but potentially different issue:
> 
> 1. Load a file.
> 2. Display the workspace
> 3. Create an interactive cut
> 4. Save cut to workspace
> 5. Switch to MDHisto and delete the workspace
> 6. Switch back to Data Loading and load the same file again
> 7. Back on MDHisto and you'll see the "deleted" workspace in the list
> 8. Click it - crash.
> 
> Otherwise, the code looks good. If this turns out not to be related I'm happy to approve

Thanks for the review Caila.

This is an existing issue originating from a [different part of the code](https://github.com/mantidproject/mslice/blob/main/src/mslice/plotting/plot_window/slice_plot.py#L580C12-L580C65). When a cut window is closed it saves the cut workspace back causing to bring the [workspace_provider._loaded_workspaces](https://github.com/mantidproject/mslice/blob/main/src/mslice/models/workspacemanager/workspace_provider.py#L18) dictionary into an inconsistent state. 

The same crash can be observed from main branch by following exactly the same steps as given above, with an additional step 5.1 as below

`<same steps 1-5>`
5.1 Close the interactive cut plot
`<same steps 6-8>`

Please check the fix added for that.

cailafinn (changes requested):
Sorry to do this, but I think the last change may have re-introduced the original issue. I can now:
1. Load a workspace.
2. Display it.
3. Create an interactive cut
4. Delete the original 2D workspace in Data Loading
5. Select the Cut plot's window
6. Change the intensity
7. Crash.

cailafinn (approved):
Happy that the code works as described. Retested with the changes and looking good. Thanks for creating that issue.

PR #1178, opened 2026-05-08
Simplified GitHub action by installing into conda base

SilkeSchomann:
**Description of work:**

The GitHub action for publishing a Conda package used a Conda environment for building the package in two separate steps. This sometimes causes problems as the environment is not always switched reliably. By installing into the base environment this problem is avoided. In a CI runner, polluting the base environment does not cause new problems as it does not persist past running the GitHub action.

**To test:**

This can only be tested by merging into the code base and monitoring the respective GitHub action for a while as the original problem occurred randomly.


RabiyaF (approved):
It looks good to me. Thank you for making the changes 👍🏼

PR #1183, opened 2026-05-18
Set parent windows on dialogs so they are transient for their owning …

daym:
… application windows.

**Description of work:**

This PR sets a parent window on all dialogs so they are transient for their owning application windows, which means they will be shown together close to their application windows--and not somewhere else, or, the other extreme, just obscuring the entire application window they belong to.

**To test:**
Open the respective dialog window on a big screen. You'll see each dialog now is a (smaller) popup centered over the owning application window and stays close to it.

### DIFF
--- a/src/mslice/models/workspacemanager/file_io.py
+++ b/src/mslice/models/workspacemanager/file_io.py
@@ -20,19 +20,22 @@ import numpy as np
 from scipy.io import savemat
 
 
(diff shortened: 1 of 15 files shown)


daym:
I can explain further if needed. Basically, it just sets the parent window of the dialogs so the system knows what dialog is [requested for] for what window.

PR #1184, opened 2026-05-21
Fixed for publish action

SilkeSchomann:
**Description of work:**

The last MSlice nightly did not upload since there was a line continuation missing. Also, the removal step did not take into account that some nightly packages have both labels `nightly` and `main`. The one with a `main` label should never get removed.

**To test:**

Code review only.


thomashampson (approved):
Happy for this to go in to test itself

PR #1185, opened 2026-05-21
Add ability to manually trigger nightly build

thomashampson:
**Description of work:**
Adds a [workflow_dispatch trigger](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch) for manually setting off the nightly run

**To test:**
Just examine the change. Once merged, you should be able to go to the action and click "run workflow".


SilkeSchomann (approved):
Code looks fine and matches documentation for GitHub workflows.

PR #1192, opened 2026-05-26
Use fetch-tags instead of tags

### LINKED ISSUE #1189: Fix warning in deploy conda nightly workflow
For `actions/checkout@v4`, the parameter `tags` is not available anymore and should be removed.

SilkeSchomann:
**Description of work:**

There was a warning for the nightly conda deploy workflow saying that actions/checkout@v4 does not have a parameter tags. Now it is using "fetch-tags: true" instead. I have also upgraded checkout from v4 to v6.

**To test:**

Code review and double check documentation for actions/checkout@v6: https://github.com/actions/checkout

Fixes #1189.

### DIFF
--- a/.github/workflows/deploy_conda.yml
+++ b/.github/workflows/deploy_conda.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout MSlice
-        uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


thomashampson (approved):
This looks good and will hopefully allow versioningit to create the correct version number

PR #1203, opened 2026-06-17
Update to support Qt6. Drop pyqt5 and just use qtpy.

rosswhitfield:
**Description of work:**

We don't need to include `pyqt` in the conda recipe since `mantidqt` will pull in correct `pyqt` version. The should allow mslice package to work with both Qt5 and Qt6 versions of Mantid.

This should address the workbench package build issue in mantidproject/mantid#41652

I haven't actually tested mslice to see if anything fails when running under Qt6. That should be a follow up once we actually have it running with a Qt6 version of Mantid.

**To test:**

Fixes #xxxx.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -31,7 +31,6 @@ requirements:
   - ipython
   - matplotlib
   - numpy
-  - pyqt
   - qtconsole
   - qtpy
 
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
I've tested this MSlice version with the corresponding Mantid Qt6 version and did not find any issues.

PR #1209, opened 2026-07-07
Fix for rare ghost plot situation in slices or cuts

### LINKED ISSUE #1195: Ghost plot when adding grids to mslice.
**Describe the bug**

Found in Manual Testing, I've checked in 6.15 and it still appears. 

Following Manual Testing instructions for [Mslice](https://developer.mantidproject.org/Testing/Direct/MSliceTestGuide.html#the-plots-tab). 
When trying to add a grid on a plot by double-clicking on the axes: 

Image

 A new empty plot appears:

Image

This happens if you `keep` the plot. This is very important. All opened plots should be marked as `keep` before trying to change the grid. 

**Expected behavior**
No ghost plot.

**Screenshots**
If applicable/possible, add screenshots to help explain your problem.

 - MSlice Version Nightly
 - Seen Ubuntu and MacOS

**Additional context**
More context around this issue provided in #1209

adriazalvarez:
**Description of work:**

Fix for #1195 

The ghost plots are happening in both mac and linux (and probably windows too) when trying to modify a plot (from settings or quicksettings) and all opened plots are marked as kept. 
The way mslice is designed is that there is only one active plot at a given time, so if a plot is marked as kept, it automatically becomes inactive for the global figure manager, and any new cut or slice is displayed on a new plot window. But if all open plots are marked as `keep`, the figure manager will interpret that there are no active plots and set the active figure property to None (instead of the plot number). When a change that involves getting a handle of the active plot number (like changing the grid) is tried and all plots are `keep`, the global figure manager  creates an empty window plot. 

The fix that seems more correct for this design is to automatically mark any plot that is interacted with as active (`current`). This situation probably already presented before (as when bragg peaks are added to an open plot), as I only had to use the already implemented methods `report_as_current_and_return_previous_status` and `reset_current_figure_as_previous` before and after a plot interaction in the plot manager instances. 

bonus: There was a test that was failing on my local machine due to the relative folder path the python runner was using, I have changed it so that the path is insensitive to this by using pathlib.

**To test:**

- Follow #1195 test instructions, no ghost plots should appear when trying to change settings by clicking on the axes or on the settings (like axes limits or grid on/off).

Fixes #1195

### DIFF
--- a/src/mslice/plotting/plot_window/plot_figure_manager.py
+++ b/src/mslice/plotting/plot_window/plot_figure_manager.py
@@ -192,15 +192,27 @@ class PlotFigureManagerQT(QtCore.QObject):
         self.canvas.draw()
 
     def plot_clicked(self, event):
+        last_fignum, disable_make_curr_after = (
+            self.report_as_current_and_return_previous_status()
+        )
         if event.dblclick or event.button == 3:
             self.plot_handler.plot_clicked(event.x, event.y)
+        self.reset_current_figure_as_previous(last_fignum, disable_make_curr_after)
 
     def object_clicked(self, event):
+        last_fignum, disable_make_curr_after = (
+            self.report_as_current_and_return_previous_status()
+        )
         if event.mouseevent.dblclick or event.mouseevent.button == 3:
             self.plot_handler.object_clicked(event.artist)
+        self.reset_current_figure_as_previous(last_fignum, disable_make_curr_after)
 
     def _plot_options(self):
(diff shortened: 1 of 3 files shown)


RabiyaF (approved):
Thanks for fixing this and adding a detailed description of what was causing the ghost plot. I tested the `main` branch and saw a blank plot. However, When testing the `ghost_plot` branch, it did not appear. The code looks good and a test is included as well.

PR #1213, opened 2026-07-14
Added the `pixi` configuration files (pixi.toml and pixi.lock) to mslice

### LINKED ISSUE #1126: Investigate whether the mslice repo should start using pixi like Mantid

RabiyaF:
Fixes #1126.

**Description of work:**
This PR adds the **pixi.toml** and **pixi.lock** files to **mslice**. The `conda` files are **not** removed. This is to allow for a transition period to test both `conda` and `pixi` side by side. An additional unit test workflow has been added that uses `pixi`. 

**Summary of integrating pixi into mslice:**
1) The `pixi init --import environment.yml` command was run ([see documentation](https://pixi.prefix.dev/latest/reference/cli/pixi/init/)). This automatically generates the **pixi.toml** file from the specification defined in `environment.yml`. This command automatically added the .**gitattributes** file and updated the **.gitignore** file.

   I then modified the **pixi.toml** file so that the same file can be used to store the two environments defined in `environment.yml` and `mslice-developer.yml`. An additional environment called `mslice-env-test` was also added. This environment is the `mslice-env` with `mantid` and `mantidqt` installed. This is used in the **MSlice unit tests (Pixi)** workflow. It avoids the [Install Mantid](https://github.com/mantidproject/mslice/blob/main/.github/workflows/unit_tests.yml#L27L29) step in the **MSlice unit tests (Conda)** workflow. The **mantidnightly** environment has been renamed  **mslice-developer**. 

   I also added custom tasks called `test` and `coverage` that make the workflows neater. The abstraction of the whole command can be hidden behind a single word like shown below. 
image

2) The `pixi lock` command was run ([see documentation](https://pixi.prefix.dev/latest/reference/cli/pixi/lock/#config-options)). This generated the **pixi.lock** file for all the supported platforms. However, there was an issue with resolving the **qtconsole** packages on `linux`. 
image

   So, I changed the requirement of **qtconsole** to **qtconsole-base** which did not cause the tests to fail. But as Tom mentioned in the meeting, mantid also had to do [this](https://github.com/mantidproject/mantid/blob/a2d388d042b268bdae93b35eaa58af26e6bde47c/pixi.toml#L146). The `conda` environments were also updated to use **qtconsole-base** instead of **qtconsole** for consistency.

**To test:**

1) The **MSlice unit test (Pixi)** workflow should pass. It uses `pixi` to setup the environment on the `ubuntu` runners.

**Note** - The **MSlice unit test (Conda)** and **MSlice unit test (Pixi)** workflows has been updated so that the workflow will run on a PR created through a fork. A concurrency group has also been added to cancel old actions incase of pushes in quick succession.

**Further Work:**
The [workflow](https://github.com/mantidproject/mantid/blob/main/.github/workflows/update-pixi-lockfile.yml) that automatically updates the `pixi.lock` file needs to be added to mslice after this PR has been merged. Tom suggested lowering the frequency of the updates from daily to weekly on mslice and using the `Github Actions` bot rather than using the `Mantid Builder` bot to create the PRs.


pixi.toml
```diff
@@ -0,0 +1,55 @@
...
+
+[feature.mantidnightly.tasks]
+test = """
+xvfb-run --server-args='-screen 0 640x480x24' --auto-servernum \
+coverage run -m pytest
+"""
+coverage = """
+coveralls
+coverage report
+"""
```
jhaigh0, line 38: Confused how this task runs when `coveralls` is only a dependency of the `mslice-env` feature

pixi.toml
```diff
@@ -0,0 +1,55 @@
...
+
+[environments]
+# Equivalent to environment.yml
+mslice-env = ["mslice-env"]
+
+# Equivalent to mslice-developer.yml
+mantidnightly = ["mantidnightly"]
```
jhaigh0:
```suggestion
mslice-developer = ["mantidnightly"]
```
I think this would be less confusing

pixi.toml
```diff
@@ -0,0 +1,55 @@
...
+    "mantid",
+    "mantid/label/nightly",
+]
+
+[feature.mslice-env.dependencies]
+pyqt = "*"
+qtpy = "*"
+ipython = "*"
+coveralls = "*"
+qtconsole = "*"
```
jhaigh0: should this be `qtconsole-base` too? (and at that point it can be moved to the default dependencies)

pixi.toml
```diff
@@ -0,0 +1,55 @@
...
+channels = [
+    "conda-forge",
+    "mantid/label/nightly",
+]
+
+[feature.mantidnightly.dependencies]
+mantid = "*"
+mantidqt = "*"
+pre-commit = "*"
+qtconsole-base = "*"
```
jhaigh0: You should also update the conda developer recipe to user `qtconsole-base`, Tom let me know he should've made that change.

.github/workflows/unit_tests_pixi.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install Xvfb
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes xvfb
+
+      - name: Run tests
+        run: pixi run -e mantidnightly test
+
+      - name: Report coverage
+        run: pixi run -e mantidnightly coverage
```
jhaigh0: This job runs the tests using the same 'mantidnightly' environment as the other one, is that correct?

jhaigh0 (changes requested):
Had a few questions, also I think that we shouldn't leave it too long before removing the conda workflows, not any huge reason to have the tests running twice.

pixi.toml
RabiyaF: This resolves correctly and `pixi` is able to create the lock file. The idea was keep the `conda` and `pixi` environments as similar as possible. However, because of `qtconsole = "*"` there is only a difference in the `mantidnightly` `linux` environment. `qtconsole = "*"` gives an error in the `mantidnightly` environment because `mantid` and `mantidqt` are installed. 

Because of the difference, there are two workflows because I asked Silke about it yesterday and decided it was better to have the two workflows running side by side because this is a beta feature and some people would be using `conda`. And as the tests take a few minutes to run, I did not think that it was a big issue in terms of sustainability . The `conda` workflow can eventually be dropped.

pixi.toml
RabiyaF: If it is`qtconsole-base` in the `conda` recipe as well then it would make things consistent and I can clean up some of the conditions on target from the `pixi.toml` file.

.github/workflows/unit_tests_pixi.yml
RabiyaF: This should be `mslice-env` to make it consistent with the `MSlice unit tests (Conda)`. Both environments are named the same but one is created through `conda` and one through `pixi`.

pixi.toml
RabiyaF: You are right. But I have changed the environment now. But I am surprised that it did not log any error.

jhaigh0 (approved):

PR #1215, opened 2026-07-23
Added the versioningit to the docs optional-dependencies in pyproject.toml

### LINKED ISSUE #1211: Fix docs build and publish
**Describe the bug**
Documentation build fails because `versioningit` is not present to compute the version number inside the documentation conf.py. This problem started when the hard-coded version number was replaced [here](https://github.com/mantidproject/mslice/pull/1208). I think there are a few options to fix it. Here are some untested ideas for fixing it:
1. Add dependencies under the `[project.optional-dependencies]` section of pyproject.toml:
```
docs = [
    "versioningit",
]
```
2. Install `versioningit` in the docs publishing github workflow (might also need to explicitly checkout the repo to fetch the tags).
3. Change the code in conf.py to retrieve the version number some other way.

**To Reproduce**
See [here](https://github.com/mantidproject/mslice/actions/runs/29006040038) for the failed workflow run. You could add the `workflow_dispatch` trigger to help with testing.

**Expected behaviour**
Version number is dynamically picked up correctly.

RabiyaF:
**Description of work:**
Fixes #1211 .
Added versioningit to the docs optional-dependenices in `pyproject.toml`.

**To test:**
Verify that the **sphinx-notes/pages@v3** step is successful in the [triggered](https://github.com/mantidproject/mslice/actions/runs/30010633835/job/89217498313) workflow. It was unsuccessful in a previous [workflow](https://github.com/mantidproject/mslice/actions/runs/29985102977/job/89135119497).

### DIFF
--- a/.github/workflows/online_doc_update.yml
+++ b/.github/workflows/online_doc_update.yml
@@ -4,13 +4,9 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    branches:
-      - main
-  workflow_dispatch:
 
 concurrency:
-  group: pull-request-${{ github.event.pull_request.number }}-doc-build
+  group: push-on-main-doc-build
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/online_doc_update.yml
+++ b/.github/workflows/online_doc_update.yml
@@ -4,6 +4,14 @@ on:
   push:
     branches:
       - main
+  pull_request:
+    branches:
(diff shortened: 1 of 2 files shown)


SilkeSchomann (approved):
The code looks good and the test run does not fail because of an unavailable versioningit package anymore. Failing on permissions is the expected behaviour in this setup.

PR #1216, opened 2026-07-24
Fix __version__ reporting +uncommitted on clean builds

### LINKED ISSUE #1210: __version__ should not be "dirty" or "distance-dirty" in clean state
**Describe the bug**
Since the [fix to the conda package version number](https://github.com/mantidproject/mslice/pull/1208), the internal python package version number is being computed incorrectly. The value of mslice.__version__ is given a `+uncommitted` suffix, even when there are no modifications to the local repo. Figure out how to fix that.

**To Reproduce**
Build package:
```
mamba create -n rattler-build-env
mamba activate rattler-build-env
rattler-build build --recipe ./conda/recipe.yaml -c conda-forge -c mantid/label/nightly -c mantid --variant "version=$(python -m versioningit)" --output-dir output
```
Install and launch  package:
```
mamba create -n mslice-test
mamba activate mslice-test
mamba install -c ./output -c mantid/label/nightly mslice
mslice
```
Print the version number:
```
import mslice
print(mslice.__version__)
```

**Expected behaviour**
Version number should be consistent with conda package version number.

MohamedAlmaki:
**Description of work:**
mslice.__version__ has `+uncommitted` suffix (e.g. 2.16.0.dev8+uncommitted) even when built from a clean checkout, and it didn't match the conda package version. 

The root cause is that `tests/testdata/MAR21335_Ei60meV.nxs` is tracked in git, but the *.nxs rule in .gitignore matches it. When rattler-build copies the source into its build sandbox, it respects .gitignore so it skips the file. Inside the sandbox git then sees the file as deleted, the working tree looks dirty, and versioningit appends __version__ with +uncommitted. The conda package version is computed separately in the clean CI checkout, so only the internal __version__ was affected.

To fix the issue, the nxs test file is excluded in .gitignore so the tracked test file is no longer ignored, keeping the build tree clean.

**To test:**
1. Build the conda package as CI does (rattler-build ... --variant "version=$(python -m versioningit)").
2. Install into a fresh env and run python -c "import mslice; print(mslice.__version__)".
3. Expect 2.16.0.devN with no +uncommitted, matching conda list mslice.

Fixes #1210.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,8 @@ ENV/
 # Data files
 *.nxs
 *.nxspe
+# Keep this tracked test file so the conda build tree stays clean (mantidproject/mslice#1210):
+!tests/testdata/MAR21335_Ei60meV.nxs
 
 # Diff files
 *.diff


SilkeSchomann (approved):
This was tested locally and worked fine.

PR #1219, opened 2026-07-28
Added the setup-pixi action to install pixi on mslice for the pixi.lock autoupdate

RabiyaF:
**Description of work:**
This is an intermediary step between #1213 and #1218. The reason I need to merge this into `main` is because I am working through a fork and while testing #1218 I get a permissions error while creating a new PR. If this is merged into `main` then I can set `main` as the update branch as follows:
```
env:
  UPDATE_BRANCH: 'main'
```

And this will allow end to end testing of the `update-pixi-lockfile.yml` workflow added in the #1218 PR.

It is recommended to add the `.pixi-version` file containing the pixi version in the main folder on open source softwares rather than use a variable like `${{ vars.PIXI_VERSION }}`. Mantid does it [also](https://github.com/mantidproject/mantid/blob/895377ba19a1d4cf69dff5ba8a9499efd0285c69/.pixi-version#L1).

**Note:** I have renamed the environment `mslice-env-test` to `mslice-ci`. This environment was previously resolved for all platforms. As we only run the unit tests on `ubuntu` runners, we need to create it for the `linux` platform only. This can be changed in the future if we decide to test on more types of os runners. This change is causing the clean up of the lock file.

**To test:**

Verify that the `Run ./.github/actions/setup-pixi` passes [here](https://github.com/mantidproject/mslice/actions/runs/30356400380/job/90265506823?pr=1218). The create PR is understandably failing because of permissions on my fork.


RabiyaF:
I am also going to push a `pixi.toml` clean up here. This will make sure that when I run `update-pixi-lockfile.yml` through the other PR, I should definitely see an update to the lock file and a PR created.

SilkeSchomann (approved):
The renaming of the environment makes sense. Also, the changes are reasonable as an intermediate step for the ongoing work for the pixi setup 👍🏻

PR #1220, opened 2026-07-28
Updated the mslice codebase to be compatible with ruff v0.16.0

RabiyaF:
**Description of work:**
Updates the codebase so that it is compatible with `v.0.16.0` of **ruff** which was added in #1217.  The latest version of **ruff** enables **413** rulesets by default and that is why so many changes are required in this version update. See full release notes [here](https://astral.sh/blog/ruff-v0.16.0).
image

The code updates are made through this **PR** so that they are unlinked with any other kind of work.

image

There are **172** changes that could not be auto fixed that were causing the `precommit.ci` to fail in other PRs. Most of those errors are fixed in this PR. Some have been ignored through `--extend-ignore`.

**Note:** To make reviewing easier, I have squashed the commits. The first commit is the automatic updates by ruff. **The second commit is the one that needs closer inspection.** 

**To test:**

The `precommit.ci` workflow should pass.

**Further Information:**
In case something in mslice breaks due to the changes, then we can quickly revert to `v0.15.22` of **ruff** by reverting #1217 and #1220.


SilkeSchomann (approved):
I have reviewed the code and think these changes are reasonable.

